PR #126, opened 2015-01-12
Disable the loading of external entities in XML documents.

thijskh:
This is to proactively defend against local and remote file inclusion attacks.
This is in most cases already disabled by default in system libraries, but to
be safe we explicitly disable it also.


relaxnow:
Thanks, though I'd much rather just move to require an officially maintained PHP version (in which this has been fixed for quite some time).

thijskh:
I could not find a definitive answer on which PHP version fixes this (if any).

relaxnow:
Sounds like a nice topic for a blog post for the Ibuildings Blog, I'll add it as a topic.

PR #127, opened 2015-01-15
Feature/98 require only e ppn

relaxnow:


library/EngineBlock/Corto/Adapter.php
```diff
@@ -563,20 +563,27 @@ protected function _configureMetadataRepository(
...
-            new RequestedAttribute('urn:mace:terena.org:attribute-def:schacHomeOrganizationType', true),
-            new RequestedAttribute('urn:mace:dir:attribute-def:uid', true), // UID (example: john.doe)
-            new RequestedAttribute('urn:mace:dir:attribute-def:cn'),
-            new RequestedAttribute('urn:mace:dir:attribute-def:eduPersonAffiliation'),
-            new RequestedAttribute('urn:mace:dir:attribute-def:eduPersonEntitlement'),
-            new RequestedAttribute('urn:mace:dir:attribute-def:eduPersonPrincipalName'),
-            new RequestedAttribute('urn:mace:dir:attribute-def:preferredLanguage'),
-        );
+        $openConextIdentifierType = $this->_getOpenConextIdentifierTypeFromConfig();
+        if ($openConextIdentifierType != 'eduPersonPrincipalName') {
```
relaxnow: This should be solved differently, we should make use of the attribute metadata that we have in ```application/config/```.

library/EngineBlock/Corto/Filter/Command/ValidateRequiredAttributes.php
```diff
@@ -24,23 +24,35 @@ public function execute()
...
 
-        if ($this->_identityProvider->schacHomeOrganization) {
-            // ServiceRegistry override of SchacHomeOrganization, set it and skip validation
-            $this->_responseAttributes[self::URN_MACE_TERENA_SCHACHOMEORG] = array(
-                $this->_identityProvider->schacHomeOrganization
-            );
-        }
-        else {
-            $error = $this->_requireValidSchacHomeOrganization($this->_responseAttributes);
+        $openConextIdentifierType = $this->_getOpenConextIdentifierTypeFromConfig();
```
relaxnow: Again, this could better be accomplished with the attribute metadata.

library/EngineBlock/Saml2Attributes/FieldMapper.php
```diff
@@ -104,6 +114,37 @@ public function saml2AttributesToLdapAttributes($attributes)
                 'Missing required SAML2 fields in attributes'
             );
         }
+
+        if ($openConextIdentifierType == 'eduPersonPrincipalName') {
+            if (!isset($ldapAttributes['o'])) {
+                $ldapAttributes['o'] = $this->_getDomainFromEppn($ldapAttributes, 'none');
```
relaxnow: A 'none' domain?

relaxnow: Also we're trying to get rid of this magic LDAP attribute creation, not add some more... Why is this necessary?

relaxnow: Oh, I get it now, we usually get the ```o``` from the SHO, but in this case it doesn't have a SHO.

relaxnow: This shouldn't be necessary though, ```o``` is a MAY in the ```inetOrgPerson``` spec and we only use it for the deprovisioning query which can easily be rewritten to use ```findByIdentifier```.

library/EngineBlock/Saml2Attributes/FieldMapper.php
```diff
@@ -74,7 +78,13 @@ public function saml2AttributesToLdapAttributes($attributes)
     {
         $log = EngineBlock_ApplicationSingleton::getLog();
 
-        $required = $this->_saml2Required;
+        $openConextIdentifierType = $this->_getOpenConextIdentifierTypeFromConfig();
+
+        if ($openConextIdentifierType != 'eduPersonPrincipalName') {
+            $required = $this->_saml2RequiredShoUid;
+        } else {
+            $required = $this->_saml2RequiredEPPN;
```
relaxnow: I think we can just get rid of this required bit, move it down to the UserDirectory.

library/EngineBlock/Saml2Attributes/FieldMapper.php
relaxnow: Oh, but we allow querying with ```uid=x,o=y```, binding to that dn. Sounds useful but hasn't actually been used.

library/EngineBlock/Saml2Attributes/FieldMapper.php
```diff
@@ -104,6 +114,37 @@ public function saml2AttributesToLdapAttributes($attributes)
...
+
+        if ($openConextIdentifierType == 'eduPersonPrincipalName') {
+            if (!isset($ldapAttributes['o'])) {
+                $ldapAttributes['o'] = $this->_getDomainFromEppn($ldapAttributes, 'none');
+            }
+        }
         return $ldapAttributes;
     }
+
+    protected function _getDomainFromEppn($ldapAttributes, $defaultDomain) {
```
relaxnow: Nice, but you don't request the uid in the metadata and you don't autofill it with _getUidFromEppn?

library/EngineBlock/UserDirectory.php
```diff
@@ -222,7 +250,7 @@ protected function _addUser($newAttributes)
     {
         $newAttributes[self::LDAP_ATTR_COLLAB_PERSON_HASH]          = $this->_getCollabPersonHash($newAttributes);
 
-        $newAttributes[self::LDAP_ATTR_COLLAB_PERSON_ID]            = $this->_getCollabPersonId($newAttributes);
+        $newAttributes[self::LDAP_ATTR_COLLAB_PERSON_ID]            = $this->_getCollabPersonById($newAttributes);
```
relaxnow: Uh, so you're storing the EPPN in the collabPersonId attribute?

relaxnow:
Fixed by https://github.com/OpenConext/OpenConext-engineblock/pull/128

PR #145, opened 2015-05-06
Monilog

rjkip:
This is a work-in-progress PR for the Monilog project for easy feedback. Branch is subject to rebasing at @rjkip's discretion.

[Pivotal](https://www.pivotaltracker.com/n/projects/1327084)


relaxnow:
that's so money!

library/EngineBlock/ApplicationSingleton.php
```diff
@@ -103,6 +103,25 @@ public static function getLog()
...
+    public function flushLog($reason)
+    {
+        $activationStrategy = EngineBlock_Log_Monolog_Handler_FingersCrossed_ManualOrErrorLevelActivationStrategyFactory::getManufacturedStrategy();
+        $logger = $this->getLog();
+
+        if ($activationStrategy) {
+            $activationStrategy->activate();
+            $logger->notice($reason);
+        } else {
+            $logger->warning(sprintf("Unable to flush the log buffer (given reason to flush: '%s')", $reason));
```
rjkip: @thijskh, when the application is requested to flush any buffered log messages, but no log buffer is present (typically due to not being configured), should this be ignored, logged on warning level, logged on error level or be an outright exceptional case?

thijskh: I would say log on warning level: if something is broken about this you want to know but on the other hand it doesn't seem right to stop processing outright.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -12,11 +12,12 @@ public function serve($serviceName)
...
-            $this->_server->getConfig('debug', false) ||
-            EngineBlock_SamlHelper::doRemoteEntitiesRequireAdditionalLogging(array($sp, $idp))
-        ) {
-            EngineBlock_ApplicationSingleton::getInstance()->getLogInstance()->flushQueue();
+        if (EngineBlock_SamlHelper::doRemoteEntitiesRequireAdditionalLogging(array($sp, $idp))) {
+            $application = EngineBlock_ApplicationSingleton::getInstance();
+            $application->flushLog('Activated additional logging for the SP or IdP');
+
+            $log = $application->getLogInstance();
+            $log->info('Raw HTTP request', array('http_request' => (string) $application->getHttpRequest()));
```
rjkip: This exposes the user's cookies. Is this acceptable? @thijskh @pmeulen

rjkip:
@boy Could you review these changes?

relaxnow:
@rjkip I'm on holiday this week and the next. Is this blocking you? I'm at OWASP.EU next Thursday and could try to review then?

rjkip:
@relaxnow Thanks for responding anyway. @thijskh is also on holiday, but will be back next Monday, then we will be looking to get this PR merged and deployed. If you could take time to review it next Thursday, that would be great!

library/EngineBlock/Application/Bootstrapper.php
```diff
@@ -119,28 +120,35 @@ protected function _getAllConfigFiles()
...
         }
 
-        $this->_application->setLogInstance(
-            EngineBlock_Log::factory($this->_application->getConfiguration()->logs)
-        );
+        $loggerConfiguration = $configuration->logger->toArray();
+
+        /** @var EngineBlock_Log_LoggerFactory $loggerFactory */
+        $loggerFactory = $loggerConfiguration['factory'];
+        $logger = $loggerFactory::factory($loggerConfiguration['conf'], $configuration->debug);
```
relaxnow: A little more error checking here with something like ```is_callable``` might be useful.

library/EngineBlock/Application/Error.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->line = $line;
+    }
+
+    /**
+     * @param EngineBlock_Application_Error $other
+     * @return bool
+     */
+    public function equals(self $other)
+    {
+        return $this == $other;
```
relaxnow: Interesting use of double equals with type hinting!

library/EngineBlock/Application/ErrorHandler.php
```diff
@@ -60,24 +65,20 @@ public function error($errorNumber, $errorMessage, $errorFile, $errorLine)
             return false;
         }
 
-        $errorMessage = $errorMessage . " [$errorFile:$errorLine]";
-        $exception = new EngineBlock_Exception($errorMessage, EngineBlock_Exception::CODE_ERROR);
+        $this->lastErrorHandled = null;
```
relaxnow: Seems like this is not necessary?

relaxnow: ah, sorry, read this wrong.

library/EngineBlock/Application/Error.php
```diff
@@ -0,0 +1,100 @@
+<?php
+
+final class EngineBlock_Application_Error
```
relaxnow: Interesting, inspired by http://ocramius.github.io/blog/when-to-declare-classes-final/ ?
Sounds like a good idea though it may cause a proliferation of single use interfaces, but those are easy enough to manage.

rjkip: Since I read that blog I always make all classes final by default. To me, it's just like private properties, but for the inheritance chain. It can be removed should need be. I don't always create interfaces, like is the case here.

library/EngineBlock/Application/Bootstrapper.php
rjkip: Will do.

library/EngineBlock/Application/ErrorHandler.php
```diff
@@ -89,35 +90,31 @@ public function error($errorNumber, $errorMessage, $errorFile, $errorLine)
...
         if (ini_get('display_errors')) {
             echo "<br />" . PHP_EOL;
         }
         $message = 'A very serious error occurred, it has been logged and sent to the administrator.';
-        $message .= PHP_EOL . '<br /><br /> ERROR: ' . PHP_EOL;
-        $message .= '<br /><strong style="color: red"><pre>' . var_export($lastError, true) . '</pre></strong>';
-        die($message);
+        $message .= PHP_EOL . '<br /><br /> ERROR: ' . PHP_EOL . '<br />';
+        $message .= '<strong style="color: red"><pre>' . var_export($lastError->toArray(), true) . '</pre></strong>';
+        echo($message);
```
relaxnow: You're right, no need to die in the shutdown :smile:

library/EngineBlock/Log/Monolog/Handler/FingersCrossed/ManualOrErrorLevelActivationStrategyFactory.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @param array $config
+     * @return array
+     * @throws InvalidConfigurationException
+     */
+    private static function validateAndNormaliseConfig(array $config)
+    {
+        if (!isset($config['action_level'])) {
+            throw InvalidConfigurationException::missing('action_level', 'string');
+        } elseif (!is_string($config['action_level'])) {
```
relaxnow: I don't think this needs to be an elseif, but could simply be an ```if```.

library/EngineBlock/Log/Monolog/Handler/SyslogHandlerFactory.php
```diff
@@ -0,0 +1,79 @@
...
+
+        if (!LogLevel::isValid($config['min_level'])) {
+            throw new InvalidConfigurationException(
+                sprintf("'min_level' is not a valid log level string, '%s' given", $config['min_level'])
+            );
+        }
+
+        if (!isset($config['formatter']['factory'])) {
+            throw InvalidConfigurationException::missing("formatter.factory", 'string');
+        } elseif (!is_string($config['formatter']['factory'])) {
```
relaxnow: I also believe ```elseif``` statements in this method could be simplified to ```if```.

library/EngineBlock/Rest/Client.php
```diff
@@ -37,15 +37,11 @@ public function get($args = array())
...
-        $this->_logRequest();
 
         $this->_data = array();//Initializes for next Rest method.
 
         if ($response->getStatus() !== 200) {
-            EngineBlock_ApplicationSingleton::getLog()->attach(
-                $response->getHeadersAsString() . PHP_EOL . $response->getBody(),
-                'Response'
-            );
+            $this->_logRequest('Received response to REST request with a status code other than 200');
```
relaxnow: Why did you remove the logging of the response? This is actually something that is *very* useful in setting up EB.

relaxnow: *sigh* forget I said anything... reviewing in between talks is hardly conducive to the review, sorry about that.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
relaxnow: Logs are sensitive anyway, contain lots of SAML Responses. 
EB cookies in contrast don't particularly allow an attacker to do anything interesting on the gateway that I can think of.

relaxnow:
:+1: nice work!

library/EngineBlock/Log/Monolog/Handler/FingersCrossed/ManualOrErrorLevelActivationStrategyFactory.php
rjkip: Meh, it's shorter and equally legible and comprehensible imo. Unless you have concrete arguments I'd like to keep it this way.

rjkip:
Thanks for reviewing @relaxnow. Will do something about checking the callability of the factories. Enjoy OWASP and your vacation!

library/EngineBlock/Log/Monolog/Handler/FingersCrossed/ManualOrErrorLevelActivationStrategyFactory.php
relaxnow: It's shorter only in the sense that you're omitting whitespace by conflating two dependant but otherwise independent guard clauses. It's longer because you're using 4 more letters :wink:
But I would have to disagree on the legibility, it is much easier to debug 2 simple if statements than 1 long one. Also more maintainable as it is less tempting to add more clauses as else branches.
Then again, certainly for this use case and in this code base, it is very much picking at tiny nits an otherwise excellent PR.

library/EngineBlock/Application/Bootstrapper.php
rjkip: @relaxnow When the factory class or method does not exist, a fatal error is triggered. This halts the application and requires immediate administrator attention, just like any other configuration/bootstrapping issue. What would good would checking for callability add?

relaxnow: @rjkip It would make someone debugging the problem happier ("oh shit, I misspelled the class name") and would keep Fatal errors (that should never occur) separated from configuration errors.

rjkip:
:sparkles: @relaxnow

rjkip:
@relaxnow Any chance you could review these four changes today or tomorrow morning?

relaxnow:
:+1:

relaxnow:
@rjkip, @pmeulen, @thijskh this is now available on test environment.

Note that you should be able to switch between this PR and https://github.com/OpenConext/OpenConext-engineblock/pull/147 by doing the following on the test server:

```bash
sudo rm /opt/openconext/OpenConext-engineblock && 
sudo ln -s /opt/openconext/OpenConext-engineblock-redesign /opt/openconext/OpenConext-engineblock && 
service httpd restart
```

and back:
```bash
sudo rm /opt/openconext/OpenConext-engineblock && 
sudo ln -s /opt/openconext/OpenConext-engineblock-monilog /opt/openconext/OpenConext-engineblock &&
service httpd restart
```

Note that the following will need to be inserted in ```/etc/rsyslog.conf``` at the top:
```
# Increase max message size for OpenConext-engineblock
# See also: https://github.com/OpenConext/OpenConext-engineblock/pull/145
$MaxMessageSize 20k
```

And rsyslog needs a restart:
```bash
sudo service rsyslog restart
```

PR #148, opened 2015-05-08
Add logo generator script to create thumbnails for logos

Martin1982:
In the bin folder the script can generate logo images from the service registry and place these in a 'cached' directory on the 'www'-directories.

The command usage is 
`bin/logogenerator.php`

The command will ask you for the required parameters. For the purpose of putting this command in a cronjob these parameters can also be given as arguments:

* `-s <url>`: Service Registry URL to the connections JSON
* `-u <username>`: Service Registry HTTP username
* `-p <password>`: Service Registry HTTP password

### DIFF
--- a/bin/logogenerator.php
+++ b/bin/logogenerator.php
@@ -1,0 +1,114 @@
+#!/usr/bin/env php
+<?php
+require realpath(__DIR__ . '/../vendor') . '/autoload.php';
+
+define ('__LOGO_DIR__', __DIR__ . '/../www/authentication/cached');
+
+function getInput() {
+  $handle = fopen ("php://stdin","r");
+  $line = fgets($handle);
+  return trim($line);
+}
+
+function getFromCurl($urlSuffix, $serviceRegistry) {
+  $curl = curl_init();
+  curl_setopt($curl, CURLOPT_URL, $serviceRegistry->url . $urlSuffix);
+  curl_setopt($curl, CURLOPT_HTTPHEADER, array("Content-type: application/json"));
+  curl_setopt($curl, CURLOPT_SSL_VERIFYHOST, false);
+  curl_setopt($curl, CURLOPT_HTTPGET, true);
+  curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
+
+  if (!empty($serviceRegistry->user)) {
+    $userstring = $serviceRegistry->user;
+    if (!empty($serviceRegistry->password)) {
+      $userstring.= ':' . $serviceRegistry->password;
+    }
+    curl_setopt($curl, CURLOPT_USERPWD, $userstring);
+  }
+
+  $data = curl_exec($curl);
+  if (curl_errno($curl)) {
+    print "Error: " . curl_error($curl) . " on " . $serviceRegistry->url . $urlSuffix;
+    exit(1);
+  }
+
+  $data = json_decode($data);
+  curl_close($curl);
+  return $data;
+}
+
+function loadService($id, $serviceRegistry) {
+  return getFromCurl("/$id", $serviceRegistry);
+}
+
+function createThumb($logoUrl) {
+  $cachedName = md5($logoUrl);
+  $cachedLocation = __LOGO_DIR__ . "/" . $cachedName . ".png";
+  $image = @file_get_contents($logoUrl);
+  if (!$image) {
+    return;
+  }
+  file_put_contents($cachedLocation, $image);
+  try {
+    $image = \PHPImageWorkshop\ImageWorkshop::initFromPath($cachedLocation);
+    $image->resizeInPixel(64, null, true);
+    $image->save(__LOGO_DIR__, $cachedName . ".png");
+  } catch(Exception $e) {
+    echo $e->getMessage();
+  }
+}
+
+$options = "s:u:p:";
+$inputOptions = getopt($options);
+
+$registry = new stdClass();
+$registry->url = array_key_exists('s', $inputOptions) ? $inputOptions['s'] : "";
+$registry->user = array_key_exists('u', $inputOptions) ? $inputOptions['u'] : "";
+$registry->password = array_key_exists('p', $inputOptions) ? $inputOptions['p'] : "";
+
+if ($registry->url === "") {
+  $registry->url = "https://serviceregistry.surfconext.nl/janus/app.php/api/connections";
+}
+
+if (!array_key_exists('s', $inputOptions)) {
+  echo "What is the Service Registry URL (default: " . $registry->url . "):";
+  $url = getInput();
+  if (!empty($url)) {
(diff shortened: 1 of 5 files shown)


bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env php
+<?php
+require realpath(__DIR__ . '/../vendor') . '/autoload.php';
```
relaxnow: Could you add a file doc block that states what this script does?

relaxnow:
I'm missing the integration with push metadata.

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+
+define ('__LOGO_DIR__', __DIR__ . '/../www/authentication/cached');
+
+function getInput() {
```
relaxnow: If you're working with functions it's best to move then to the bottom of the file. This is no longer Pascal where the interpreter needs to read the definition before use :wink: .

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+  $handle = fopen ("php://stdin","r");
+  $line = fgets($handle);
+  return trim($line);
+}
+
+function getFromCurl($urlSuffix, $serviceRegistry) {
+  $curl = curl_init();
+  curl_setopt($curl, CURLOPT_URL, $serviceRegistry->url . $urlSuffix);
+  curl_setopt($curl, CURLOPT_HTTPHEADER, array("Content-type: application/json"));
+  curl_setopt($curl, CURLOPT_SSL_VERIFYHOST, false);
```
relaxnow: the horror...

Please remove this.

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+  curl_setopt($curl, CURLOPT_HTTPGET, true);
+  curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
+
+  if (!empty($serviceRegistry->user)) {
+    $userstring = $serviceRegistry->user;
```
relaxnow: Could you change to ```$curlUserPassword```?

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+    if (!empty($serviceRegistry->password)) {
+      $userstring.= ':' . $serviceRegistry->password;
+    }
+    curl_setopt($curl, CURLOPT_USERPWD, $userstring);
+  }
+
+  $data = curl_exec($curl);
+  if (curl_errno($curl)) {
+    print "Error: " . curl_error($curl) . " on " . $serviceRegistry->url . $urlSuffix;
+    exit(1);
```
relaxnow: :+1: for proper exit codes.
:-1: for not using stderr.

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  $data = json_decode($data);
```
relaxnow: This may have a failure condition when SR throws an exception. Would be very useful print to stderr and send appropriate exit code.

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+  curl_close($curl);
+  return $data;
+}
+
+function loadService($id, $serviceRegistry) {
+  return getFromCurl("/$id", $serviceRegistry);
```
relaxnow: A function for 1 character?

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+}
+
+function createThumb($logoUrl) {
+  $cachedName = md5($logoUrl);
+  $cachedLocation = __LOGO_DIR__ . "/" . $cachedName . ".png";
+  $image = @file_get_contents($logoUrl);
+  if (!$image) {
+    return;
+  }
+  file_put_contents($cachedLocation, $image);
```
relaxnow: I think this is missing some error handling in case the user doesn't have write rights?

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+  try {
+    $image = \PHPImageWorkshop\ImageWorkshop::initFromPath($cachedLocation);
+    $image->resizeInPixel(64, null, true);
+    $image->save(__LOGO_DIR__, $cachedName . ".png");
+  } catch(Exception $e) {
+    echo $e->getMessage();
+  }
+}
+
+$options = "s:u:p:";
```
relaxnow: SUP!

relaxnow:
Engineblock already has the configuration values for SR, it doesn't make sense to request the SR if it's a script embedded in EB.

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+function loadService($id, $serviceRegistry) {
+  return getFromCurl("/$id", $serviceRegistry);
+}
+
+function createThumb($logoUrl) {
+  $cachedName = md5($logoUrl);
+  $cachedLocation = __LOGO_DIR__ . "/" . $cachedName . ".png";
+  $image = @file_get_contents($logoUrl);
+  if (!$image) {
+    return;
```
relaxnow: Don't you want to generate a default? It will be pretty likely that a lot of entities will not have a logo on some environments.

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+
+$data = getFromCurl('', $registry);
+
+if (!property_exists($data, 'connections')) {
+  echo "Error: Invalid response, the key connections is not available";
+  exit(2);
+}
+
+foreach ($data->connections as $srItem) {
+  if (property_exists($srItem, 'id')) {
```
relaxnow: Could you change this to an early return?

relaxnow: Also feel free to log an error if a connection does not have an id, that is something that really really has to be there.

bin/logogenerator.php
```diff
@@ -0,0 +1,114 @@
...
+    $srId = $srItem->id;
+    $serviceData = loadService($srId, $registry);
+    $logo = @$serviceData->metadata->logo[0]->url;
+    if (!is_null($logo)) {
```
relaxnow: Again here don't you want to generate a default?

relaxnow:
Done reviewing, requires a bit more work I'm afraid. If you have any questions, feel free to look to your right and start shouting :stuck_out_tongue_winking_eye: .

relaxnow:
Discussed this with @Martin1982, won't be accepting this in the current form into EngineBlock.

Some additional work will need to be done (I'm estimating no more than half a day) but budget is an issue.

Alternative is to keep this out of EB and host it separately.

bin/logogenerator.php
Martin1982: No, the cached images are an assisting resource, the code for displaying the images for an entity have the responsibility of handling the way these are used.

PR #150, opened 2015-05-11
Voot migration

oharsta:
@Boy, can you please review? This pull request needs to be merged when the new components VOOT and Authorization Server are on acceptance and that will take some time, but it is good to have comments now.


composer.json
```diff
@@ -20,7 +20,7 @@
         "openconext/stoker-metadata": "~0.1",
         "openid/php-openid": "dev-master#a287b2d85e753c84b3b883ed8ee3ffe8692c8477 as 2.2.2",
         "pimple/pimple": "~2.1",
-        "simplesamlphp/saml2": "~0.5",
+        "simplesamlphp/saml2": "~0.6",
```
relaxnow: Thanks for the upgrade! But could you use 1.5? 0.6 and 1.5 are tags to the same commit but we're going to continue with the 1.x versions as we already have a stable API.

relaxnow: oh, this was my own change from before the 1.x versions. Heh, I'll check the other PR.

PR #151, opened 2015-05-11
Voot migration

oharsta:
New pull request against develop


library/EngineBlock/VirtualOrganization/GroupValidator.php
```diff
@@ -25,19 +25,17 @@ protected function _validateGroupMembership($subjectId, array $groups, $requireN
...
             if ($response->getStatus() === 200) {
-                $result = json_decode($response->getBody(), true);
-                if (isset($result['entry'])) {
-                    $memberShips = array();
-                    foreach ($result['entry'] as $group) {
-                        $memberShips[] = $group['id'];
-                        if (in_array($group['id'], $groups)) {
-                            return true;
-                        }
+                $groupsOfUser = json_decode($response->getBody(), true);
```
relaxnow: Could you check for a falsey value here and throw an exception? 
One common error condition for JSON APIs is to return an error in HTML (or just an empty response) with a 200 OK.

relaxnow:
:+1: for the code.
Do all OpenConext envs have a VOOT service yet? Is that part of OpenConext-API?

thijskh:
This migration has been postponed while we monitor other developments that might make the entire voot interface obsolete.

relaxnow:
@thijskh @oharsta @pmeulen what's the status of this? Any update?

thijskh:
Will most likely not happen at all; rather, the authorization performed by the new PEP will supplant the functionality. Will keep the PR for now until this is fully certain.

PR #184, opened 2015-09-30
On the WAYF, search in IdP keywords in a case-insensitive manner

rjkip:
This fixes a bug where IdPs like "Hogeschool van Amsterdam", having keywords `HVA`, could not be found by typing `hva`, `HVA`, `HvA` etc. into the WAYF IdP search bar.

The search term was lowercased, but the keywords were not. This meant keywords with uppercase characters could never match any search term that would normally match those characters case-insensitively.


rjkip:
Closed, opened new PR against develop. #185

PR #185, opened 2015-09-30
On the WAYF, search in IdP keywords in a case-insensitive manner

rjkip:
This fixes a bug where IdPs like "Hogeschool van Amsterdam", having keywords `HVA`, could not be found by typing `hva`, `HVA`, `HvA` etc. into the WAYF IdP search bar.

The search term was lowercased, but the keywords were not. This meant keywords with uppercase characters could never match any search term that would normally match those characters case-insensitively.


theme/material/javascripts/application.js
```diff
@@ -250,9 +250,21 @@ OpenConext.Discover = function () {
...
-      return value === value.toString() && value.indexOf(equalTo) !== -1;
+  /**
+   * @param {String} equalTo
+   * @return {ContainsCaseInsensitiveStringPredicate}
+   */
+  function containsCaseInsensitiveStringPredicate(equalTo) {
+    equalTo = equalTo.toLowerCase();
+
+    /**
+     * @name ContainsCaseInsensitiveStringPredicate
```
relaxnow: nice!

relaxnow: Though the naming here is really confusing... you call containsCaseInsensitiveStringPredicate which returns ContainsCaseInsensitiveStringPredicate ?

theme/material/javascripts/application.js
```diff
@@ -250,9 +250,21 @@ OpenConext.Discover = function () {
...
+     * @function
+     * @param {String} string
```
relaxnow: I wouldn't use ```string``` as a variable name here, for a minute I thought that you were calling toLowerCase on the prototype.

theme/material/javascripts/application.js
rjkip: You're absolutely right. Ideas? I can go more verbose with `createContainsCaseInsensitiveStringPredicate`, but maybe something shorter like `containsCaseInsensitive` is more readable. It's already quite verbose.

theme/material/javascripts/application.js
rjkip: This is perhaps due to a past experience? I don't know more about the variable than that it is *a* string.

theme/material/javascripts/application.js
relaxnow: ```createComparatorForString``` that returns ```caseInsensitiveStringComparator```?
This kind of lacks the part where you're doing partial application but then I don't know if there is any consensus in JavaScript how to denote that in the name.

But seeing as you have Underscore.js you could also rewrite it to:
```js
_.partial(function(a, b) { b.toLowerCase().indexOf(b.toLowerCase()); }, filterValue);
```
Maybe that is more eloquent?

theme/material/javascripts/application.js
relaxnow: I don't have much experience with this JavaScript as it's all brand new by Martins hand.
But I have done a fair bit of JS and it is not uncommon to reference ```String.``` for something. Also this might trip up the Java devs.

Both [PHP](http://php.net/usort) and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) use ```a``` and ```b``` for comparison arguments, maybe that's a better convention?

By the way this I'm really picking at nits here, feel free to just merge the damned thing after testing :smile: .

PR #199, opened 2015-10-28
Swap delete and enter button after removing IdP from preferred list

rjkip:
Addresses the issue where a IdP that has been moved from the preferred IdP list to the regular IdP list still shows a delete button.


rjkip:
@arnoutt @thijskh

arnoutt:
:beers:

thijskh:
:accept:

PR #210, opened 2015-11-06
Convert NotFoundHttpExceptions to proper 404 pages

DRvanR:
See https://www.pivotaltracker.com/story/show/107472822


DRvanR:
@rjkip @arothuis

rjkip:
This not only affects authentication, but also API. IMO it's about the integration of Symfony and the old framework. What about putting it in the compatibility layer?

I know the ticket is about Symfony's HTTP exceptions, but what about other exceptions? The old EB has a nice error page when exceptional conditions occur:

https://engine.surfconext.nl/authentication/idp/process-wayf
vs
https://engine.vm.openconext.org/authentication/idp/process-wayf

DRvanR:
Api is already covered by the use of `ApiHttpException` (furthermore, this is not limited to occurrences only from within the AuthenticationBundle (it's an eventlistener). 

All known and identified exceptions that were caught in controllers are handled in the same way by Symfony. What I did miss is that the old dispatcher has a "catch all" exception handling that allowed graceful recovery.
Will add that to this PR

rjkip:
> furthermore, this is not limited to occurrences only from within the AuthenticationBundle (it's an eventlistener)

Point in case. Feels to me the AuthenticationBundle is a specific part of the application that now handles not found pages for the entire application.

DRvanR:
Fair enough, will relocate it

DRvanR:
@rjkip :bell:

rjkip:
:sparkles::ship::sparkles:

Awe-struck.

arothuis:
:+1:

PR #215, opened 2015-11-16
Feature pep

relaxnow:


.gitignore
```diff
@@ -6,3 +6,4 @@
 /vendor/
 /www/authentication/generated/
 bin/ignore_me.php
+.DS_Store
```
relaxnow: Please add system dependant ignores to your global .gitignore.

PR #216, opened 2015-11-16
Feature pep

relaxnow:


composer.json
```diff
@@ -54,6 +54,7 @@
             "EngineBlock_": "library/",
             "EngineBlock_Test": "tests/library/",
             "Janus_": "library/",
+            "Pdp": "library/",
```
relaxnow: "Pdp" is probably not such a good vendor name: https://packagist.org/about

library/EngineBlock/Corto/Exception/PEPNoAccess.php
```diff
@@ -1,9 +1,9 @@
 <?php
 
-class EngineBlock_VirtualOrganization_VoIdentifierNotFoundException extends EngineBlock_Exception
+class EngineBlock_Corto_Exception_PEPNoAccess extends EngineBlock_Exception
```
relaxnow: We try to follow the ZF guidelines for naming in EB, meaning that evey abbreviations must be CamelCased. Like so "Php" instead of "PHP".

library/EngineBlock/Corto/Filter/Command/PolicyDecisionPoint.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public function execute()
+    {
+
+        $requirePep = $this->requirePep();
+
+        EngineBlock_ApplicationSingleton::getLog()->debug("Policy Enforcement Point consult " . $requirePep);
+
+        if ($requirePep)
```
relaxnow: Oi vey, that's what you get without PHPCS I guess. Here the newline is placed on the next line, but the very next if has the brace on the same line.

relaxnow: Also an early return here would be nice.

library/EngineBlock/Corto/Filter/Command/PolicyDecisionPoint.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return $this->_responseAttributes;
+    }
+
+    public function execute()
+    {
+
+        $requirePep = $this->requirePep();
+
+        EngineBlock_ApplicationSingleton::getLog()->debug("Policy Enforcement Point consult " . $requirePep);
```
relaxnow: Direct boolean to string conversions are generally pretty ugly (true becomes "1" and false becomes "").

library/EngineBlock/Corto/Filter/Command/PolicyDecisionPoint.php
```diff
@@ -0,0 +1,60 @@
...
+
+        if ($requirePep)
+        {
+            $validator = $this->_getValidator();
+            $hasAccess = $validator->hasAccess(
```
relaxnow: Nice to assign this to a variable and name it properly!

library/Pdp/PolicyRequest.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+/**
+ * Class Pdp_PolicyRequest
+ */
+class Pdp_PolicyRequest
+{
+    protected $Request;
```
relaxnow: member variables are "camelCased", not "CamelCased" :smile:

PR #230, opened 2015-12-24
Remove metadata composer lock file

arothuis:
The vendor directory is .gitignored, because vendors are installed via composer. Therefore, the vendor directory and its contents should not be present in the repository.

It was removed earlier, in [this commit](https://github.com/OpenConext/OpenConext-engineblock/commit/302b6f576b14b418cafc9b757d7124ebb27255a6), but possibly re-added by accident during the resolving of merge conflicts for [this merge](https://github.com/OpenConext/OpenConext-engineblock/commit/46b9967488a0716d950916cbe72a7eb800c6a39b).


arothuis:
@thijskh, @relaxnow

Issue #202, opened 2015-11-02
Log hashing algoritm of incoming assertions


thijskh:
We want to know which of our IdPs still use the deprecated SHA1 hashing algorithm. Therefore it would be nice to log this information. This should both be assertions and signed authn requests.

Log with level notice so that it appears in the log by default.

PR #238, opened 2016-01-13
Log signing algorithms

arothuis:
Adds logging of signature method algorithms of sent AuthnRequests and incoming Assertions.

[Ticket 110650550](https://www.pivotaltracker.com/story/show/110650550)
[Issue 202](https://github.com/OpenConext/OpenConext-engineblock/issues/202)

### DIFF
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -83,6 +83,17 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
             );
         }
 
+        if ($sspRequest->isMessageConstructedWithSignature()) {
+            $application = EngineBlock_ApplicationSingleton::getInstance();
+            $log = $application->getLogInstance();
+
+            $log->notice(sprintf(
+                'Received signed AuthnRequest from Issuer "%s" with signature method algorithm "%s"',
+                $sspRequest->getIssuer(),
+                $sspRequest->getSignatureMethod()
+            ));
+        }
+
         $ebRequest = new EngineBlock_Saml2_AuthnRequestAnnotationDecorator($sspRequest);
 
         // Make sure the request from the sp has an Issuer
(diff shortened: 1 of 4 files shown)


arothuis:
@thijskh @DRvanR

thijskh:
Looks good, but I think we want to log the algorithm for _incoming_ signed authentication requests instead of (next to) _outgoing_ ones, if that's straightforward to do. Just like we log them for _incoming_ assertions.

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -427,6 +438,15 @@ public function send(
             $log = $this->_server->getSessionLog();
             $log->info('HTTP-Post: Sending Message', array('saml_message' => $xml));
 
+            $samlMessage = SAML2_Message::fromXML($messageElement);
+            if ($samlMessage instanceof SAML2_AuthnRequest && $samlMessage->isMessageConstructedWithSignature()) {
```
DRvanR: If this logging isn't absolutely required (consult @thijskh) I'd remove this. the parsing of the XML back to objects has quite some impact (especially with signature verification) for the sake of just logging the signing method. There are better places to do that (where you set the key and algorithm to sign with for instance), and we tell it which algo to use, so it seems rather redundant.

DRvanR:
One thing left pending. If that has been resolved to satisfaction of @thijskh, :+1:  from me.

library/EngineBlock/Corto/Module/Bindings.php
thijskh: I already discussed with Alex that this is not the intended thing to log. It will be logging whatever we decide to send which we supposedly already know. Agree with you that it's probably better to remove it since it has some impact.

arothuis: Will remove it. We could, indeed, always add logging of the signature method (from the used signature) of outgoing messages whenever it is needed.

arothuis:
@thijskh

thijskh:
:+1:

PR #240, opened 2016-01-13
Update from 5.x-dev

arothuis:
Brings refactor up-to-date with dev.


arothuis:
@DRvanR

DRvanR:
After fixing #241, and having it merged correctly (see the comment in #241), please update this as well so that 5.x-refactor also gets the .travis.yml file. After that :+1:

Issue #245, opened 2016-01-15
develop has dependencies incompatible with PHP 5.3.3


relaxnow:
[vagrant@localhost engineblock]$ rm -rf vendor && ./bin/composer.phar install
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for symfony/symfony v2.7.4 -> satisfiable by symfony/symfony[v2.7.4].
    - symfony/symfony v2.7.4 requires php >=5.3.9 -> your PHP version (5.3.3) or value of "config.platform.php" in composer.json does not satisfy that requirement.
  Problem 2
    - symfony/symfony v2.7.4 requires php >=5.3.9 -> your PHP version (5.3.3) or value of "config.platform.php" in composer.json does not satisfy that requirement.
    - phpunit/phpunit 4.8.21 requires symfony/yaml ~2.1|~3.0 -> satisfiable by symfony/symfony[v2.7.4].
    - Installation request for phpunit/phpunit 4.8.21 -> satisfiable by phpunit/phpunit[4.8.21].

PR #246, opened 2016-01-15
Reverts mismatch of dependencies and php 5.3

arothuis:
In commit a93db11f (updates saml2 dependency), the saml2 package and its dependencies were updated without regard to the required php version.

This commit reverts that mismatch and updates saml2 and its dependencies to match the php 5.3 version requirement.

Fixes issue #245.


relaxnow:
:+1:

PR #251, opened 2016-01-19
Embed Engine Test Stand

DRvanR:
- [x] Fix failing features
- [ ] ~~Enable running tests on CI~~ _Postponed for now_
- [ ] ~~Ensure metadata repo is always mocked (currently only works if using janus backend)~~ _For now fixed in fixture files, will be improved when embedding metadata_
- [x] Either enable running tests in pre-commit-hook if runtime <10 seconds, otherwise in pre-push hook.


DRvanR:
@arothuis

arothuis:
:+1: works on my machine

PR #253, opened 2016-01-21
Load ini configuration in Symfony

arothuis:
[Ticket 111131614](https://www.pivotaltracker.com/story/show/111131614)


library/EngineBlock/Application/Bootstrapper.php
```diff
@@ -3,7 +3,7 @@
 class EngineBlock_Application_Bootstrapper
 {
     const CONFIG_FILE_DEFAULT       = 'configs/application.ini';
-    const CONFIG_FILE_ENVIRONMENT   = '/etc/surfconext/engineblock.ini';
+    const CONFIG_FILE_ENVIRONMENT   = '/etc/openconext/engineblock.ini';
```
DRvanR: this requires a change in deploy I'd recon?

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockConfiguration.php
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        $value = $this->configuration[$subPath];
+
+        if ($value instanceof EngineBlockConfiguration && !empty($subPaths)) {
+            return $value->get(join('.', $subPaths), $default);
+        }
+
+        if (is_string($value) || is_array($value)) {
+            return str_replace('%%', '%', $value);
```
DRvanR: `str_replace` on an array?

DRvanR: as an aside, could it even be an array?

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockConfigurationLoader.php
```diff
@@ -0,0 +1,33 @@
...
+    private $container;
+
+    /**
+     * @var EngineBlockIniFileLoader
+     */
+    private $fileLoader;
+
+    public function __construct(ContainerBuilder $container, EngineBlockIniFileLoader $fileLoader)
+    {
+        $this->container  = $container;
```
DRvanR: YUNEED container?

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockConfiguration.php
```diff
@@ -0,0 +1,55 @@
...
+            return $default;
+        }
+
+        $value = $this->configuration[$subPath];
+
+        if ($value instanceof EngineBlockConfiguration && !empty($subPaths)) {
+            return $value->get(join('.', $subPaths), $default);
+        }
+
+        if (is_string($value) || is_array($value)) {
```
arothuis: I've made this check explicit in order to prevent type conversion of non-string types. Therefore, `is_array` is not sufficient as an array could also contain non-string values. I'm not sure if it is required as arrays would be converted to `EngineBlockConfiguration` objects.

DRvanR: > I'm not sure if it is required as arrays would be converted to EngineBlockConfiguration objects.

That is exactly the point :) As far as I can tell, there is no possible way to get an array as value?

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockIniFileLoader.php
```diff
@@ -0,0 +1,72 @@
...
+     * @param array $mappedValues
+     * @return array
+     */
+    private function map($value, array $keys, array $mappedValues)
+    {
+        if (count($keys) > 1) {
+            return $this->map(array(array_pop($keys) => $value), $keys, $mappedValues);
+        }
+
+        $valueWithoutParameters = str_replace('%', '%%', $value);
```
DRvanR: please add a comment as to why it is important to do this.

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockConfigurationLoader.php
arothuis: I think the separate `EngineBlockConfigurationLoader` can be removed altogether as it is does not add anything at this point.

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockIniFileLoader.php
```diff
@@ -0,0 +1,72 @@
...
+     * @param string $file
+     * @return array
+     */
+    private function parseIniFile($file)
+    {
+        if (!is_string($file) || trim($file) === '' ) {
+            throw InvalidArgumentException::invalidType('non-empty string', 'file', $file);
+        }
+
+        if (!stream_is_local($file)) {
```
DRvanR: check can be removed as it is a filepath

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockIniFileLoader.php
```diff
@@ -0,0 +1,72 @@
...
+            throw new InvalidArgumentException(sprintf('This is not a local file "%s"', $file));
+        }
+
+        if (!file_exists($file)) {
+            throw new InvalidArgumentException(sprintf('File "%s" does not exist', $file));
+        }
+
```
DRvanR: missing the [`is_readable`][1] check

[1]: http://php.net/manual/en/function.is-readable.php

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockIniFileLoader.php
```diff
@@ -0,0 +1,72 @@
...
+        $parsedFile = parse_ini_file($file, true);
+
+        if ($parsedFile === false || $parsedFile === array()) {
```
DRvanR: would an empty ini file be considered invalid? Think this check could be removed (the `=== array()`)

library/EngineBlock/Application/Bootstrapper.php
arothuis: I'm not sure, because `engineblock.ini` is [already placed](https://github.com/OpenConext/OpenConext-deploy/blob/0d01f3ba9e3108c05312176d1d1dda3eac5e76f6/roles/engineblock/tasks/main.yml#L26) in `/etc/openconext/`. However `/etc/surfconext` is [symlinked](https://github.com/OpenConext/OpenConext-deploy/blob/a88d4292d98c5da8965b0ac4cd9334405172efb0/roles/openconext-common/tasks/main.yml#L10-L11) to `/etc/openconext`.

relaxnow: This was changed long, long ago in the deployment, waiting of the day a hero would come and change it in the application. :smile:

DRvanR: Then the creation of the symlink can be removed, better to remove unused functionality immediately.

src/OpenConext/EngineBlock/CompatibilityBundle/Configuration/EngineBlockIniFileLoader.php
```diff
@@ -0,0 +1,73 @@
...
+     * @return array
+     */
+    private function parseIniFile($file)
+    {
+        if (!is_string($file) || trim($file) === '' ) {
+            throw InvalidArgumentException::invalidType('non-empty string', 'file', $file);
+        }
+
+        if (!file_exists($file)) {
+            throw new InvalidArgumentException(sprintf('File "%s" does not exist', $file));
```
DRvanR: personally would change the messages to `'Could not parse given ini file, file "%s" does not exist'` etc., bit more informative.

src/OpenConext/EngineBlock/CompatibilityBundle/DependencyInjection/OpenConextEngineBlockCompatibilityExtension.php
```diff
@@ -20,5 +22,15 @@ public function load(array $configs, ContainerBuilder $container)
         $loader = new Loader\YamlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('services.yml');
         $loader->load('event_listeners.yml');
+
+        $engineBlockLoader = new EngineBlockIniFileLoader;
+        $engineblockConfig = $engineBlockLoader->load(array(
+            ENGINEBLOCK_FOLDER_APPLICATION . EngineBlock_Application_Bootstrapper::CONFIG_FILE_DEFAULT,
```
DRvanR: Any chance we could use `kernel.root_dir` here? Bit more Symfony and all :tongue:

library/EngineBlock/Application/Bootstrapper.php
arothuis: Said directory contains the files `engineblock.pem` and `engineblock.crt` and it could be that other applications depend on these files. I would suggest removing the link only when/if we are sure these files can be missed. Removing the link entirely would also be a BC break that should be documented accordingly.

arothuis:
@DRvanR

DRvanR:
:+1:

PR #259, opened 2016-01-27
Update from develop

DRvanR:
Inludes update from develop. What is NOT included is the API test page created in https://github.com/OpenConext/OpenConext-engineblock/commit/02a12bc134955fe0137df2869c2f29efa3f9b63e - the session bug was not something we had and thus does not need to be resolved.
If the API test page is a requirement (I'd advise against it), this needs to be added.

Functional tests pass
```
behat-travis:
     [exec] ...................................................................... 70
     [exec] ...................................................................... 140
     [exec] ...................................................................... 210
     [exec] ...................................................................... 280
     [exec] ...................................................................... 350
     [exec] ...................................................................... 420
     [exec] ...................................................................... 490
     [exec] ...................................................................... 560
     [exec] ...................................................................... 630
     [exec] ...................................................................... 700
     [exec] ...................................................................... 770
     [exec] ...................................................................... 840
     [exec] ...................................................................... 910
     [exec] ...................................................................... 980
     [exec] ...................................................................... 1050
     [exec] ...........
     [exec] 
     [exec] 38 scenarios (38 passed)
     [exec] 1061 steps (1061 passed)
     [exec] 2m41.359s
```


DRvanR:
@quartje @thijskh @arothuis : all conflicts resolved, all tests are green.

After having decided on a course of action with respect to the API test page, this can be merged and 5.0.0-alpha1 can be tagged for deployment, being equal in feature completeness to develop.
the use of `bin/makeRelease.sh` has not been verified yet, but I imagine that should not be much trouble (even to do manually should it be needed).

thijskh:
I don't understand the problem with the API test page and why it can't be merged. I'd like to understand it.

Nonetheless it's not crucial so if it currently impedes progress we can continue without it for now.

DRvanR:
It is not a problem of merging it (it has been merged), however the file modified (the controller) no longer exists in the 5.x range. Furthermore the template has not been added as part of the theme, but rather as a separate page and would have been removed when applying the theme. Both the controller code and template would thus have to be recreated in the 5.x range code.
This recreation is also not an issue in and of itself, but I was wondering what the use of a test page for an API that should not be publicly accessible is. Also, it seems that either it does nothing but check if your basic auth is correct, or actually process the posted configuration (if I understand it correctly) which to me would seem rather unsafe.

If the functionality is required I would suggest adding it, but with a feature toggle that by default is off so that it has to be a conscious choice to enable this page.

That said it could well be I misunderstood the functionality, but I opted for "rather safe then sorry" as it seemed unsafe to me and adding the functionality at a later point is trivial.

thijskh:
The page is intended so haproxy has something to do a healthcheck against. It should basically just respond with a 200 so we now the machine in question is up. It could do more but we need at least something that can return a 200 on GET or HEAD.

DRvanR:
That is already provided in the heartbeat controller
(src/OpenConext/EngineBlockBundle/Controller/Api/HeartbeatController), this
would be a test page for metadata push api if I read it correctly?

thijskh:
Right. I don't think the test page is currently essential.

arothuis:
:+1:, will merge and tag.

PR #261, opened 2016-02-01
Add symfony support to makeRelease, exclude theme from release

DRvanR:


DRvanR:
@thijskh @arothuis questions, remarks?

thijskh:
I'm wondering if there aren't cases where having the theme files available for a local update would be convenient.

DRvanR:
I don't have any issue with keeping them. However the theme is dumped and the dumped files are in the source tree as well, editing the source tree within a release to me makes more sense than editing the theme and then dumping it to overwrite the actual files being used. Thoughts?

thijskh:
Because of minification I think editing something like css is then a bit harder..

DRvanR:
Keeping themes dir now, any other thoughts/objections/comments?

thijskh:
Fine!

PR #265, opened 2016-02-03
Added Attribute Aggregation as Corto inbound filter

oharsta:
Feature branch for attribute aggregation in EB for 5.x branch.

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -209,6 +209,17 @@ pdp.username = "pdp_admin"
 pdp.password = "secret"
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;; ATTRIBUTE AGGREGATION SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Location of AA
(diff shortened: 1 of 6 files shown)


DRvanR:
Note: requires https://github.com/OpenConext/OpenConext-deploy/tree/engineblock-5.x-dev to be up to date with master before being able to merge.

application/configs/application.ini
```diff
@@ -209,6 +209,17 @@ pdp.username = "pdp_admin"
 pdp.password = "secret"
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;; ATTRIBUTE AGGREGATION SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+; Location of AA
+attributeAggregation.baseUrl = "https://aa.surfconext.nl/aa/api/attribute/aggregate"
```
DRvanR: Is this a wanted default? Can imagine that this should be something not-surfconext?

thijskh: `aa.demo.openconext.org` seems to follow what has been done in other places. Although we'd have to admit that many other of these defaults in application.ini currently suffer from the same problem.

DRvanR: Personally I'd opt for `aa.vm.openconext.org` to ensure that it works with the basic local install using OpenConext-deploy.

DRvanR: The usage of SURFconext is indeed quite prolific, this will be fixed in a separate PR.

PR #269, opened 2016-02-05
Introduce Assertions

DRvanR:
This means we have a standardized method to enforce assertions without having to rely on custom implementations. Replaced the current implementations, updated the test-suite accordingly. Also adjusted the grouping of test to be more overarching concern or functionality based than file-based.


DRvanR:
@arothuis

src/OpenConext/EngineBlock/Assert/Assertion.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Assert\Assertion as BaseAssertion;
+
+/**
+ * @method static void nullOrNonEmptyString($value, $message = null, $propertyPath = null)
+ * @method static void allNonEmptyString($value, $message = null, $propertyPath = null)
+ */
+class Assertion extends BaseAssertion
+{
+    const INVALID_NON_EMPTY_STRING = 501;
```
arothuis: Why is this set to 501? Is it to have our own range of (Assertion) exception codes?

DRvanR: Yes, the gap between this and existing codes.is big enough to not cause issues in the foreseeable future. Furthermore 5xx range coincides nicely with 5xx range http status codes (issue on our side), but that's more a nice coincidence than anything else really.

src/OpenConext/EngineBlock/Authentication/Value/ConsentType.php
```diff
@@ -38,13 +38,11 @@ public static function implicit()
...
-            throw InvalidArgumentException::invalidType(
-                'ConsentType::TYPE_EXPLICIT|ConsentType::TYPE_IMPLICIT',
-                'consentType',
-                $consentType
-            );
-        }
+        Assertion::choice(
+            $consentType,
+            array(self::TYPE_EXPLICIT, self::TYPE_IMPLICIT),
+            'ConsentType must be one of ConsentType::TYPE_EXPLICIT, ConsentType::TYPE_IMPLICIT'
```
arothuis: Although you could use a parameterized string for the `Assertion::choice` method to handle, I like the mentioning of the class constants as such instead of the values they represent (what would happen if you let `choice` manage the message formatting).

arothuis:
`Assertion::wellDone($pr)`

:+1:

PR #270, opened 2016-02-05
Introduce mockery

DRvanR:


DRvanR:
@arothuis - so we can both start using it :wink:

arothuis:
:+1:

PR #273, opened 2016-02-11
Fix engineblock fixtures dependency

arothuis:
As Engineblock uses parts of EngineBlockFixtures in non-test code,
it should not be a development-only dependency and it should be
shipped on release.

This dependency should be fixed in the future; production code
typically should not depend on EngineBlockFixtures.


arothuis:
@rjkip

rjkip:
:ship:

PR #275, opened 2016-03-04
Add OpenLDAP ldif schema files

mrvanes:
OpenLDAP can't handle .schema files. These files are generated by slapd based on the schema files.


thijskh:
Thanks. In the longer term we expect LDAP to disappear from EB entirely, but meanwhile this is a useful fix.

PR #277, opened 2016-03-04
De-SURFnet Engineblock

mrvanes:
Renamed surfnet pointers to have a cleaner OpenConext base when checking out.


thijskh:
Thanks!

PR #285, opened 2016-04-01
Log logins to syslog

DRvanR:
This will replace the current logging to `log_logins` table in the database.

Current implementation will output the following message in syslog:
```
Apr  1 11:49:00 apps EBAUTH[4861]: {"channel":"authentication","level":"INFO","message":"login granted","context":{"login_stamp":1459504140,"user_id":"urn:collab:person:engine-test-stand.openconext.org:test145950413313365","sp_entity_id":"https:\/\/engine.vm.openconext.org\/functional-testing\/No%20ARP\/metadata","idp_entity_id":"https:\/\/engine.vm.openconext.org\/functional-testing\/TestIdp\/metadata","key_id":null,"user_agent":"Symfony2 BrowserKit"},"extra":{"session_id":"8zxzInsnhMWUOLJpXyz9uxX2TCa","request_id":"56fe440bc31ea"}}
```


DRvanR:
@thijskh @quartje An example message can be found in the description. Currently `time()` is being used to create a timestamp (as there is a timestamp in the DB). Is that still wanted? Can imagine that when parsing from strings it would be simpler to have something like `2016-04-01T13:38:01+02:00`....

thijskh:
Yes, I would prefer ISO 8601 format timestamps.
It looks OK to me. I doubt whether we need the UA to be in there. I think it can be dropped.

@baszoetekouw Can you confirm that when engineblock logs the above info, this would be sufficient for our stats processing?

baszoetekouw:
The information in the example line above (specifically: timestamp, userid, sp_entityid, idp_entityid) should suffice for stats processing.
Note though, that this change requires a rewrite of te log_analyze component of stats.  Fixing this function https://github.com/OpenConext/log-analyze/blob/master/scripts/lib/logins.php#l143-l320 is the main part of that, but probably the threading model should be fixed also, if we're going to be reading sequentially from a stream or text file.

thijskh:
User-Agent can be removed.
Otherwise: :+1:

PR #287, opened 2016-04-08
UserDirectory with database and optionally without LDAP

DRvanR:


thijskh:
This looks fine to me

DRvanR:
It has been decided that both backends should be supported, with default using both backends (to allow for graceful rollover). The LDAP integration can be disabled through usage of a feature toggle (provisioned in https://github.com/OpenConext/OpenConext-deploy/pull/65). An additional migration script will be provided that allows for migrating the users from LDAP to the database backend.

DRvanR:
There may be better tools to migrate the users from LDAP to database than a PHP script, seeing the SURFnet case has ~1,400,000 distinct users. Therefor a separate tool/script will be created for the migration.

library/EngineBlock/Corto/Filter/Command/ProvisionUser.php
```diff
@@ -24,28 +28,34 @@ public function getCollabPersonId()
...
-            'subjectIdAttribute',
-            EngineBlock_UserDirectory::LDAP_ATTR_COLLAB_PERSON_ID
-        );
-        if (empty($user[$subjectIdField])) {
+        $uid = $this->_responseAttributes['urn:mace:dir:attribute-def:uid'][0];
+        $sho = $this->_responseAttributes['urn:mace:terena.org:attribute-def:schacHomeOrganization'][0];
+        if (!$uid || !$sho) {
             throw new EngineBlock_Exception(
-                "SubjectIdField '$subjectIdField' does not contain data for user: " . var_export($user, true)
+                'Cannot register user due to missing uid and/or schacHomeOrganization attribute'
```
baszoetekouw: Could you please make the error message more specific and include in which of the attributes are missing?

DRvanR: sure thing, was done to keep it the same as the [current EB4 implementation][1]

[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/8151a7bfc06e698726811ba498efdb074dd49a4a/library/EngineBlock/Saml2Attributes/FieldMapper.php#L104-L106

thijskh: I believe this is not normally triggered when an end user is missing certain attributes, since this is handled earlier in the code:
https://github.com/OpenConext/OpenConext-engineblock/blob/8151a7bfc06e698726811ba498efdb074dd49a4a/library/EngineBlock/Corto/Filter/Input.php#L38

src/OpenConext/EngineBlockBridge/Authentication/Repository/UserDirectoryAdapter.php
```diff
@@ -0,0 +1,149 @@
...
+        if (!isset($attributes[Uid::URN_MACE][0])) {
+            throw new EngineBlock_Exception_MissingRequiredFields(sprintf(
+                'Missing required SAML2 field "%s" in attributes',
+                Uid::URN_MACE
+            ));
+        }
+        if (!isset($attributes[SchacHomeOrganization::URN_MACE][0])) {
+            throw new EngineBlock_Exception_MissingRequiredFields(sprintf(
+                'Missing required SAML2 field "%s" in attributes',
+                SchacHomeOrganization::URN_MACE
```
DRvanR: @baszoetekouw better?

DRvanR:
depends on https://github.com/OpenConext/OpenConext-deploy/pull/65

src/OpenConext/EngineBlockBridge/Authentication/Repository/UserDirectoryAdapter.php
baszoetekouw: Excellent, thanks!

PR #296, opened 2016-06-15
Lazily bind the LDAP client on demand

DRvanR:


mrvanes:
Tried the patch and can confirm I now have an LDAP-less engineblock/janus installation that works.

DRvanR:
@thijskh as indicated to @quartje (and verified by@mrvanes :wink: ) this indeed allows EngineBlock to run without an LDAP installation, if the Ldap feature toggle is set to disabled.

Issue #298, opened 2016-06-23
Fails hard when provisioning new user


thijskh:
This is with EB 5.0 ß 4 but can also be reproduced with 5.0 ß 3.

Logging in to EB as a user that has no row yet in `users` fails with a hard error:
```
Catchable fatal error: Object of class OpenConext\Component\EngineBlockMetadata\Service could not be converted to string in /opt/openconext/OpenConext-engineblock-5.0.0-beta4/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php on line 70
```

Interestingly enough it seems that EB tries to interpolate something into this query when it happens: `UPDATE sso_provider_roles SET certificates = ?, name_id_formats = ?, response_processing_service_binding = ?, allowed_idp_entity_ids = ?, requested_attributes = ? WHERE id = ?`. EB should probably never update sso_provider_roles during the login flow in the first place.

This might be related to the fact that `class UserRepository extends EntityRepository` as introduced in 1cd9adac19885137579b13bd2074efec7eca1bae?

DRvanR:
Unfortunately I seem not to be able to reproduce this locally... However, some investigation does seem to indicate a possible cause:

It seems that this was existing, but unintended and untriggered behavior. What seems to be the case is that some code actually alters the `requestedAttributes` and other properties as listed on the ServiceProvider object (which is the case in some places, e.g. [here][2]) and that that causes Doctrine to detect and attempt to persist these changes when creating the new user.

A bit more detail: 
When a new user is registered, the user is persisted and the changes are committed. This causes the UnitOfWork to calculate all changes that occurred for registered Entities (all entities are tracked in an IdentityMap). If an Entity (such as a ServiceProvider) is loaded and altered before the new User is created and persisted, this means that the UnitOfWork would detect this Entity as "dirty", which means it differs from the Database version. Once the commit is made (the `flush` call), all changes are persisted to the database. Previously, there was no flush call made during a login, only when updating the metadata (metadata push). This means that any changes made to an Entity would be lost. This is now no longer the case, as the changes are committed (or at least, attempted to) when a new user is registered. Therefor it seems that this behavior existed, but was simply not triggered due to not committing changes to the database.

It seems the commit that is referenced indeed causes what is happening here, but only as an accidental side-effect. This also explains why this was not apparent in the dev envs: apparently the configuration there is such that there are no changes made to the ServiceProvider Entity, thus not triggering this effect.

[...]

PR #301, opened 2016-07-07
Feature/engineblock5 compatible metadata

DRvanR:
_note: the build will always fail due to missing database table_

Included updated engineblock-metadata library, see https://github.com/OpenConext/OpenConext-engineblock-metadata/pull/9.

# WARNING

This should only be used for testing purposes!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "kriswallsmith/assetic": "~1.2.0",
         "monolog/monolog": "~1.13",
         "mrclay/minify": "~2.2",
-        "openconext/engineblock-metadata": "~1.4",
+        "openconext/engineblock-metadata": "dev-engineblock5-compat",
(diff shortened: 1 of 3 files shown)


DRvanR:
@thijskh @quartje :  This should resolve #298 by only persisting the metadata when explicitly told to (only done on push). 
Furthermore, as requested, the metadata is now stored in a different table (`sso_provider_roles_eb5`). This to allow for safe testing of metadata push on a production node without tampering with the EB4 metadata. Do note that this requires a copy of `sso_provider_roles` to be made manually. Should this new table name be required for the 5.x version, a migration can be made that does so automatically.

quartje:
:+1:

thijskh:
:+1:

quartje:
I will merge this to create a 5.0.0-beta5-1 release

PR #306, opened 2016-07-21
Use latest stable version of metadata.

### LINKED ISSUE #298: Fails hard when provisioning new user

DRvanR:
The latest version of metadata lib is used, migration is provided

Fixes #298

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "kriswallsmith/assetic": "~1.2.0",
         "monolog/monolog": "~1.13",
         "mrclay/minify": "~2.2",
-        "openconext/engineblock-metadata": "dev-engineblock5-compat",
+        "openconext/engineblock-metadata": "^2.0",
         "openconext/stoker-metadata": "~0.1",
         "openid/php-openid": "dev-master#ee669c6a9d4d95b58ecd9b6945627276807694fb as 2.2.2",
         "pimple/pimple": "~2.1",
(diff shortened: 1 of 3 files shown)


DRvanR:
Overwrites the https://github.com/OpenConext/OpenConext-engineblock-metadata version introduced in #301 with the released version containing the same code, see https://github.com/OpenConext/OpenConext-engineblock-metadata/releases/tag/2.0

DRvanR:
@quartje @thijskh, this should resolve #298 and has a migration for the new metadata table.

thijskh:
:+1:

Issue #302, opened 2016-07-14
Unsollicited SSO fails with "Unable to find the current binding."


thijskh:
In EB 5, when triggering an unsollicited SSO request, EB throws an error:

```
Jul 14 09:15:24 papp2 EBLOG[2371]: [2016-07-14 09:15:24] engineblock.INFO: Handling incoming request: GET /authentication/idp/unsolicited-single-sign-on/ba573f07093978e3852ddef0d2465b84 {"session_id":null,"request_id":"57873c0c5e898"} []
Jul 14 09:15:24 papp2 EBLOG[2371]: [2016-07-14 09:15:24] request.INFO: Matched route "authentication_idp_unsolicited_sso_idphash". {"session_id":null,"request_id":"57873c0c5e898"} {"route_parameters":{"_controller":"engineblock.controller.authentication.identity_provider:unsolicitedSingleSignOnAction","virtualOrganization":null,"keyId":null,"domain":"surfconext.nl","idpHash":"ba573f07093978e3852ddef0d2465b84","_route":"authentication_idp_unsolicited_sso_idphash"},"request_uri":"https://engine.surfconext.nl/authentication/idp/unsolicited-single-sign-on/ba573f07093978e3852ddef0d2465b84?sp-entity-id=https%3A%2F%2Flogin.adp.nl%2Ffederation%2Fsamlsp%2F"}
Jul 14 09:15:24 papp2 EBLOG[2371]: [2016-07-14 09:15:24] security.INFO: Populated the TokenStorage with an anonymous Token. {"session_id":null,"request_id":"57873c0c5e898"} []
Jul 14 09:15:24 papp2 EBLOG[2371]: [2016-07-14 09:15:24] engineblock.WARNING: Unable to find the SAML 2 binding used for this request. {"session_id":"VtfodJjhxjXSeQS3XF,olUGIl41","request_id":"57873c0c5e898"} []
Jul 14 09:15:24 papp2 EBLOG[2371]: [2016-07-14 09:15:24] engineblock.WARNING: Request method: 'GET' {"session_id":"VtfodJjhxjXSeQS3XF,olUGIl41","request_id":"57873c0c5e898"} []
Jul 14 09:15:24 papp2 EBLOG[2371]: [2016-07-14 09:15:24] engineblock.WARNING: GET parameters: 'sp-entity-id' {"session_id":"VtfodJjhxjXSeQS3XF,olUGIl41","request_id":"57873c0c5e898"} []
[...]

Issue #304, opened 2016-07-15
Incomplete routing for unknown-preselected-idp


thijskh:
Being redirected to `/authentication/feedback/unknown-preselected-idp?idp-hash=…` (when attempting login with a scoped sso login URL with an unknown hash) results in a 404 not found page. This seems to be due to the final stanza in `src/OpenConext/EngineBlockBundle/Resources/config/routing/feedback.yml` being incomplete.

DRvanR:
Resolved through #307

PR #307, opened 2016-07-22
Correct IdP initiated single sign on

DRvanR:


DRvanR:
Resolves #302 and #304

DRvanR:
@quartje @thijskh

thijskh:
:+1: for 302 and :+1: for 304

PR #308, opened 2016-07-25
Attribute Manipulation Functional Tests

### LINKED ISSUE #303: Setting NameId from attribute manipulation broken
The attribute manipulation functionality allows to override attributes but also the SAML NameId by setting the `$subjectId` variable. Since EB 5.0 setting the `$subjectId` variable no longer has an effect and the original NameId is passed through unchanged.

Manipulation of specific attributes through the `$attributes` array remains working.

This regression was observed once before in #142 (but I don't know if the causes are related).

DRvanR:
resolves #303

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,17 @@ php:
   - 7.0
 
 matrix:
+  fast_finish: true
   allow_failures:
   - php: 7.0
+  - env: COMMAND="functional-tests-wip"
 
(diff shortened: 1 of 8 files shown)


DRvanR:
Verifying https://github.com/OpenConext/OpenConext-engineblock/issues/303, if a fix is needed, the fix will be added here as well.

Issue #64, opened 2014-08-11
Full support for SP Proxies


relaxnow:
EngineBlock has no concept of SP Proxying, as such it's persistent NameID algorithm, designed to give different values for different Service Providers will always give out the same NameID for a proxying SP.
- [X] ACL / WAYF
- [x] Attribute Manipulation
- [x] Attribute Release Policies
- [x] Consent
- [x] Logging
- [x] Persistent NameIDs

PR #309, opened 2016-07-28
Pass $response by reference to the filter commands.

thijskh:
This explains the buggy behaviour described in #303 and restores that behaviour in PHP 5.4 and up.

### DIFF
--- a/library/EngineBlock/Corto/Filter/Abstract.php
+++ b/library/EngineBlock/Corto/Filter/Abstract.php
@@ -30,7 +30,7 @@ abstract class EngineBlock_Corto_Filter_Abstract
      * @throws Exception
      */
     public function filter(
-        EngineBlock_Saml2_ResponseAnnotationDecorator $response,
+        EngineBlock_Saml2_ResponseAnnotationDecorator &$response,
         array &$responseAttributes,
         EngineBlock_Saml2_AuthnRequestAnnotationDecorator $request,
         ServiceProvider $serviceProvider,


thijskh:
@DRvanR

relaxnow:
Oi, the difference between object by pointer and by reference, that is obscure behaviour. Nice find!
Though I don't see why this is related to the PHP version as that is unchanged since the beginning of OOP support for PHP: https://3v4l.org/SRMHY

relaxnow:
Would have to be related to https://github.com/thijskh/OpenConext-engineblock/commit/d85f8c3bcb777420dc2c2eb6fe9c8eb661be0f3c I fear... 😅

DRvanR:
Just taking a quick glance, but I don't really see how this would fix the reported issue. The reported issue was about changing the `$subjectId`, which is a variable that is being passed by reference, from the [filter command][1] to the [manipulator][2], which hands it down to a [protected method][3], and does so [by reference][4], which then allow for runtime modification, also [by reference][5]. Scary, but done properly.

Also, as @relaxnow noted, since the introduction of the object model in PHP, this kind of handling of object seems to be the same regardless of by reference or not, as is also outlined in [the manual][manual].

Furthermore, you're not [actually dealing with that in an attribute manipulation][6], but with a [different object][7] whose modifications are [copied back to the referenced value][8], which in turn results in [a new object][9].

So while this may allow overwriting an object with a different object at a high level, which kinda makes it like a form of global state manipulation (yikes), it still does not explain to me how an SP with only `$subjectId = 'NOOT'` as AM could have a differing Subject NameID and EPTI NameID, and how this would fix that :confused: :question: ([referring to this][link])

That said, if it fixes the issue, it fixes the issue. Just don't see how :astonished: 

All in all, the more I see of this code, the more I am in favor of replacing this code ASAP. Using return values in stead of reference passing would already improve everything, and allow for stricter checking on what is actually going on, as well as logging any potential modification which would significantly simplify testing and debugging. As it stands, _this would actually be my priority number one_. Oh, and it would allow for better sandboxing, as closures now have a different scope than in php 5.3, they are actually scoped to the object they are defined in. Using [bind][10] we could bind it to an object that we could sandbox as much as possible to prevent possible contamination of the runtime.

Thoughts?

[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Corto/Filter/Command/RunAttributeManipulations.php#L67-L75
[2]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php#L19-L26
[3]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php#L41-L50
[4]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php#L56-L65
[5]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php#L70-L82
[manual]: http://php.net/manual/en/language.oop5.references.php
[6]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php#L32-L35
[7]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Corto/Mapper/Legacy/ResponseTranslator.php#L28
[8]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php#L52
[9]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Corto/Mapper/Legacy/ResponseTranslator.php#L41-L52
[10]: http://php.net/manual/en/closure.bind.php
[link]: https://github.com/OpenConext/OpenConext-engineblock/issues/303#issuecomment-234997708

thijskh:
The commit message has a bit more text, did you see that? The problem seems related to a behaviour change of `call_user_func_array`.

DRvanR:
Yeah, read that, but where is that method called? A [search][1] doesn't show any related calls, hence why I don't get it :grinning: 

[1]: https://github.com/OpenConext/OpenConext-engineblock/search?l=php&q=call_user_func_array&utf8=%E2%9C%93

relaxnow:
It's this one: https://github.com/OpenConext/OpenConext-engineblock/blob/8151a7bfc06e698726811ba498efdb074dd49a4a/library/EngineBlock/Corto/ProxyServer.php#L802

As that is where Corto gives the response to the 'filter' command on the Input or Output filter class.

thijskh:
I have updated Boy's example code above to show how the different PHP versions differ in behaviour in this specific case with `call_user_func_array`. https://3v4l.org/3g5cb

`call_user_func_array` is [used in][1] ProxyServer in `callAttributeFilter()` which in turn is [used to][2] call all Output attribute filters. In `callAttributeFilter`, `$response` and `$responseAttributes` are passed by reference via `call_user_func_array`. This reference is however lost because called function, `filter()`, does not have the reference in its signature.

This used to work in PHP 5.3, because there when using call_user_func_array, variables were passed into the called function by reference _regardless_ of that function's signature. In 5.4 and up the called function needs to explicitly specify that it needs the reference.

It explains precisely the difference between the Subject NameID and the ePTID value. Because the latter is an attribute. In filter(), `$responseAttributes` was already passed by reference, so any modifications made by filters is kept. The Subject NameID however changes a property of the `$response`.

[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Corto/ProxyServer.php#L800
[2]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/library/EngineBlock/Corto/ProxyServer.php#L780

DRvanR:
Check! ~~However, in my mind the proper fix would then be to remove the `call_user_func_array` (and the method that invokes it), doesn't really seem to add anything to be honest - and prevents this issue from re-occurring at this level.~~ (this might actually break even more stuff relying on this behaviour in other commands)

Just as a summary (please confirm):
- An SP requesting a NameID with format unspecified and an AM of $subjectId = 'NOOT' receives a Subject with a NameID with value NOOT and format unspecified and an EPTI containing a NameID valued NOOT and format unspecified  
- An SP requesting a NameID with format persistent and an AM of $subjectId = 'NOOT' receives a Subject with the persistent NameID and format, and an EPTI containing a NameID valued NOOT and format unspecified
- An SP requesting a NameID with format transient and an AM of $subjectId = 'NOOT' receives a Subject with a transient NameID and format, and an EPTI containing a NameID valued NOOT and format unspecified

(FYI: These are all cases I can verify as working using the functional tests)

What does not work is:
- setting the customNameID, originalNameID or intendedNameID on the response object in an AM. 

This is due to the fact that the changed values are correctly carried over to the new object, but since the by-reference chain was broken in the newer PHP version at the `call_user_func_array`, the object was never overwritten at the level of the EngineBlock_Corto_ProxyServer, which therefor kept on using the original (and unmodified) response object, while everything from the filter itself down had the response object overwritten due to the creation and returning of a new object using the new values in the ResponseTranslator, where the function parameter was by reference, hence overwriting the actual object itself.

All in all it seems a bit brittle to rely on this behaviour: https://3v4l.org/N6Bbk behaviour. Notice that the eval does not actually change the original object, but that values are copied over to a new object and then copied again after the modification (eval) to a new object, which works due to the by referencing everywhere. 

So yeah, I'm (albeit reluctantly :stuck_out_tongue: ) :+1: to merge this fix, but I would strongly urge to prioritize restructuring the commands and endless by reference handling as relying on object references of parameters being replaced somewhere on the call stack is rather implicit and brittle to me.

DRvanR:
Forgot to add: good find indeed!

thijskh:
@DRvanR re the bulleted list: I believe only bullet 1 holds. Or put it another way: the epTID is always a copy of the Subject NameId.

The only exception is when the SP is a trusted_proxy (see #64).

DRvanR:
Closed through #308

PR #310, opened 2016-07-28
Make functional test for NameId manipulation more specific.

thijskh:
I believe this would catch the problem in #303.


DRvanR:
hehe, yeah, already had these changes as well (not all tests have been pushed yet) :+1:

DRvanR:
Closed through #308

PR #313, opened 2016-07-29
Log which scope the IdP was sending that fails our data.

thijskh:
This is essential information to understand violations.


DRvanR:
:+1:

PR #327, opened 2016-08-23
Update requirements

thijskh:
Grouper and wget are not requirements of EB. PHP can also be used other than as an Apache module. Make Serviceregistry a link to janus.


DRvanR:
:+1:

PR #330, opened 2016-08-30
Document that PHP LDAP module is not optional.

thijskh:
Even when not using LDAP, current code still needs PHP LDAP module present, in order to pass a check for this module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ See the [UPGRADING.md](UPGRADING.md) file
 * Linux
 * Apache
 * PHP 5.6:
-    - ldap (optional)
+    - ldap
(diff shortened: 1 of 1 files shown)


DRvanR:
Hrms, I thought that with#296 the LDAP was optional and that it no longer required the php module if not using the LDAP functionality :disappointed: 

In any case :+1:

thijskh:
That is right, you don't need it, but there's a check when loading Zend Ldap that checks for module presence in the constructor. If you just install the module, everything works fine and no actual LDAP configuration needs to be present. So it's easy enough to work around and not much of a burden to keep the module installed for now.

PR #331, opened 2016-09-01
Update dump-required-ini-params.sh

mrvanes:
For some reason on my install + composer.phar here, this script borked in the toArray() method on something that was already an array. Removing the toArray() solved all my problems.


DRvanR:
@arothuis could you please verify this? Haven't had this issue before.
@mrvanes did you have anything configured for the `trustedProxyIps`?

arothuis:
Hi @mrvanes, thanks for the PR. we have not encountered this issue before and will look into it.

In the meantime, could you tell us what you have configured for the `trustedProxyIps` in both your `application.ini` (in the codebase) and in the `engineblock.ini` (on the VM)? 
Also, what VM (branch, environment) did you provision?

mrvanes:
I'm doing a manual install without OpenConext-deploy and never supplied trustedProxyIps in engineblock.ini so I suspect it's the default. I did see two ip addrs in trustedProxyIps config as a result of prepare-env (after I fixed dump-required-ini-params.sh).

arothuis:
Indeed, if the configured value is empty, `$config->get('trustedProxyIps', array())` will return an array and `->toArray()` will not work. That should be fixed.

However, removing `->toArray()`, results in `trusted_proxies = null` in `ini_parameters.yml` (the result of `dump_ini_params.sh`), if `trustedProxyIps` was configured in `engineblock.ini`. 

We could fix this by:
1. Getting the configuration first and only calling `toArray()` on it if it is not already an array; or 
2. Passing an `EngineBlockConfiguration` wrapping an empty array as a default value

I think solution 1 would be most transparent. If there were more array configurations to be extracted we could encapsulate such behaviour in a function.

DRvanR:
:+1:  for the first option.

DRvanR:
:+1:

tvdijen:
This now breaks on line 34.. I guess the $trustedProxies var should be used there, instead of pulling it from the config again.

arothuis:
That's odd, thanks for reporting it. 

Looking into it, something might have gone wrong using `commit -p`.

PR #332, opened 2016-09-02
Proof of concept: changing the order of corto filters

arothuis:
This relates to issue [127990709](https://www.pivotaltracker.com/story/show/127990709).

This PR aims to proof that it is feasible to move the Corto Output Filters to the Corto Input Filters and that there are challenges to overcome, which relate to the NameID in the EduPersonTargetedId being represented as a DOMNodeList (this also applies to `urn:oid:1.3.6.1.4.1.5923.1.1.1.10`):

1. **Displaying the EPTI in the provide consent screen:** regular attributes are represented as strings, the EPTI is represented as a DOMNodeList. If the EPTI should be shown in this screen, there should be a way to correctly represent it as a string in the same way as the other attributes are represented. If the EPTI should not be shown, there should be a way to filter them.
2. **Storing the EPTI in the session:** in order to use the same Response before consent is provided as when the consent is processed, the Response is stored in `$_SESSION`. This no longer works correctly as the contents of DOMNodeLists in PHP cannot be stored in the $_SESSION. When processing consent, this will mean that the DOMNodeLists will turn up empty.

This is related to the discussion at https://github.com/simplesamlphp/saml2/pull/60#issuecomment-235610299.

In this PR, I have introduced quick and dirty ways to mitigate these issues in order to verify the feasibility of moving all Corto Output Filters to Corto Input Filters.

### As this is a proof of concept, it is not intended to be merged!

### DIFF
--- a/library/EngineBlock/Corto/Filter/Input.php
+++ b/library/EngineBlock/Corto/Filter/Input.php
@@ -16,6 +16,8 @@ class EngineBlock_Corto_Filter_Input extends EngineBlock_Corto_Filter_Abstract
      */
     protected function _getCommands()
     {
+        $diContainer = EngineBlock_ApplicationSingleton::getInstance()->getDiContainer();
+
         $logger = $this->_server->getSystemLog();
         return array(
             // Show an error if we get responses that do not have the Success status code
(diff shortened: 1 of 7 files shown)


DRvanR:
Development is blocked by https://github.com/simplesamlphp/saml2/pull/76 and the backported version of that to the 1.x range.

arothuis:
Closing this as the underlying issue has been fixed and an implementation has been prepared. See #336.

PR #333, opened 2016-09-06
Remove hardcoded Strict-Transport-Security header.

thijskh:
We want to control this per application in the webserver, which seems a more appropriate place to handle this kind of configuration. Its settings are not easily changed when it's hardcoded.


DRvanR:
:+1: 

as an FYI: for the long run it was the intention to use https://github.com/nelmio/NelmioSecurityBundle for additional security related features so that we could get rid of as much hardcoded elements as possible  - HSTS would be configurable (or of course disabled :wink: ).

thijskh:
Great idea to use that, but things like HSTS I consider more webserver configuration than application configuration, e.g. in the same category as TLS ciphers. Would surely be nice to use some of the Nelmio features that are more application-centric.

PR #334, opened 2016-09-15
Adding Makefile

DRvanR:
Apache Ant adds significant overhead to all the tasks. This leads to quite some slowdown. To combat this, a PoC Makefile has been introduced that mimics the build.xml functionality. For each task this proves to add a 20%-50% speed increase. There are however some drawbacks:
- ~~it does not fully work on Mac, Mac uses BSD style `sed` which does not work with the `util-enable-func-tests` and `util-disable-func-tests`. No idea how to fix this.~~ This has been resolved in baeb5ba
- ~~tasks cannot be interrupted. Make will keep on executing the current task before exiting on a sigkill. This might be annoying when running the full regression functional test suite.~~ resolved in 	412f6e9

Still wanted to share it, perhaps there are people more experienced at writing these that could help and/or other people see use in this, removing dependency on an external task runner to be present (apache ant).

make


thijskh:
@pmeulen

arothuis:
`sed` is very difficult to get right on both Mac (BSD) and Linux (GNU). I committed some changes to see if we can use perl instead. It works on my machine 😉

DRvanR:
@thijskh @pmeulen any objections to merging this? I've been using this for quite some time now as alternative to ant.

thijskh:
By all means, if it works for you, please merge it.

My only concern is that we have Makefile and ant next to each other, we'll be fixing stuff in two places.

DRvanR:
A valid concern. However I would recommend sticking with ant for now (as that is used by the CI as well for instance) and treating the Makefile as a bonus, not something critical. 

I'll merge it for now, can always be removed again if it proves to be a burden.

PR #336, opened 2016-09-16
Optionally run all manipulations prior to consent

DRvanR:
Implementation of OpenConext/OpenConext-engineblock#332

For now copied the AM and ARP functional tests to ensure those work correctly regardless of when the filters run, don't know if there are more features/scenario's that make sense to add...


DRvanR:
@thijskh @surfnet-niels This is the PR for optionally running the output filters before consent rather than after consent. This should also be tested on a test env I think.

library/EngineBlock/Corto/Filter/Output.php
```diff
@@ -1,10 +1,36 @@
 <?php
+use OpenConext\Component\EngineBlockMetadata\Entity\IdentityProvider;
+use OpenConext\Component\EngineBlockMetadata\Entity\ServiceProvider;
 
 /**
  * Called by Corto *after consent*, just as it prepares to send the response to the SP
```
thijskh: Should this comment be updated too?

DRvanR: will do

DRvanR:
Your understanding is correct. The tests differ from the existing tests by having [this line][1], which changes the configuration used to run all filters before the consent screen. This should result in the exact same behaviour as in the original tests, hence why the features are copied. Given a newer version of Behat this could be solved in a different (better) manner, but migrating Behat just for this is quite a bit more work than the feature itself is, and is out of scope.

[1]: https://github.com/OpenConext/OpenConext-engineblock/pull/336/files#diff-0d7c360b5863d66d1ead00795ebb6c61R15

thijskh:
Alright; so the tests test whether the behaviour is the same in the given scenarios, but not if the feature toggle has any and/or the right desired effect.

DRvanR:
That's my point in the description, I'm not really sure how to test that... Thought about (and just discussed with @surfnet-niels ) adding the following scenario's to a new feature:

Scenario A: feature toggle off

0. feature toggle is off  [input filters > consent > output filters]  
Add AM that adds an attribute with distinct value  
1.  Start at SP, login
2. At EB, pass through WAYF
3. At IdP, log in
4. At EB, verify that the consent screen does not list added attribute
5. At EB, pass through consent
6. At SP, verify Assertion contains added attribute

Scenario B: feature toggle on

0. feature toggle is on [input filters > output filters > consent]  
Add AM that adds an attribute with distinct value
1. Start at SP, login
2. At EB, pass through WAYF
3. At IdP, log in
4. At EB, verify that the consent screen lists added attribute
5. At EB, pass through consent
5. At SP, verify Assertion contains added attribute

This would verify that indeed the AM are run before the consent screen when the feature toggle is on (same could be done with ARP). This however seems to test the feature toggle, but both tests would pass if someone just moved the AM filter from output to input filtering. So it kinda gives a false sense of security, but it may give enough security about that it works as intended. Not really sure if there is a payoff here...

arothuis:
@DRvanR, @thijskh; tests have been added to verify the attribute manipulations' impact on the consent screen with the feature toggle on and off. Tests regarding ARPs and the consent screen have also been added.

library/EngineBlock/Corto/Filter/Output.php
DRvanR: @arothuis could you also resolve this?

DRvanR:
:+1: from me!

PR #337, opened 2016-09-16
Reorganize and optimize WAYF JavaScript

arothuis:
#### Issues:
[127991057](https://www.pivotaltracker.com/story/show/127991057): Optimize WAYF
[127716355](https://www.pivotaltracker.com/story/show/127716355): Fix keyboard controls

#### Improvements:
* optimize the WAYF by tackling search from a more data-centric approach, instead of the previous (slower) view-centric approach: data is modified/filtered and then rendered
* improve usability and rendering speed by only showing a fixed amount of results per list
* organize the code into modules in a manner that is more decoupled from the UI
* allow developers to develop in modern ES2015 (ECMAScript 6) by transpiling it via babel and grunt
* remove some dependencies, reducing JavaScript package size from 129 KB to 19.5 KB
* remove the need for Bower in favor of NPM for JavaScript package management
* improve compatibility with older browsers (IE8) by conditionally adding (minified) shims and polyfills

#### Related issues:
[130523403](https://www.pivotaltracker.com/story/show/130523403): Further decouple WAYF JavaScript logic from UI 
[130523139](https://www.pivotaltracker.com/story/show/130523139): Reconsider WAYF Request Access functionality


DRvanR:
@thijskh @surfnet-niels This is the PR for the client-side optimizations of the WAYF. We have tested it with 10k-50k services, and then the server seems to become the real bottleneck, although from 50k onwards initial searches do seem to have a noticeable delay.

There is one difference with the current UI, the amount of Services shown are limited to optimize the (re)rendering. 

This should be tested to behave correctly on the test-env prior to merging.

thijskh:
What is the concrete meaning of "the amount of Services shown are limited"?

DRvanR:
The concrete meaning is that currently a maximum of [50][1] services are shown, if more services are available, a block is shown at the end of the list [informing the user about this][2].

The rationale to do this is that less UI elements to update makes it faster and rendering even a 10th of the target of 10k services in the WAYF will give browser related issues, especially on lower end devices (pc's/laptop/tablet/etc.). 50 services seems to be around the sweet spot of being big enough to accomodate a lot of cases without showing aforementioned "there are more" and small enough to always render fast. Furthermore, based on our purely subjective opinion, it seems the be high enough to warrant that a user might instinctively start searching rather than scrolling :wink:
The number can easily be changed, and the goal was to have it configurable through application configuration, but as a nice-to-have that was deemed out of scope for now.

[1]: https://github.com/OpenConext/OpenConext-engineblock/pull/337/files#diff-1e2a3cca9c3bbe075603830c57fe9efcR146
[2]: https://github.com/OpenConext/OpenConext-engineblock/pull/337/files#diff-5c098204d95878f54d90f99b62cd8221R49

thijskh:
@arnoutt

arnoutt:
- @DRvanR I guess where you say services you meen IdPs right? ;-)
- 1 thing that is not completely clear to me: does the WAYF *always* show 50 (or whatever) IdPs or does it only limit the number of IdPs when the user starts searching?
- Is lazy loading still supported? (When the user reaches the bottom of the page, another 50 (or whatever) IdPs are loaded additionally)
- How is it decided which IdPs are shown? A-Z? The ones logged into most? Random? :-) Are there any options there?
- I'm not sure whether it makes sense to show 50 IdPs when there are, say, 1k with access (when the user didn't enter any search terms yet). Wouldn't it make more sense to only show IdPs when the user enters a search term? (And of course the previously chosen.) Compare the default pyFF WAYF on openconext.org. It either let's you search or let's you choose from a list, but not a combination of both ;-) (I guess for SURFconext this would make sense, since there are few cases where an SP is accessible to more than 50 IdPs. But that's only one implementation.)
- The OpenConext WAYF also contains an option to show the IdPs *without* access to a service. Has this been taken into account as well?

arnoutt:
Thanks, it does!

app/Resources/views/modules/Authentication/View/Proxy/discover.phtml
```diff
@@ -99,72 +99,114 @@ $layout->assign('helpLink', '/authentication/idp/help-discover?lang=' . $lang);
...
-          <noscript>
+
+            <div class="idp-list">
+            </div>
+
+        <noscript>
+          <?php foreach ($idpList as $key => $idp) : ?>
+          <a class="result active access" href="#" tabindex="-1">
+              <div class="logo">
+                <img class="logo" src="/images/placeholder.png"/>
```
DRvanR: Will do

app/Resources/views/modules/Authentication/View/Proxy/discover.phtml
```diff
@@ -99,72 +99,114 @@ $layout->assign('helpLink', '/authentication/idp/help-discover?lang=' . $lang);
...
+        'requestAccessUrl' => '/authentication/idp/requestAccess?' . http_build_query([
+            'lang' => $lang,
+            'spEntityId' => $metaDataSP->entityId,
+            'spName' => $spDisplayName,
+        ]),
+    ];
+    ?>
+
+    <script id="wayf-configuration" type="application/json">
+        <?php echo json_encode($configuration); ?>
```
DRvanR: yes :smile:

baszoetekouw:
In this context, would it perhaps make sense to also address some related pagespeed issue with the WAYF?  For example, running Google's pagespeed tools on the WAYF gives:
screencapture-developers-google-speed-pagespeed-insights-1474452751954
screencapture-developers-google-speed-pagespeed-insights-1474452784949

I'm not all of those suggestions make sense.  I would for example argue that minimizing JS en CSS isn't very useful as enabling gzip content compression solved the problem of large js/css file transfers.  In fact, I consider ti harmful, as it makes debugging quite a bit harder.

DRvanR:
Most of the pagespeed issues (content compression, expiry headers) should be done on the server configuration level, in OpenConext/Openconext-deploy. Those things are definitely out of scope for this PR. 

A small aside with respect to minification: it will not conflict with debugging - source maps can (and should) be used for that.

arothuis:
@arnoutt, @thijskh (cc: @surfnet-niels), the suggested changes regarding a cut off point for showing IdPs on the WAYF have been implemented. You might want to check how everything looks with this solution.

To me, this seems like the way to go: the searchbox is more intuitive than searching through a humongous list of IdPs, yet the list is shown when there are less IdPs than a certain configured amount.

Issue #293, opened 2016-05-25
Detect browser locale setting to determine default language


arnoutt:
Upon first visit, the locale setting of the user's browser should be detected to determine the language of the page. Currently in SURFconext, only Dutch and English are supported. If a Dutch locale is detected, the page should be rendered in NL; otherwise (if any other language is detected of no language is detected), the page should be rendered in EN.

The language setting should be remembered (in a cookie?) and should be used on all pages (WAYF, consent, profile, error messages). If a user manually switches to another language, this should obviously be remembered as well.

thijskh:
Indeed use `Accept-Language` to determine the default/initial language. Because this information is available in every browser request, we need not store it anywhere.

I would only store it in a cookie if the user manually selects a different language, as an override.

arnoutt:
+1

thijskh:
Fixed in 5.3

PR #338, opened 2016-09-21
Update Bootstrapper.php

mrvanes:
Add browser locale detection.

### DIFF
--- a/library/EngineBlock/Application/Bootstrapper.php
+++ b/library/EngineBlock/Application/Bootstrapper.php
@@ -179,6 +179,22 @@ class EngineBlock_Application_Bootstrapper
 
         $this->_application->setLayout($layout);
     }
+    
+    protected function _getPreferredLanguage($available_languages)
+    {
+        $accepted_languages = explode(',', $_SERVER['HTTP_ACCEPT_LANGUAGE']);
+        $preferred_language = $available_languages[0];
+
+        foreach($accepted_languages as $idx => $lang) {
+            $lang = substr($lang, 0, 2);
+            if (in_array($lang, $available_languages)) {
+                $preferred_language = $lang;
+                break;
+            }
+        }
+        return $preferred_language;
+    }
+
 
     protected function _bootstrapTranslations()
     {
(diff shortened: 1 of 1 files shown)


library/EngineBlock/Application/Bootstrapper.php
```diff
@@ -220,8 +236,9 @@ protected function _bootstrapTranslations()
             $this->_application->getHttpResponse()->setCookie('lang', $lang, $cookieExpiry, '/', $cookieDomain);
         }
         else {
-            $translate->setLocale('en');
-            $this->_application->getHttpResponse()->setCookie('lang', 'en', $cookieExpiry, '/', $cookieDomain);
+            $preferredLanguage = $this->_getPreferredLanguage(array_keys($translationFiles));
```
DRvanR: Hey @mrvanes , Symfony is now available and provides a lot of functionality like this out of the box. For instance, this could be simplefied to:
```
$preferredLanguage = $this->_application->getDiContainer()->getSymfonyRequest()->getPreferredLanguage(array_keys($translationFiles));
```

That also makes the additional method obsolete :smile:

thijskh:
This relates to #293.

library/EngineBlock/Application/Bootstrapper.php
DRvanR: Hrms, this is due to the fact that Symfony has distinct phases with different abilities w.r.t. what is or isn't available per phase (the mentioned scope).
This specific error is caused by the fact that the ApplicationSingleton of the Corto application is booted when the Symfony application boots. When booting, it also boots the Corto application, which in turn invokes this bootstrapper and thus attempts to access the request (in my suggestion). However, in the booting phase of Symfony, the Request is not yet available, that will only be available when actually handling the request.

I have no objection to keeping this as is then, as having to adjust the booting order or request availability is not worth the hassle at this point, nor is trying to delay the determination of the preferred language as that would also be more work.

DRvanR: The timing of the code is indeed a bit off (would rather run this after the boot sequence when actually handling the request), but nevertheless it is early enough for the rest of the application. The template rendering the WAYF has some, eh, convoluted logic for determining the locale to use for the different elements with different fallback strategies etc. Ideally this would be fixed with a single (request and user scoped) PreferredLocaleProvider of sorts, but that would be a bit much to do I reckon. Also note that that is planned to be done at a to be determined moment as part of the refactoring effort. As to when, I cannot give an answer, perhaps @thijskh has an idea?

thijskh: Sorry, really no concrete idea at this point

DRvanR:
This has been resolved in #405 - will be part of next release.

PR #342, opened 2016-09-26
Correct dumping trusted proxy ips configuration

arothuis:
This corrects the oversight when fixing the issue in #331, as reported by @tvdijen in the same PR.

### DIFF
--- a/bin/composer/dump-required-ini-params.sh
+++ b/bin/composer/dump-required-ini-params.sh
@@ -31,7 +31,7 @@ if (!is_array($trustedProxies)) {
 $ymlContent = array(
     'parameters' => array(
         'domain'                                                  => $config->get('base_domain'),
-        'trusted_proxies'                                         => $config->get('trustedProxyIps', array())->toArray(),
+        'trusted_proxies'                                         => $trustedProxies,
         'api.users.janus.username'                                => $config->get('engineApi.users.janus.username'),
         'api.users.janus.password'                                => $config->get('engineApi.users.janus.password'),
         'api.users.profile.username'                              => $config->get('engineApi.users.profile.username'),


arothuis:
@tvdijen, does this solve the issue? Thanks again for your feedback.

tvdijen:
I think that's what I did on my install to fix it. Let me verify this tonight.

tvdijen:
It's probably totally unrelated, and more likely a configuration issue on my end, but now I run into the following error that doesn't really make much sense to me.

> Updating the "app/config/functional_testing.yml" file
> > Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::buildBootstrap
> > Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache
> 
>   [Symfony\Component\DependencyInjectionExceptionRuntimeException]                                                                                   
>   A string value must be composed of strings and/or numbers, but found parameter "domain" of type NULL inside string value "engine-api.%domain%".
> Script Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache handling the post-install-cmd event terminated with an exception
> 
>   [RuntimeException]
>   An error occurred when executing the "'cache:clear --no-warmup'" command:
>     [Symfony\Component\DependencyInjection\Exception\RuntimeException]
>     A string value must be composed of strings and/or numbers, but found parameter "domain" of type NULL inside string value "engine-api.%domain%".

arothuis:
This is possibly related to this issue. The `domain` parameter should be set after the `dump-ini-parameters` script in this PR is run.
 
Could you verify `app/config/ini_parameters.yml` is actually created after running `composer install`?

tvdijen:
Yes, it is created:

```
# This file is auto-generated
parameters:
    domain: null
    trusted_proxies: {  }
    api.users.janus.username: null
    api.users.janus.password: null
    api.users.profile.username: null
    api.users.profile.password: null
    logger.channel: engineblock
    logger.fingers_crossed.passthru_level: NOTICE
    logger.syslog.ident: EBLOG
    database.host: null
    database.port: null
    database.user: null
    database.password: null
    database.dbname: null
    feature_eb_ldap_integration: false
    feature_eb_encrypted_assertions: true
    feature_eb_encrypted_assertions_require_outer_signature: true
    feature_api_metadata_push: false
    feature_api_consent_listing: false
    feature_api_metadata_api: false
    minimum_execution_time_on_invalid_received_response: 5000
```

tvdijen:
OK, I think I see where this is going wrong..
I used [this](https://github.com/OpenConext/OpenConext-Howto/blob/master/05_Engineblock.md) doc as a starting point. I knew in advance that it was written for EB4, but since I couldn't find a newer version I went on with it.

It tells me to:
`cp docs/example.engineblock.ini /etc/openconext/engineblock.ini`
And then add a 'hostname' variable to the file.
After renaming 'hostname' to 'base_domain', everything started to work.

An example-variable should probably be added to the example config-file

arothuis:
Thanks for getting to the bottom of this configuration issue.

For EB5.x, it is advised to use the installation instructions in the [README][1]. 
Similar instructions are given there, so feel free to submit an issue or pull request regarding updating `docs/example.engineblock.ini`.

[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/master/README.md#installation

PR #344, opened 2016-09-26
Do not send referrer header on POST to SP.

thijskh:
Some SP's claim their (e.g. Piwik) stats are 'polluted' by referrers from
engine.surfconext.nl when people have logged into the SP, which obscure
the 'real' external referrer for this user.

We might not have much to gain by sending this header. So we can avoid
doing so, helping these SP's at no cost to us (?).

NB: this only modifies the source theme, theme still needs to be compiled.


thijskh:
After internal discussion we concluded not to apply this. The problem is that IE/Edge do not support it, so it would work for 80% of the cases. We believe that it working sometimes and not others would confuse things even more. We'd rather have consistent, predictable behaviour.

PR #349, opened 2016-10-19
Add tests for shibmd:scope positive behaviour

thijskh:
The functionality was tested for logging warnings, but not whether it worked ok for the expected case (scope matches value, no warning is logged). This add that test both for normal comparison and for case-insensitive comparison. The latter now fails because saml-value-object needs to be fixed for that.


thijskh:
OpenConext/SamlValueObject#13 should fix the test failure.

thijskh:
Added a small fix for a corner case when dealing with multiple @ signs in the ePPN. This is mostly to enhance logging of these cases, because they should be discouraged anyway.

arothuis:
This looks fine to me. :ship:

PR #350, opened 2016-10-21
Add ARP API in order to let clients apply ARPs to a set of attributes

arothuis:
This implements OpenConext/OpenConext-Profile#51.

Relates to OpenConext/OpenConext-Profile#70.

### DIFF
--- a/src/OpenConext/EngineBlock/Service/MetadataService.php
+++ b/src/OpenConext/EngineBlock/Service/MetadataService.php
@@ -52,11 +52,18 @@ final class MetadataService
     }
 
     /**
-     * @param ServiceProvider $serviceProvider
+     * @param EntityId $entityId
      * @return null|AttributeReleasePolicy
+     * @internal param ServiceProvider $serviceProvider
      */
-    public function findArpFor(ServiceProvider $serviceProvider)
+    public function findArpForServiceProviderByEntityId(EntityId $entityId)
     {
+        $serviceProvider = $this->findServiceProvider($entityId);
+
+        if ($serviceProvider === null) {
+            return null;
+        }
+
         return $this->metadataRepository->fetchServiceProviderArp($serviceProvider);
     }
 }
--- a/src/OpenConext/EngineBlock/Service/MetadataService.php
+++ b/src/OpenConext/EngineBlock/Service/MetadataService.php
@@ -2,6 +2,7 @@
 
 namespace OpenConext\EngineBlock\Service;
 
+use OpenConext\Component\EngineBlockMetadata\AttributeReleasePolicy;
 use OpenConext\Component\EngineBlockMetadata\Entity\IdentityProvider;
 use OpenConext\Component\EngineBlockMetadata\Entity\ServiceProvider;
 use OpenConext\Component\EngineBlockMetadata\MetadataRepository\EntityNotFoundException;
@@ -48,5 +49,14 @@ final class MetadataService
         }
 
         return $serviceProvider;
+    }
+
+    /**
+     * @param ServiceProvider $serviceProvider
+     * @return null|AttributeReleasePolicy
+     */
+    public function findArpFor(ServiceProvider $serviceProvider)
+    {
+        return $this->metadataRepository->fetchServiceProviderArp($serviceProvider);
     }
 }
--- a/src/OpenConext/EngineBlock/Service/MetadataService.php
+++ b/src/OpenConext/EngineBlock/Service/MetadataService.php
@@ -54,7 +54,6 @@ final class MetadataService
     /**
      * @param EntityId $entityId
      * @return null|AttributeReleasePolicy
-     * @internal param ServiceProvider $serviceProvider
      */
     public function findArpForServiceProviderByEntityId(EntityId $entityId)
     {
--- a/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
@@ -12,10 +12,6 @@ use OpenConext\Value\Saml\EntityId;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 
-/**
- * @SuppressWarnings(PHPMD.CyclomaticComplexity) Extensive request validation
- * @SuppressWarnings(PHPMD.NPathComplexity) Extensive request validation
- */
 final class AttributeReleasePolicyController
 {
     /**
@@ -46,6 +42,9 @@ final class AttributeReleasePolicyController
     /**
      * @param Request $request
      * @return JsonResponse
+     *
+     * @SuppressWarnings(PHPMD.CyclomaticComplexity) Extensive request validation
+     * @SuppressWarnings(PHPMD.NPathComplexity) Extensive request validation
      */
     public function applyArpAction(Request $request)
     {
--- a/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
@@ -64,14 +64,7 @@ final class AttributeReleasePolicyController
 
         $releasedAttributes = [];
         foreach ($body['entityIds'] as $entityId) {
-            $serviceProvider = $this->metadataService->findServiceProvider(new EntityId($entityId));
-
-            if ($serviceProvider === null) {
-                continue;
-            }
-
-            $arp = $this->metadataService->findArpFor($serviceProvider);
-
+            $arp = $this->metadataService->findArpForServiceProviderByEntityId(new EntityId($entityId));
             $releasedAttributes[$entityId] = $this->arpEnforcer->enforceArp($arp, $body['attributes']);
         }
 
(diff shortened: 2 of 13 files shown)


src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
```diff
@@ -0,0 +1,80 @@
...
+     * @param Request $request
+     * @return JsonResponse
+     */
+    public function applyArpAction(Request $request)
+    {
+        if (!$this->authorizationChecker->isGranted('ROLE_API_USER_PROFILE')) {
+            throw new ApiAccessDeniedHttpException('Access to the ARP API requires the role ROLE_API_USER_PROFILE');
+        }
+
+        $body = json_decode($request->getContent(), true);
```
arothuis: We have a JSON helper class for this, but it converts JSON to objects. As the ARP enforcer expects arrays, using a native json_decode with some validation afterwards removes weird mapping logic from objects to arrays.

arothuis:
@rjkip, 🍗

src/OpenConext/EngineBlockBundle/Exception/InvalidArgumentException.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace OpenConext\EngineBlockBundle\Exception;
+
+use InvalidArgumentException as CoreInvalidArgumentException;
+
+class InvalidArgumentException extends CoreInvalidArgumentException
+{
+    public static function invalidType($expectedType, $propertyPath, $parameter)
```
rjkip: Why was this needed?

src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @var EngineBlock_Arp_AttributeReleasePolicyEnforcer
+     */
+    private $arpEnforcer;
+
+    public function __construct(
+        AuthorizationCheckerInterface $authorizationChecker,
+        MetadataService $metadataService,
+        EngineBlock_Arp_AttributeReleasePolicyEnforcer $arpEnforcer
```
rjkip: Is it project style to not alias classes with underscore namespacing?

src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
```diff
@@ -0,0 +1,80 @@
...
+    public function applyArpAction(Request $request)
+    {
+        if (!$this->authorizationChecker->isGranted('ROLE_API_USER_PROFILE')) {
+            throw new ApiAccessDeniedHttpException('Access to the ARP API requires the role ROLE_API_USER_PROFILE');
+        }
+
+        $body = json_decode($request->getContent(), true);
+
+        if (!is_array($body)) {
+            throw new BadApiRequestHttpException('Unrecognized structure for JSON: ' . var_export($body, true));
```
rjkip: This log message doesn't really help me towards a solution; what is wrong with the JSON structure, what is expected? What about `Root value submitted in request must be an array, got <type>`? Also, I'm not that fond of logging entire request bodies (we are dealing with a trusted user here, but still) and, implicitly, newlines.

src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
rjkip: Hmm, why not create a helper that returns arrays?

src/OpenConext/EngineBlock/Service/MetadataService.php
```diff
@@ -49,4 +50,13 @@ public function findServiceProvider(EntityId $entityId)
 
         return $serviceProvider;
     }
+
+    /**
+     * @param ServiceProvider $serviceProvider
+     * @return null|AttributeReleasePolicy
+     */
+    public function findArpFor(ServiceProvider $serviceProvider)
```
rjkip: Why not accept an ID? Now we need to retrieve the service provider itself before retrieving its ARP.

src/OpenConext/EngineBlockBundle/Resources/config/routing_api.yml
```diff
@@ -25,3 +25,9 @@ api_metadata_idp:
     methods: [GET]
     defaults:
         _controller: engineblock.controller.api.metadata:idpAction
+
+api_attribute_release_policy:
+    path:   /arp
+    methods: [POST]
```
thijskh: Why is this a POST? Does it modify any state?

src/OpenConext/EngineBlockBundle/Exception/InvalidArgumentException.php
arothuis: Because it wasn't there, yet there was a class depending on it. I got an error when the Exception could not be thrown.

src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
arothuis: Thanks for noticing this. Considering we have the old and the new EB, I will look into what to do regarding underscore namespacing for older components.

src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
arothuis: This was a mistake on my part. I used it when debugging / exploring the request handling and I did not create a proper Exception message afterwards.

src/OpenConext/EngineBlock/Service/MetadataService.php
arothuis: Nice idea, that would reduce a call to the service. We could give it the entityId and let this method deal with the rest.

src/OpenConext/EngineBlockBundle/Resources/config/routing_api.yml
arothuis: After giving it some more thoughts and exchanging thoughts with colleagues, I think POST is appropriate. In principle, GET is the correct thing to do when retrieving a resource. Here, we are not retrieving a resource but we are doing a non-idempotent RPC-call: clients are basically saying "EngineBlock, please process this data for us and return the result".

thijskh: :+1:

src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
arothuis: For future reference: we are still using underscore namespacing for using older components. I think it makes sense to make it as explicit as possible whether you're dealing with some kind of compatibility component or not.

arothuis:
The meat of this PR, the ARP enforcer, is already unit tested. As it's an RPC-type API, it would make sense to create a component test at the controller-level. PHPUnit with some mocks would suffice IMHO.

src/OpenConext/EngineBlock/Service/MetadataService.php
```diff
@@ -49,4 +50,20 @@ public function findServiceProvider(EntityId $entityId)
...
+    /**
+     * @param EntityId $entityId
+     * @return null|AttributeReleasePolicy
+     * @internal param ServiceProvider $serviceProvider
+     */
+    public function findArpForServiceProviderByEntityId(EntityId $entityId)
+    {
+        $serviceProvider = $this->findServiceProvider($entityId);
+
+        if ($serviceProvider === null) {
```
arothuis: This is an edge case: our client (Profile) sends us a list of entityIds he retrieved earlier, but in the meantime the service provider could have been removed.

The code here means that if we cannot find a service provider, we assume there is no ARP. This causes us to show all attributes being released for this service provider, even though that might not have been the case. How should EngineBlock deal with this?

Thoughts, @thijskh?

thijskh: So this happens when there's a race condition: the SP was present when the list of SP's was retrieved, but removed before the ARP was applied. This is very rare and may in fact never happen in practice. So I'm fine with a simple approach that assumes no ARP.

src/OpenConext/EngineBlock/Service/MetadataService.php
```diff
@@ -49,4 +50,20 @@ public function findServiceProvider(EntityId $entityId)
 
         return $serviceProvider;
     }
+
+    /**
+     * @param EntityId $entityId
+     * @return null|AttributeReleasePolicy
+     * @internal param ServiceProvider $serviceProvider
```
arothuis: Remove.

tests/unit/OpenConext/EngineBlockBundle/Controller/AttributeReleasePolicyControllerTest.php
```diff
@@ -0,0 +1,385 @@
...
+    {
+        $this->expectException(BadApiRequestHttpException::class);
+        $this->expectExceptionMessage('expected decoded root value to be an array');
+
+        $request = $this->createRequestWithContent('null');
+
+        $arpController = new AttributeReleasePolicyController(
+            $this->mockAuthorizationChecker(),
+            $this->createDummyMetadataService(),
+            new EngineBlock_Arp_AttributeReleasePolicyEnforcer
```
rjkip: We're this :ok_hand: far away from a unit test. But on the other hand, I'm interested in whether this works on the application level. Maybe it's worth looking into how feasible a WebTestCase is (timebox).

arothuis:
I'm afraid we have to save the WebTestCase for another time. @rjkip, is this mergeable? :medal_military:

PR #352, opened 2016-11-15
Add new style database-configuration to docs

tvdijen:
The example engineblock.ini file contains database-configuration the old style.
When using this file as an example for a manual install, you run into all kinds of database-related errors (like: hostname cannot be null).
This PR should prevent putting manual installers on the wrong track.

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -146,9 +146,9 @@ ldap.baseDn             = "dc=surfconext,dc=nl"
 ;;;;;;;;;;;; DATABASE SETTINGS ;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-database.host = db.demo.openconext.org
+database.host = "localhost"
 database.port = 3306
-database.password = ""
+database.password = "secret"
 database.user = "engineblock"
 database.dbname = "engineblock"
 
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -146,23 +146,11 @@ ldap.baseDn             = "dc=surfconext,dc=nl"
 ;;;;;;;;;;;; DATABASE SETTINGS ;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-; Database masters are servers that can used for reading & writing
-; Database slaves are servers that can used only for reading
-; Define your servers on database.servername
-; then add them to the proper pool, like database.masters[] = "servername"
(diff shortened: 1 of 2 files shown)


arothuis:
Could you provide some more context in your PR's description? 

I cannot check this right now, but was the earlier syntax not working?
The new syntax is used in [OpenConext-deploy](https://github.com/OpenConext/OpenConext-deploy/blob/895afbe15c311ffbfd9dff23a2dc68c138a9647a/roles/engineblock/templates/engineblock.ini.j2#L63-L67) as well, but the old one is still used in [EngineBlock's application.ini](https://github.com/OpenConext/OpenConext-engineblock/blob/master/application/configs/application.ini#L145-L165).

DRvanR:
> I cannot check this right now, but was the earlier syntax not working?
It should not :wink: 

> The new syntax is used in OpenConext-deploy as well, but the old one is still used in EngineBlock's application.ini.

Then that should be changed as well, as that is no longer supported, see [UPGRADING][1].

[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/master/UPGRADING.md#database-configuration

tvdijen:
I've added some more context and added a commit for the application/configs/application.ini as well

application/configs/application.ini
```diff
@@ -146,23 +146,11 @@ ldap.baseDn             = "dc=surfconext,dc=nl"
...
-database.master1.password = ""
-;database.master1.use_persistent = true
-;database.masters[] = "master1"
-
-database.slave1.dsn = ""
-database.slave1.user = "engineblock"
-database.slave1.password = ""
-;database.slave1.use_persistent = true
-;database.slaves[] = "slave1"
+database.host = db.demo.openconext.org
```
arothuis: This is a bit different than we are using when a development VM is provisioned through the recommended [OpenConext-deploy](https://github.com/OpenConext/OpenConext-deploy/blob/c5d62d3fc5be8833152f1ac5fe0221529b354bf1/environments/vm/group_vars/vm.yml#L90-L94) process.

If possible, I would rather set the database.host to "localhost". As stated in [the documentation](https://github.com/OpenConext/OpenConext-engineblock#then-configure-application-config), these are meant to be the default values. 'Localhost' sounds like a sensible default to me.

The other file in this PR is meant for documentation which is probably why the `demo` notation was used.

tvdijen: I'll fix it.. Any default for the password too??  Like 'eb_password' or something?

arothuis: Thanks. I believe 'secret' is an OpenConext favourite. ;)

arothuis:
Looks nice. Thanks for your help!

application/configs/application.ini
thijskh: Yes, `secret` is our sample secret…

Issue #61, opened 2014-08-05
Show more informative error when IdP denies user access


surfnet-niels:
In case an IdP denies a user access to a resource, this may be signaled using:

samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Responder"
           samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:RequestDenied"

When this occurs, engine throws an error which is technically incorrect ("RequestDenied" is a valid SAML statement, not an error!) and is also not informative nor understandable by an enduser.

![error](https://cloud.githubusercontent.com/assets/1491826/3808789/0253153a-1c7c-11e4-84ea-96ace7b3a4bf.png)

In case urn:oasis:names:tc:SAML:2.0:status:RequestDenied is signaled, the enduser should be told to contact the IdP, as this is not a generic error, but a specific message that the user is denied access by the IdP

thijskh:
Apparently this does not work. An authentication response with the following status code:

```xml
    <samlp:Status>
        <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Responder">
            <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:RequestDenied" />
        </samlp:StatusCode>
    </samlp:Status>
```
generates "Error - An error occured".

thijskh:
The condition is detected:
```
Oct 23 15:00:28 t03 EBLOG[8427]: [2015-10-23 15:00:28] engineblock.NOTICE: Received an Error Response {"request_id":"562a2f6c0f431","session_id":"amUdW4jrkwYmRH9wcZkosiVDAU7"} {"exception_message":"Responder: SimpleSAML_Error_Exception: Something is wrong...","status":"urn:oasis:names:tc:SAML:2.0:status:Responder","substatus":null,"status_message":"SimpleSAML_Error_Exception: Something is wrong..."}
```

However the code defers to the Corto input filter for actually handling it. While calling the filter, EB bombs out on missing an assertion:
```
[...]

PR #355, opened 2016-11-22
Correct error message when response has status code

arothuis:
[127716439](https://www.pivotaltracker.com/story/show/127716439)
#61

This fixes several issues with the incorrect exceptions being thrown and the wrong feedback page being shown.

This also removes the validate successful response command from the Corto Input filter and moves it earlier in the process: in the Bindings module. This command caused errors when Corto receives a response without Assertions, even though that happens in the wild. Therefore, we want to throw the correct exception early so Corto does not have to deal with these kinds of issues.


arothuis:
Be sure to remove the Behat WIP-tags when this PR is ready to merge.

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/ResponseFactory.php
```diff
@@ -62,7 +62,11 @@ private function setResponseStatus(MockIdentityProvider $mockIdp, Response $resp
             $status['SubCode'] = $secondStatusCode;
         }
 
-        $status['Message'] = $mockIdp->getStatusMessage();
+        $statusMessage = $mockIdp->getStatusMessage();
+        if ($statusMessage !== null) {
+            $status['Message'] = $mockIdp->getStatusMessage();
```
rjkip: Probably just asking due to lack of context: shouldn't `$status['Message']` be set to an empty string when `$statusMessage` is null?

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/ResponseFactory.php
```diff
@@ -27,6 +27,10 @@ public function createForEntityWithRequest(
 
         $this->encryptAssertions($mockIdp, $response);
 
+        if ($mockIdp->shouldNotSendAssertions()) {
+            $response->setAssertions([]);
```
rjkip: Just something that made me ask myself and attempt to figure out: why encrypt the assertions and *then* clear them?

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -332,6 +332,9 @@ public function receiveResponse()
 
             try {
                 $assertions = $className::processResponse($sspSpMetadata, $sspIdpMetadata, $sspResponse);
+            } catch (sspmod_saml_Error $exception) {
+                // Pass through, show specific feedback for responses with error status codes
+                throw $exception;
```
rjkip: What other types of exceptions than `sspmod_saml_Error` pass through here? Is 5aa11d31fe6acd2999692d2370d9600053fac572 still relevant? What do different types of exceptions result in? I think we should sprinkle some comments here and there :)

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -367,6 +366,28 @@ public function receiveResponse()
                     'status_message'    => $e->getStatusMessage(),
                 )
             );
+
+            $status = $sspResponse->getStatus();
```
rjkip: So the Corto output filter code has been moved here, because Corto expects an assertion. What was the reason for moving it here vs. "fixing" Corto to handle 0 assertions? Honest question, I have no clue :zipper_mouth_face:

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
```diff
@@ -16,27 +16,81 @@ Feature:
       And SP "Unconnected SP" uses a whitelist for access control
       And an unregistered Service Provider named "Unregistered SP"
 
+  @WIP
```
rjkip: Do these still need to be WIP?

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/ResponseFactory.php
arothuis: Initially, I had the check first but in the other actions there is some logic assuming there are assertions. This way, you don't have to think about that and you can just drop the assertions afterwards.

thijskh:
Would we want a database migration to drop the obsolete tables?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
arothuis: WIP tags will be removed once the PR has been approved.

thijskh:
:+1:

library/EngineBlock/Corto/Module/Bindings.php
arothuis: I have to admit, I have not done extensive research regarding all Exceptions that can be produced when `$className::processResponse` is executed. However, there are functional tests for (some of the) other Exceptions that still pass.

I'm not sure what comments should be added to make this better.

Either way, this is a solid candidate for refactoring when the [feedback mechanism is reworked](https://www.pivotaltracker.com/n/projects/1425900/stories/107565968).

library/EngineBlock/Corto/Module/Bindings.php
rjkip: Hmm, but it didn't deal with any assertions in `ValidateSuccessfulResponse` :wink:

Very well.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -12,11 +12,13 @@ public function serve($serviceName)
...
         $log = $application->getLogInstance();
 
-        $log->notice(sprintf(
-            'Received Assertion from Issuer "%s" with signature method algorithm "%s"',
-            $receivedResponse->getIssuer(),
-            $receivedResponse->getAssertion()->getSignatureMethod()
-        ));
+        if ($receivedResponse->hasAssertion()) {
+            $log->notice(sprintf(
+                'Received Assertion from Issuer "%s" with signature method algorithm "%s"',
```
rjkip: What about logging when the response contains *no* assertions?

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/ResponseFactory.php
rjkip: Ah, so `$status['Message']` isn't present, initially? It's the modification of an existing status that makes me ask these things. Wouldn't it be prudent to create a new status array?

library/EngineBlock/Corto/Module/Bindings.php
arothuis: That is correct. It happens before the actual filter is called.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
arothuis: That will already be logged somewhere else when it, inevitably, fails because there basically is an assumption that there is an assertion. In other words, this check can be removed.

I think the validation of responses and the handling of exceptions should be improved during refactoring.

arothuis:
This has been approved IRL.

Issue #312, opened 2016-07-29
Remove broken VO functionality


thijskh:
In 5.0 the VO functionality has been broken. However, it has already been obsoleted by the PEP functionality introduced in 4.7. Simplest is just to remove it entirely instead of wasting time repairing it.

arothuis:
This has been removed in #365

PR #356, opened 2016-11-22
Remove Virtual Organization (VO) functionality completely

arothuis:
[118568103](https://www.pivotaltracker.com/story/show/118568103)
#312

VOs are broken in their current state, removing the functionality is more prudent than attempting to fix it at this time.


thijskh:
Would we want a database migration to drop the obsolete tables?

arothuis:
I don't think removing the VO's from the database as well would hurt.
If possible, I will try and leverage Doctrine for these migrations (instead of the older `dbpatch.php`).

PR #360, opened 2016-11-28
Replace @-sign in collab person ids for compatibility

arothuis:
[134915765](https://www.pivotaltracker.com/story/show/134915765)

In the past, at-signs were replaced by underscores because LDAP had trouble dealing with them. Recently, the way collabPersonIds are generated has changed, breaking compatibility with earlier versions of EB across the platform. In order to reinstate this, a method has been added to reinstate compatibility with all the collabPersonIds present in the platform already.

I have deliberately chosen to create another method for this to be explicit about the quirky implementation of collabPersonIds.

In the future, a way of migrating from old to new collabPersonIds should be devised. Then, we can use the original `CollabPersonId::generateFrom()` method.

### DIFF
--- a/src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
+++ b/src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
@@ -22,21 +22,6 @@ final class CollabPersonId
     private $collabPersonId;
 
     /**
-     * @param Uid                   $uid
-     * @param SchacHomeOrganization $schacHomeOrganization
-     * @return CollabPersonId
-     */
-    public static function generateFrom(Uid $uid, SchacHomeOrganization $schacHomeOrganization)
-    {
-        $collabPersonId = implode(
-            ':',
-            [self::URN_NAMESPACE, $schacHomeOrganization->getSchacHomeOrganization(), $uid->getUid()]
-        );
-
-        return new self($collabPersonId);
-    }
-
-    /**
      * This method solely exists for compatibility between EB versions and existing SPs.
      * Replacing the @-sign for underscores in collabPersonIds was part of the LDAP module.
      *
--- a/src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
+++ b/src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
@@ -72,7 +72,11 @@ final class CollabPersonId
             self::URN_NAMESPACE,
             sprintf('a CollabPersonId must start with the "%s" namespace', self::URN_NAMESPACE)
         );
-        Assertion::maxLength($collabPersonId, self::MAX_LENGTH, 'CollabPersonId length may not exceed 400 characters');
+        Assertion::maxLength(
+            $collabPersonId,
+            self::MAX_LENGTH,
+            sprintf('CollabPersonId length may not exceed %d characters', self::MAX_LENGTH)
+        );
 
         $this->collabPersonId = $collabPersonId;
     }
--- a/src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
+++ b/src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
@@ -37,6 +37,31 @@ final class CollabPersonId
     }
 
     /**
+     * This method solely exists for compatibility between EB versions and existing SPs.
+     * Replacing the @-sign for underscores in collabPersonIds was part of the LDAP module.
+     *
+     * See: https://www.pivotaltracker.com/story/show/134915765 and
+     * https://github.com/OpenConext/OpenConext-engineblock/commit/e6631acd4c4299329c5c34899de2f3a464975a5a
+     *
+     * @param Uid $uid
+     * @param SchacHomeOrganization $schacHomeOrganization
+     * @return CollabPersonId
+     */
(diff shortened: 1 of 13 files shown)


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AttributeManipulation.feature
```diff
@@ -26,7 +26,7 @@ Feature:
     When I give my consent
      And I pass through EngineBlock
     Then the url should match "functional-testing/SP-with-Attribute-Manipulations/acs"
-     And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="nl:surf:test:something"]/saml:AttributeValue["arbitrary-value"]'
+     And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="nl:surf:test:something"]/saml:AttributeValue[text()="arbitrary-value"]'
```
rjkip: What prompted this change?

src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
```diff
@@ -37,6 +37,31 @@ public static function generateFrom(Uid $uid, SchacHomeOrganization $schacHomeOr
...
+     * Replacing the @-sign for underscores in collabPersonIds was part of the LDAP module.
+     *
+     * See: https://www.pivotaltracker.com/story/show/134915765 and
+     * https://github.com/OpenConext/OpenConext-engineblock/commit/e6631acd4c4299329c5c34899de2f3a464975a5a
+     *
+     * @param Uid $uid
+     * @param SchacHomeOrganization $schacHomeOrganization
+     * @return CollabPersonId
+     */
+    public static function generateWithReplacedAtSignFrom(Uid $uid, SchacHomeOrganization $schacHomeOrganization)
```
rjkip: Because I'm not in this EB context that often, here are some of my understandings:

0. Previously, all collabPersonIds passed through LDAP.
0. To allow storage of said IDs, their @ signs were replaced with underscores.
0. This fact applied to all configurations of EB, including configuration of SPs and IdPs.
0. The feature toggle removed LDAP from the picture, suddenly causing @ signs to appear in collabPersonIds.
0. This change moves the @-sign replacement from the integration with LDAP to the "generation phase" of collabPersonIds.
0. There is currently no place in EB for collabPersonIds with @ signs in them.

Correct?

Also:

0. Hasn't LDAP been disabled for some time now? How did this only come up today? Shouldn't this affect all SPs?
0. I don't say this often, but I think this explicit method name is not needed. It generates collabPersonIds exactly in the format that was stored and sent to SPs as before. Also, I have a feeling that will stay the accepted algorithm for a long time to come, or am I wrong there? As for the comprehensive method description: :100:.
0. Will the migration to @-sign collabPersonIds be a thing? Is there some ticket to dump information in? Perhaps it would be nice to have a reference to that as well (or as the only reference).
0. Could you unit test this?

src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
```diff
@@ -47,7 +72,11 @@ public function __construct($collabPersonId)
             self::URN_NAMESPACE,
             sprintf('a CollabPersonId must start with the "%s" namespace', self::URN_NAMESPACE)
         );
-        Assertion::maxLength($collabPersonId, self::MAX_LENGTH, 'CollabPersonId length may not exceed 400 characters');
+        Assertion::maxLength(
```
rjkip: Nice catch :basketball_man:

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AttributeManipulation.feature
arothuis: Before the change it disregarded the `["arbitrary-value"]`, afterwards it enforces it.

rjkip: > it disregarded

Nice... :unamused:

src/OpenConext/EngineBlock/Authentication/Value/CollabPersonId.php
arothuis: 0. I'm not sure but it's probably because there were no tests for this and it probably took a while before the LDAP feature toggle was extensively tested.
1. I've discussed this with SURF. There are plans to make collabPersonIds behave as expected.
2. Although there are plans, these are not concrete. Somewhere in 2017.
3. Certainly.

thijskh: 0. Due to some dependencies, LDAP feature toggle was only disabled on prod 1w ago. Not many idp's have `@` in uid, 12 in total. So the problem was not found in testing because this case was not part of tests. It was rolled back later that day to fix the 12 idp's.
1. There's a concrete plan for another approach which we will schedule early 2017.
2. Likely not.

arothuis: Conclusion: for now, we want the collabPersonIds to behave as they once did.

thijskh:
My tests confirm to fix the issue for me :+1:

PR #362, opened 2016-12-02
Update simplesamlphp/saml2 dependency

arothuis:
This PR updates the SAML2 dependency in order to have the security fix for signature validation noted here: simplesamlphp/saml2#81

This is meant to be cherry-picked into a backported release as well.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "openconext/stoker-metadata": "~0.1",
         "openid/php-openid": "dev-master#ee669c6a9d4d95b58ecd9b6945627276807694fb as 2.2.2",
         "pimple/pimple": "~2.1",
-        "simplesamlphp/saml2": "1.10 as 0.8.1",
+        "simplesamlphp/saml2": "1.10.3 as 0.8.1",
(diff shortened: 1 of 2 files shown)


arothuis:
@rjkip 🐔 🍳 :eye:

composer.json
```diff
@@ -21,7 +21,7 @@
         "openconext/stoker-metadata": "~0.1",
         "openid/php-openid": "dev-master#ee669c6a9d4d95b58ecd9b6945627276807694fb as 2.2.2",
         "pimple/pimple": "~2.1",
-        "simplesamlphp/saml2": "1.10 as 0.8.1",
+        "simplesamlphp/saml2": "1.10.3 as 0.8.1",
```
rjkip: [This change](https://github.com/simplesamlphp/saml2/compare/v1.10...v1.10.3#diff-35f750cac20c45bceb545dd2e20dae85L1500) does not impact EB?

arothuis: All our tests are still going strong, so my first assumption would be that it does not really impact EB. However, it *is* worth the effort to have a look how EPTIs are currently used.

Issue #206, opened 2015-11-04
Detect and error out on infinite loop


relaxnow:
Sometimes SPs will infinitely redirect a user to EB, EB should detect this and show the user that something is wrong with the SP to try again or contact the SP.

Related: https://svn.surfnet.nl/svn/aai/sfs/phpFederate/trunk/lib/session.php line 170

thijskh:
This would be a nice enhancement indeed.

thijskh:
The svn server that @relaxnow references has since been shut down, but the linked code fragment is:

```php
         } else {
                    $_SESSION['dos-protect'][] = time();
                    $now = time();
                    while (($_SESSION['dos-protect'][0]) < ($now - 15)) array_shift($_SESSION['dos-protect']);
                    if (count($_SESSION['dos-protect']) > 5) {
                            throw new Exception('SSO loop detected: probably the service you are accessing is misconfigured!');
                    }
                    /* SSO succeeded: return to the caller SP */
                    session_return('updated');
         }
```

DRvanR:
Resolved through #365

PR #365, opened 2016-12-06
Create mechanism to deal with infinite authentication loops

arothuis:
[127716701](https://www.pivotaltracker.com/story/show/127716701)
#206 

This PR introduces a mechanism to track the authentication state and the history of authentication states. This way, we can determine whether we are in an authentication loop within our session for a specific service provider. The authentication state object can be used in the future to store more information about the request, which is currently done by the ProxyServer, Corto and the rest of EngineBlock. 

__Related__
OpenConext/OpenConext-deploy#117

__Tasks__:
- [x] functional test: requires a way to configure the authentication loop detection parameters
- [x] update OpenConext-deploy: add authentication loop detection parameters
-  ~~add a mockable time provider to rely on instead of hardcoded DateTimes~~ (other PR)


src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
```diff
@@ -0,0 +1,88 @@
...
+        $authenticationState = $this->session->get('authentication_state');
+
+        $stuckInAuthenticationLoop = $authenticationState->isInLoop(
+            $serviceProvider,
+            $this->maximumAuthenticationCyclesAllowed,
+            $this->timeFrameForAuthenticationLoopInSeconds
+        );
+
+        if ($stuckInAuthenticationLoop) {
+            throw new StuckInAuthenticationLoopException(
```
DRvanR: Please add to the Refactor Exception Listener issue that this should be refactored as well - it is a known case we actively check for, thus a proper Request->Response communication should be used.
No need to do that now as the current convention is to use exceptions for this.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $this->session                                 = $session;
+        $this->maximumAuthenticationCyclesAllowed      = $maximumAuthenticationCyclesAllowed;
+        $this->timeFrameForAuthenticationLoopInSeconds = $timeFrameForAuthenticationLoopInSeconds;
+    }
+
+    /**
+     * @param Entity $serviceProvider
+     */
+    public function assertNotStuckInLoop(Entity $serviceProvider)
```
DRvanR: s/assert/ensure/ ?

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
```diff
@@ -0,0 +1,88 @@
...
+            $serviceProvider,
+            $this->maximumAuthenticationCyclesAllowed,
+            $this->timeFrameForAuthenticationLoopInSeconds
+        );
+
+        if ($stuckInAuthenticationLoop) {
+            throw new StuckInAuthenticationLoopException(
+                sprintf(
+                    'After %d authentication cycles, we determined within a time frame of %d seconds'
+                    . 'that we are stuck in an authentication loop for service provider "%s"',
```
thijskh: Nit: this message misses a space between "seconds" and "that".

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
```diff
@@ -0,0 +1,84 @@
...
+    public function ensureNotStuckInLoop(
+        Entity $serviceProvider,
+        AuthenticationProcedureList $pastAuthenticationProcedures
+    ) {
+        $dateTime  = new DateTimeImmutable;
+        $startDate = $dateTime->modify(sprintf('-%d seconds', $this->timeFrameForAuthenticationLoopInSeconds));
+
+        $relevantProceduresInTimeFrame = $pastAuthenticationProcedures
+            ->findOnBehalfOf($serviceProvider)
+            ->findProceduresCompletedAfter($startDate);
```
DRvanR: Food for thought: since the returned result is apparently countable, how about calling `->count()` as last method? Then we can name `$relevantProceduresInTimeFrame` something like `$processedLoginProcedures`, which makes line 70 a bit more readable. And removed TimeFrame from the wording - which is not something we're actually using here, we're merely using points in time.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
```diff
@@ -0,0 +1,84 @@
...
+
+        $stuckInAuthenticationLoop = count($relevantProceduresInTimeFrame) >= $this->maximumAuthenticationProceduresAllowed;
+
+        if ($stuckInAuthenticationLoop) {
+            throw new StuckInAuthenticationLoopException(
+                sprintf(
+                    'After %d authentication procedures, we determined within a time frame of %d seconds'
+                    . ' that we are stuck in an authentication loop for service provider "%s"',
```
DRvanR: Nit:L are we stuck, or is the user stuck? 
Furthermore, seems the wording could be improved, something like: `More than "%d" authentication procedures for the current user from SP "%s" occurred within "%d" seconds, the user seems to be stuck in an authentication loop. Aborting the current authentication procedure.`

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+/**
+ * Copyright 2016 SURFnet B.V.
```
DRvanR: Either we add this to all files, or to none 😛

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuardInterface.php
```diff
@@ -0,0 +1,17 @@
...
+use OpenConext\Value\Saml\Entity;
+
+interface AuthenticationLoopGuardInterface
+{
+    /**
+     * @param Entity $serviceProvider
+     * @param AuthenticationProcedureList $pastAuthenticationProcedures
+     * @return
+     */
+    public function ensureNotStuckInLoop(
```
DRvanR: `ensure`, `assert`? `verified`?

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
```diff
@@ -0,0 +1,84 @@
...
+ */
+
+namespace OpenConext\EngineBlockBundle\Authentication;
+
+use DateTimeImmutable;
+use OpenConext\EngineBlock\Assert\Assertion;
+use OpenConext\EngineBlockBundle\Exception\StuckInAuthenticationLoopException;
+use OpenConext\Value\Saml\Entity;
+
+final class AuthenticationLoopGuard implements AuthenticationLoopGuardInterface
```
DRvanR: The more I think about it, the less I think we should name this a Guard. A Guard verifies something and returns a boolean expression with respect to whether or not the current execution branch may be continued (from the top of my head, too lazy to search for the official definition now). That is not what this does, this evaluates _and_ enforces. So either we remove the enforcing part (the throwing of the exception), or we should not call it a guard. Thoughts?

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationProcedure.php
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public static function onBehalfOf(Entity $serviceProvider)
+    {
+        return new self($serviceProvider);
+    }
+
+    /**
+     * @param Entity $identityProvider
+     */
+    public function authenticateAt(Entity $identityProvider)
```
DRvanR: shouldn't VO's be immutable? 😉  - in any case, lesser concern, perhaps a minor tech debt issue so that we don't forget that when we start using this more extensively.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationProcedureList.php
```diff
@@ -0,0 +1,96 @@
...
+        $newAuthenticationProcedures[] = $authenticationProcedure;
+
+        return new self($newAuthenticationProcedures);
+    }
+
+    /**
+     * @param Entity $entity
+     * @return AuthenticationProcedureList
+     */
+    public function findOnBehalfOf(Entity $entity)
```
DRvanR: filter rather than find? Not sure, just throwing out thoughts :)

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -0,0 +1,79 @@
...
+     */
+    public function authenticateAt(Entity $identityProvider)
+    {
+        $this->currentAuthenticationProcedure->authenticateAt($identityProvider);
+    }
+
+    /**
+     * @return void
+     */
+    public function completeCurrent()
```
DRvanR: `completeCurrent` implies I know what the current something is inside and `authenticationState`. `completeCurrentProcedure` tells me what I am completing.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->authenticationProcedures = $this->authenticationProcedures->add($this->currentAuthenticationProcedure);
+    }
+
+    /**
+     * @param Entity $identityProvider
+     * @return void
+     */
+    public function authenticateAt(Entity $identityProvider)
+    {
+        $this->currentAuthenticationProcedure->authenticateAt($identityProvider);
```
DRvanR: this (and the completion) cannot and may not be invoked when no current auth procedure is available. Perhaps we might want to add a guard clause for that?

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @return void
+     */
+    public function completeCurrent()
+    {
+        $this->currentAuthenticationProcedure->completeOn(new DateTimeImmutable);
```
DRvanR: Shouldn't this also remove the current Auth Procedure as it is completed?

languages/nl.php
```diff
@@ -247,6 +247,10 @@
     'error_received_status_code_desc'=> '<p>
         Jouw instelling stuurde een ongeldig authenticatierespons terug.
     </p>',
+    'error_stuck_in_authentication_loop' => 'Error - Je zat vast in een zwart gat',
+    'error_stuck_in_authentication_loop_desc' => '<p>
+        Je bent succesvol ingelogd bij je Identity Provider maar de dienst waar je naartoe wilt stuurt je weer terug naar SURFconext. Omdat je succesvol bent ingelogd, stuurt SURFconext je weer naar de dienst, wat resulteert in een oneindig zwart gat. Dit komt waarschijnlijk door een foutje aan de kant van de dienst. Bezoek <a href="https://support.surfconext.nl" target="_blank">de SURFconext support pagina\'s</a> voor meer ondersteuning bij deze foutmelding.
```
arothuis: Yep. They are written by SURF.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -1,9 +1,38 @@
 <?php
 
 use OpenConext\Component\EngineBlockMetadata\Entity\AbstractRole;
+use OpenConext\Value\Saml\Entity;
+use OpenConext\Value\Saml\EntityId;
+use OpenConext\Value\Saml\EntityType;
+use Symfony\Component\HttpFoundation\Session\Session;
 
-class EngineBlock_Corto_Module_Service_AssertionConsumer extends EngineBlock_Corto_Module_Service_Abstract
```
arothuis: Good point, will check it out. I might keep it just to be sure.

library/EngineBlock/Corto/Module/Services.php
```diff
@@ -89,6 +89,12 @@ private function factoryService($className, EngineBlock_Corto_ProxyServer $serve
                     $diContainer->getMailer(),
                     $preferredNameAttributeFilter
                 );
+            case EngineBlock_Corto_Module_Service_AssertionConsumer::class :
```
arothuis: Sounds like a plan. Will add an issue for it.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php
arothuis: I can remember we had a discussion about this a while ago and we for some reason settled for Guard, although I feel like its intent is not clear at once. Perhaps something obvious like `AuthenticationLoopPrevention` could work (perhaps suffixing `Service` to make it sound less passive).

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationProcedure.php
arothuis: This would not be hard to fix, though, would it?

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
arothuis: I had to remove extending the abstract in order to be able to inject the session into the constructor.

arothuis:
Hmm. Something weird went wrong when merging master. Will investigate.

Issue #316, opened 2016-08-01
consent_date and usage_date updated on login


thijskh:
In the `consent` table, both `consent_date` _and_ `usage_date` are updated on login, also when the user had previously accessed the service.

In any case `consent_date` should not be updated when logging in to something that one has already given consent for, so that's a bug to be fixed. I wonder though if we should keep `usage_date` at all. This causes one write per login to our database. Nicer approaches are possible that can avoid this number of writes.

This has been observed in 4.7 and 5.0 (but may be the case much longer already).

thijskh:
`consent` should always only store the date of (first) consent, no longer the date of last use.

DRvanR:
not storing the `last_usage` could make it hard to detect consents that have not been used for X amount of time. Could for instance be relevant if a given consent expires (or could expire) after a certain period. I don't think something like that exists at the moment, but removing the field removes that possibility as well. Not sure if that is even needed/wanted, but wanted to point it out just to be sure.

thijskh:
Thanks. We're aware of that.

arothuis:
Done in #367 and OpenConext/OpenConext-profile#73

PR #367, opened 2016-12-09
Drop usage date and only update consent once

arothuis:
[130002585](https://www.pivotaltracker.com/story/show/130002585)
OpenConext/OpenConext-engineblock#316

This PR prevents consent_date from updating upon each consent and removes the tracking of the last date on which consent has been given. The usage_date column has been removed from the consent table.

This impacts Profile, as a consumer of the Consent API, see: OpenConext/OpenConext-profile#73

### DIFF
--- a/src/OpenConext/EngineBlock/Authentication/Dto/Consent.php
+++ b/src/OpenConext/EngineBlock/Authentication/Dto/Consent.php
@@ -67,7 +67,6 @@ final class Consent
         return [
             'service_provider' => $serviceProvider,
             'consent_given_on' => $this->consent->getDateConsentWasGivenOn()->format(DateTime::ATOM),
-            'last_used_on'     => $this->consent->getDateLastUsedOn()->format(DateTime::ATOM),
             'consent_type'     => $this->consent->getConsentType()->jsonSerialize(),
         ];
     }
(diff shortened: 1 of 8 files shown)


arothuis:
Will wait before merging until OpenConext/OpenConext-profile#73 lands.

src/OpenConext/EngineBlock/Authentication/Dto/Consent.php
```diff
@@ -67,7 +67,6 @@ function (ContactPerson $contact) {
         return [
             'service_provider' => $serviceProvider,
             'consent_given_on' => $this->consent->getDateConsentWasGivenOn()->format(DateTime::ATOM),
-            'last_used_on'     => $this->consent->getDateLastUsedOn()->format(DateTime::ATOM),
```
arothuis: This is indeed a BC break, but it is reasonable to assume Profile is the only consumer for this client.

Do you suggest adding an update note somewhere?

arothuis: Discussed it with SURF, there are no other consumers.
I updated the docs for migrating to 5.2, mentioning the changes.

rjkip: :+1:

PR #369, opened 2016-12-16
Correct and comment out DB settings in ini

arothuis:
This ini overrides the engineblock.ini written during provisioning.

The previous values were added to use the correct syntax, but used the some default values that conflicted with the default OpenConext-deploy development environment, see https://github.com/OpenConext/OpenConext-engineblock/pull/352 by @tvdijen 

This change is transparent about how it can be configured (and how it is configured in OpenConext-deploy).


arothuis:
I stumbled upon this trying to get migrations to work correctly. I had a temporary fix in place that I forgot about, but this should fix it.

thijskh:
:+1:

PR #371, opened 2016-12-16
Set AllowCreate correctly for SAML2

arothuis:
The SAML2 version bump contains a change in the way AllowCreate is to be set in SAML2: https://github.com/simplesamlphp/saml2/commit/aecda70e72970c8d66af1d18d5b762a8df5731ed.

We were setting this as a string and this caused AllowCreate not to be set anymore, while there are some IdPs out there that require it.
This PR aims to fix that.

In order to properly test this and future requirements regarding AuthnRequests, this PR also introduces infrastructure to be able to query the encoded AuthnRequest XML present in the HTTP response's body.

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AuthnRequest.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AuthnRequest.feature
@@ -15,4 +15,4 @@ Feature:
     When I log in at "Dummy-SP"
      And I pass through EngineBlock
      And I pass through the IdP
-    Then the response should contain "AllowCreate"
+    Then the AuthnRequest should match xpath '/samlp:AuthnRequest/samlp:NameIDPolicy[@AllowCreate="true" or @AllowCreate="1"]'
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AuthnRequest.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AuthnRequest.feature
@@ -1,0 +1,18 @@
+@WIP
+Feature:
+  In order to maintain compatibility with certain IdPs
+  As Engineblock
+  I want to send correct AuthnRequests
+
+  Background:
+    Given an EngineBlock instance on "vm.openconext.org"
+      And no registered SPs
+      And no registered Idps
+      And an Identity Provider named "Dummy-IdP"
+      And a Service Provider named "Dummy-SP"
+
+  Scenario: The AuthnRequest contains the correct attributes
+    When I log in at "Dummy-SP"
+     And I pass through EngineBlock
+     And I pass through the IdP
+    Then the response should contain "AllowCreate"
--- a/tests/library/EngineBlock/Test/Corto/ProxyServerTest.php
+++ b/tests/library/EngineBlock/Test/Corto/ProxyServerTest.php
@@ -20,10 +20,13 @@ class EngineBlock_Test_Corto_ProxyServerTest extends PHPUnit_Framework_TestCase
             $proxyServer
         );
 
-        $expectedNameIdPolicy = ['AllowCreate' => true];
-        $actualNameIdPolicy   = $enhancedRequest->getNameIdPolicy();
+        $nameIdPolicy = $enhancedRequest->getNameIdPolicy();
 
-        $this->assertSame($expectedNameIdPolicy, $actualNameIdPolicy);
+        $this->assertNotContains(
+            'Format',
+            array_keys($nameIdPolicy),
+            'The NameIDPolicy should not contain the key "Format"'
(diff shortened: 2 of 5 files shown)


tests/library/EngineBlock/Test/Corto/ProxyServerTest.php
```diff
@@ -20,7 +24,9 @@ public function testNameIDFormatIsNotSetByDefault()
             $proxyServer
         );
 
-        $this->assertNotContains('Format', $enhancedRequest->getNameIdPolicy());
```
arothuis: Apparently (our version of) PHPUnit does not deal with a boolean of 'true' in the array when using NotContains. It was incorrectly asserting that it did contain Format while it did not.

Matching explicitly for the expected NameIdPolicy is clear and gets the job done.

arothuis:
Tagging as WIP, I would like some tests explicitly verifying the presence of the `AllowCreate` flag in the AuthnRequest.

arothuis:
I have added such a test. Sadly, we can only look inside the string representation. Soon (monday?), we should create a way to inspect the AuthnRequest using XPath.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AuthnRequest.feature
```diff
@@ -0,0 +1,18 @@
+@WIP
+Feature:
+  In order to maintain compatibility with certain IdPs
```
rjkip: Hmm, do we send correct AuthnRequests to "maintain compatibility with certain IdPs"? I'd also not name this feature `AuthnRequest`. I'd say the feature has something to do with authentication or authentication proxying (Aren't there already scenarios for this somewhere? Not that obvious since I can't find them at a glance :disappointed:) and one of the assertions is then that we allow IdPs to "create NameIDs". Thoughts?

tests/library/EngineBlock/Test/Corto/ProxyServerTest.php
```diff
@@ -9,6 +9,10 @@ class EngineBlock_Test_Corto_ProxyServerTest extends PHPUnit_Framework_TestCase
 {
     public function testNameIDFormatIsNotSetByDefault()
     {
+        $expectedNameIdPolicy = [
+            'AllowCreate' => true
+        ];
```
rjkip: Since this value is not part of the arrange phase of the test, but only relevant during the assertion phase, could you move this to the definition of the actual NameID policy? Or what's your opinion regarding that?

tests/library/EngineBlock/Test/Corto/ProxyServerTest.php
rjkip: 0. Was a weird assertion in any case... tested that `"Format"` wasn't in the _values_ of the policy, while it would appear in the _keys_?
0. Does this make a case for upgrading PHPUnit to fix the issue and/or reporting the bug?
0. This test's name doesn't seem to match the test implementation anymore.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/AuthnRequest.feature
thijskh: We do not necessarily need or want IdPs to create NameIDs (we don't care, really). The most compatible way of not caring is to tell IdPs that they can do everything they want with nameId (so to "allow create"). The goal of this test is to ensure that we always send `AllowCreate="true"` because it's the most compatible setting.

arothuis: The features regarding ARP and the AM explicitly test certain AuthnRequests as well (and proxying by EB). 
I used rather generic naming because I wanted an extendable and explicit feature to assert certain (technical!) expectations we (and clients) might have regarding AuthnRequests and how they are supposed to look. If we bind this to a specific functionality we lose this advantage, but we would have more functionally descriptive and "correct" feature specifications.

I'm having some trouble coming up with a more functionally descriptive solution that also offers easy extendability without fragmenting our assertions in separate scenario's or features. Input is very welcome.

arothuis:
Regarding the `assertNotContains`:
0. It was indeed flawed.
1. The underlying issue has been fixed this way. It is a good idea to check if this is, in fact, a PHPUnit bug (or something else was going on) and check if it still exists in newer versions of PHPUnit. It might take some work, though.
2. I agree that the test name no longer matches the actual assertion. I'd fix this by doing an `assertNotContains` on the `array_keys`.

arothuis:
This PR is temporarily frozen until a next release.

arothuis:
As discussed, instead of having to rollback the SAML2 library dependency for a 5.1.1 and a 5.2 release, we should merge this as it fixes the underlying issue.

PR #379, opened 2017-01-30
Correct the PDP client configuration

DRvanR:
This PR makes the PDP api host and path configurable without reusing a configuration setting that has been used prior to 5.2. At the same time, provides full configuration of the host and path, no more hardcoded paths :)

Should be merged after merging the supporting PR in OpenConext-deploy: https://github.com/OpenConext/OpenConext-deploy/pull/123


DRvanR:
@thijskh Thoughts?

PR #383, opened 2017-02-17
Re-enable EPPN subjecIdField choice

mrvanes:
valid values for subjectIdField are "uid+sho" for old uid & schacHomeOrganisation behaviour, and new eppn for EdupersonPrincipalName behaviour. eppn can't be enabled when feature ldap_integration is enabled.

### DIFF
--- a/library/EngineBlock/Corto/Filter/Command/ProvisionUser.php
+++ b/library/EngineBlock/Corto/Filter/Command/ProvisionUser.php
@@ -25,7 +25,13 @@ class EngineBlock_Corto_Filter_Command_ProvisionUser extends EngineBlock_Corto_F
     public function execute()
     {
         $userDirectory = EngineBlock_ApplicationSingleton::getInstance()->getDiContainer()->getUserDirectory();
-        $user = $userDirectory->identifyUser($this->_responseAttributes);
+
+        $subjectIdField = EngineBlock_ApplicationSingleton::getInstance()->getConfigurationValue(
+            'subjectIdAttribute',
+           'uid+sho' 
+        );
+
+        $user = $userDirectory->identifyUser($this->_responseAttributes, $subjectIdField);
 
         $collabPersonIdValue = $user->getCollabPersonId()->getCollabPersonId();
         $this->setCollabPersonId($collabPersonIdValue);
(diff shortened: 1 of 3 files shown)


DRvanR:
Heya,

First off, thanks for the PR. However, I'm sorry to say this PR raises some concerns. 

It seems to me that identifying users differently with severe restrictions based on features should be done using a different filter, which uses a different UserDirectory altogether. This should then be enabled (and inherently the current one disabled) based on a feature toggle. 
Using a toggle also allows the switching of the filter and adapter based on configuration rather than runtime considerations. One thing that stands out in this is the fact that you can choose for this EPPN if and only if LDAP is disabled; this would then also be caught at configuration time rather than in runtime. 
As an additional benefit, using a different filter and UserDirectory would make the planned refactoring of this code (the filters etc.) a lot simpler.

In general, duplicating the logic with respect to what identifies the user in several places and layers to me is not a good idea - that should be centralized in a single spot, this to prevent someone making changes in one place and not in the other due to simply not knowing about it. As a minor thing, using magic strings like "uid+sho" vs "eppn" should be avoided.

As an alternative to different filters (I'd strongly recommend to use feature toggles regardless), different concrete strategies could be used. Might make the solution a bit simpler - and more flexible for the future, one could define new strategies - at the cost of increased test complexity and variance. 

Thoughts @thijskh @quartje @arothuis ?

thijskh:
I'm thinking that we want to remove the LDAP backend in the near future. I see no reasons to keep it. The fact that it's currently togglable is for migration purposes, as far as I'm concerned. Once you consider LDAP deprecated, the problem that this patch is not compatible with it becomes much smaller.

thijskh:
@mrvanes This PR now has conflicts because LDAP code has been removed. Do you want ti create an updated (simpler, I think) PR? Or do you want to document the AM you can use to fix this in the EB wiki?

MKodde:
@mrvanes are you planning on taking on this issue? Otherwise I'd like to close it (feel free to reopen it at any time)

mrvanes:
I have no plans taking this up anymore, unless there's explicit request from SURFnet to fix this?

thijskh:
Let's close it then. The code has changed significantly, so if someone wants it, a new PR is probably best.

PR #384, opened 2017-02-17
Bugfix: PDP client configuration

### LINKED ISSUE #382: PDP client fails in 5.2: 'GuzzleHttp\Http\Client' not found
Invoking a policy decision on EB 5.2.0 fails with:

```
Fatal error: Class 'Guzzle\Http\Client' not found in /opt/openconext/OpenConext-engineblock-5.2.0/app/cache/prod/appProdProjectContainer.php on line 621

A very serious error occurred, it has been logged and sent to the administrator.

ERROR:

array (
  'type' => 1,
  'message' => 'Class \'Guzzle\\Http\\Client\' not found',
  'file' => '/opt/openconext/OpenConext-engineblock-5.2.0/app/cache/prod/appProdProjectContainer.php',
  'line' => 621,
)
```

arothuis:
Fixes #382

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Resources/config/services.yml
+++ b/src/OpenConext/EngineBlockBundle/Resources/config/services.yml
@@ -54,15 +54,13 @@ services:
             - "@engineblock.pdp.guzzle_http_client"
 
     engineblock.pdp.guzzle_http_client:
-        class: Guzzle\Http\Client
+        class: GuzzleHttp\Client
         arguments:
-            - "%pdp.host%"
-            - request.options:
-                auth: ["%pdp.username%", "%pdp.password%", "Basic"]
-                # Verify CAs for certificates for prod, but not for other environments
-                # as we are working with self signed signatures
-                # Sadly, Guzzle 3 does not support setting curl.cainfo in php.ini or 20-curl.ini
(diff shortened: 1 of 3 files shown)


src/OpenConext/EngineBlockBundle/Resources/config/services.yml
```diff
@@ -54,15 +54,13 @@ services:
...
-                auth: ["%pdp.username%", "%pdp.password%", "Basic"]
-                # Verify CAs for certificates for prod, but not for other environments
-                # as we are working with self signed signatures
-                # Sadly, Guzzle 3 does not support setting curl.cainfo in php.ini or 20-curl.ini
-                verify: "@=service('kernel').getEnvironment() === 'prod'"
+            - base_uri: "%pdp.host%"
+              auth: ["%pdp.username%", "%pdp.password%", "Basic"]
+              # Verify CAs for certificates for prod, but not for other environments
+              # as we are working with self signed signatures
+              verify: "@=service('kernel').getEnvironment() === 'prod'"
```
arothuis: Next step would be to let our system define this by [setting](http://php.net/manual/en/curl.configuration.php) `curl.cainfo` in `php.ini` or `20-curl.ini`. Not sure what it should point to, so using this intermediate solution for now.

Issue #380, opened 2017-02-01
debug page broken in EB 5.2 (authentication procedure has not been authenticated)


tvdijen:
After migrating to EB 5.2.0 I can't seem to get EB working anymore.
The migration itself went fine, no errors or anything.
However, when I try to use the debug-tool and authenticate against an IdP, I get the following error:

[...]

PR #385, opened 2017-02-20
Bugfix: debug page causes invalid authentication state

arothuis:
This PR solves #380.

Issues were caused by debug requests not registering authentication states correctly.

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
+++ b/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
@@ -70,6 +70,8 @@ class EngineBlock_Corto_Module_Service_AssertionConsumer implements EngineBlock_
         if ($receivedRequest->isDebugRequest()) {
             $_SESSION['debugIdpResponse'] = $receivedResponse;
 
+            // Authentication state needs to be registered here as the debug flow differs from the regular flow,
+            // yet the procedures for both are completed when consuming the assertion in the ServiceProviderController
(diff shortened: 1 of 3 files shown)


library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -69,7 +69,15 @@ public function serve($serviceName)
 
         if ($receivedRequest->isDebugRequest()) {
             $_SESSION['debugIdpResponse'] = $receivedResponse;
-            $this->_server->redirect($this->_server->getUrl('debugSingleSignOnService'), 'Show original Response from IDP');
+
+            $identityProvider = new Entity(new EntityId($idp->entityId), EntityType::IdP());
```
DRvanR: A small comment is to why this unusual code is introduced would be nice :wink:

src/OpenConext/EngineBlockBundle/Controller/DebugController.php
```diff
@@ -27,6 +39,9 @@ public function debugSpConnectionAction()
 
         $proxyServer->debugSingleSignOn();
 
+        $authenticationState = $this->session->get('authentication_state');
+        $authenticationState->startAuthenticationOnBehalfOf(new Entity(new EntityId('debug_sp'), EntityType::SP()));
```
DRvanR: Same here, please explain why we (have) to do this using a fictitious EntityID.

PR #387, opened 2017-02-21
Remove legacy 'selectedIdp' cookie.

thijskh:
This has been supplanted by 'selectedidps' (with `s`) introduced in EB 4.5.
However, the old cookie was not removed at that time.


arothuis:
This looks fine by me. Merging.

I think we have indeed replaced this functionality with the newer `selectedIdps` which is used to present a history of selected IdPs in the WAYF.

Issue #386, opened 2017-02-21
EB 5.1 stores selectedIdps only with Session length


thijskh:
Since the JS refactoring in EB 5.1, the `selectedidps` cookie is only set with a duration of "session". This means it will be forgotten if the user closes his/her browser. It used to have a lifetime of a year.

Compare:
https://github.com/OpenConext/OpenConext-engineblock/blob/5.0.2/theme/material/javascripts/application.js#L360
https://github.com/OpenConext/OpenConext-engineblock/blob/5.1.1/theme/material/javascripts/modules/PreviousSelectionStorage.js#L16

arothuis:
Closing as it has been implemented in 5.2.2 in #388 and backported to 5.1.2 in #390.

PR #388, opened 2017-02-21
Configure IDP selection cookie expiration

arothuis:
Solves #386 

Edit: also added the missing theme template for StuckInAuthenticationLoop. It was only registered as a compiled version, not as the template version.


DRvanR:
We might want to add a "how to compile the JS" section to the README for visibility?

arothuis:
I agree. However, there is a note in the README referring to the [wiki](https://github.com/OpenConext/OpenConext-engineblock/wiki/Development-Guidelines#theme-development). The information in the wiki seems up to date except for the fact that we no longer require bower.

thijskh:
Solves the problem for me :+1:

PR #389, opened 2017-02-25
No notice on ommiting protocolBinding

tvdijen:
As per mailinglist-discussion:
protocolBinding is an optional attribute according to the specs;  http://saml2int.org/profile/current/#section82


thijskh:
Looks good to me!

DRvanR:
@thijskh / @pmeulen can this be merged?

PR #393, opened 2017-03-21
Bugfix: JS and CSS compilation

arothuis:
This fixes an issue where vendor prefixes in compiled css were removed and updates the minified JavaScript because of changes in the underlying compilation/minification algorithm.


thijskh:
Alright, I can see that the webkit stuff is now back, good! Will ask affected developers to test.

thijskh:
Confirmed that it fixes the problem for embedded browser.

arothuis:
The build fails due to an unrelated issue, see: #394.

arothuis:
Merging as the failing build is caused by Travis and not by code changes on our side.

PR #394, opened 2017-03-21
Update travis apcu installation

arothuis:
* Specify manually installing APCU as this has been included in Travis (travis-ci/php-src-builder@a846a53)
* Remove 5.3 specific command from .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ before_install:
 
 before_script:
   - phpenv config-add .travis.php.ini
+  - pecl uninstall apcu
+  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then printf "\n" | pecl install apcu-4.0.10; fi
   - sudo ln -s "$(pwd)"/ci/travis/files/ /etc/openconext
   - mkdir -p /tmp/engineblock/cache/test
   - mkdir -p /tmp/engineblock/logs/test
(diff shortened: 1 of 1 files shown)


arothuis:
Interestingly, the build now fails on a single functional test: "The Service Provider can have an attribute added" (`AttributeManipulation.feature`).

arothuis:
It seems that Travis' upgrade from apcu 4.0.10 to 4.0.11 broke something. I'm not sure what it was as I'm not well-versed in C and PHP-extensions, but it could be caused by the following changes: https://github.com/krakjoe/apcu/compare/v4.0.10...v4.0.11

I have tested my assumption on a temp branch (#395), falling back to v4.0.10 of apcu and it seems to work.

arothuis:
The PHP 7 build now fails on a single functional test. This is probably because APCu is no longer compatible with APC in PHP 7. I have created an [issue](https://www.pivotaltracker.com/story/show/142383563) for that.

.travis.yml
```diff
@@ -32,13 +32,14 @@ before_install:
 
 before_script:
   - phpenv config-add .travis.php.ini
+  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then printf "\n" | pecl uninstall apcu; fi
```
arothuis: Sadly, I could not pull this off in a one liner. For instance the following failed because of Travis timing out:
```
if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then printf "\n" | pecl uninstall apcu && pecl install apcu-4.0.10; fi
```

PR #395, opened 2017-03-24
Temp: debug Travis build

arothuis:
This is a temporary PR to attempt fixing the failing Travis build.


arothuis:
Will port a0231b5 to PR #394 and hope it works

arothuis:
I've successfully debugged and fixed the issue -- and then some.

I think we can get the PHP 7 build working if we do one of the following:
* Require Doctrine Cache to 1.6, which supports ApcuCache + configure doctrine in app/config to use apcu
* Install a BC layer for APCu -> APC compatibility

The first approach would be a more stable solution. I could not get the second approach to work.

PR #397, opened 2017-03-27
Remove dead code from View.php

nicwortel:
https://www.pivotaltracker.com/story/show/142487713

The result of the expression isn't used or captured in a variable, so it's dead code and should be removed.

### DIFF
--- a/library/EngineBlock/View.php
+++ b/library/EngineBlock/View.php
@@ -133,17 +133,6 @@ class EngineBlock_View
     }
 
     /**
-     * Return the module name
-     *
-     * @return string
-     */
-    public static function moduleName()
-    {
-        $serverName = $_SERVER['SERVER_NAME'];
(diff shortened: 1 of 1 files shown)


tvdijen:
This was clearly meant as a fallback to English when language is not set.
I think it shouldn't be removed, but properly fixed..

thijskh:
Is `getDisplayName()` used at all?

arothuis:
@tvdijen 
> This was clearly meant as a fallback to English when language is not set.
I think it shouldn't be removed, but properly fixed.

Thanks for your reply.
The fact remains that it currently is dead code, so there is no harm in removing it. If there is an issue with (default) translations, we should document this and add an issue so it can be resolved accordingly.

@thijskh 
> Is getDisplayName() used at all?
Good question. A simple string search does not yield any results. The same can be said for `moduleName()`.

PR #340, opened 2016-09-23
Turn on security checker and do composer update

relaxnow:
At Ibuildings we would like to take a more proactive role at monitoring and alerting for security updates of the software we create and deliver to customers.

As part of this, for this project we'd like to:
- Upgrade to the latest patch level of Symfony 2.7
- Upgrade to the latest patch level of Zend Framework 1
- Upgrade to the latest minor level of simplesamlphp/simplesamlphp
- Turn on the security checker and make it fail pull requests.

Unfortunately I had to do a full composer update as Composer got lost with all the interdependencies of EB, SSP, SAML2 and XMLSecLibs.
Please see the commit for a full overview of all upgraded packages.

Also we'd like to turn on nightly builds ( https://docs.travis-ci.com/user/cron-jobs/ ), though I think I can request this from Travis.

PR #398, opened 2017-03-27
Update dependencies that contain security vulnerabilities

nicwortel:
Replaces #340, fixes https://www.pivotaltracker.com/story/show/142487853


composer.json
```diff
@@ -21,11 +21,11 @@
         "openconext/stoker-metadata": "~0.1",
         "openid/php-openid": "dev-master#ee669c6a9d4d95b58ecd9b6945627276807694fb as 2.2.2",
         "pimple/pimple": "~2.1",
-        "simplesamlphp/saml2": "1.10.3 as 0.8.1",
-        "simplesamlphp/simplesamlphp": "~1.13",
+        "simplesamlphp/saml2": "1.10.3 as 1.9.1",
```
thijskh: Excuse my ignorance, but what does this mean exactly? That we still use the same saml2 version as before? Or not?

thijskh:
simplesaml 1.13 -> 1.14 has quite some changes listed here https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-1.14
We should check whether any of that is relevant for us.

composer.json
nicwortel: It is known as a [Composer alias](https://getcomposer.org/doc/articles/aliases.md#require-inline-alias). It means that we are installing version 1.10.3, but are aliasing it as 1.9.1, to fix dependency issues between the simplesamlphp/saml2 package (which we need version 1.10 of) and the simplesamlphp/simplesamlphp package (which has a strict requirement on simplesamlphp/saml2 which doesn't allow version 1.10.3).

arothuis:
Ah, they moved the `ClockMock` in Symfony? That's why they could not be found! This reduces the amount of work that was necessary, but I was actually looking forward to having [131718959](https://www.pivotaltracker.com/n/projects/1425900/stories/131718959) fixed.

composer.json
```diff
@@ -21,11 +21,11 @@
...
+        "simplesamlphp/simplesamlphp": "^1.14",
         "sybio/image-workshop": "~2.0.7",
         "zendframework/zendframework1":"~1.12",
-        "symfony/symfony": "2.7.*",
+        "symfony/symfony": "^2.8",
```
arothuis: A minor version bump is kinda exciting. I think it is okay as SF (mostly) adheres to semver and our build is not failing. For changes between 2.7 to 2.8, have a look at the [upgrade notes](https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md)

nicwortel:
Well, the `ClockMock` in the PHPUnit bridge (introduced in v2.8.0) slightly differs from the one that was part of the Stopwatch component up until v2.7.4, but yes.

I think we should keep track of https://www.pivotaltracker.com/n/projects/1425900/stories/131718959 and see if it is still an issue. If it is, we might pick another solution.

composer.json
arothuis: As an aside, this upgrades the doctrine/cache to 1.6, meaning we can use the ApcuCache. This could mean [142383563](https://www.pivotaltracker.com/story/show/142383563) can be easily fixed.

We do have to check `apcu` is available in production.

nicwortel: Most changes are in the Form component, which doesn't seem to be used in Engineblock. We can expect some (Yaml) deprecation notices, but other than that we should be fine.

nicwortel:
It is probably a good idea to update other dependencies as well. See https://www.pivotaltracker.com/story/show/142582163.

arothuis:
It is not using the ClockMock properly. Register should be called on the classes/contexts that actually call time-based functions (`sleep`, `usleep`, `time`), so the `Stopwatch` and the integration test itself. Like you said IRL, the unit test does not need the ClockMock as the Stopwatch is mocked there.
Also, the ClockMock should be initiated as a mock by calling `withClockMock(<number>)` for each test.

Doing this solves the issue for me.

nicwortel:
We have checked that the upgrade to Symfony 2.8 doesn't generate a lot of deprecation notices, but it would be advisable to verify this on the accept environment as well.

arothuis:
:+1: Let's merge this. I have not found any issues with these upgrades.

PR #399, opened 2017-03-29
[PARKED] Use APCu for forward compatibility with PHP 7

arothuis:
[142383563](https://www.pivotaltracker.com/story/show/142383563)
Based on #398 

_WIP as long as this is not tested against an actual environment, as my VM has some issues_


arothuis:
I think we should park this issue, as we cannot verify that this works in PHP 7 without a PHP 7 development environment.

DRvanR:
What exactly causes the error in the functional tests? If it is doctrine+apc(u), we could simply switch to a file based cache driver for doctrine and check if that resolves the issue? See: http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/caching.html and https://github.com/doctrine/cache/tree/master/lib/Doctrine/Common/Cache for a list of available drivers.

arothuis:
@DRvanR,  I was trying to switch from the APC cache to the APCu cache, but without any success. 
In your remark, do you mean to only use a file-based cache driver for testing or for the actual application as well?
If the former, I'm worried about discrepancy between test and production. If the latter, I'm worried about the impact (performance and otherwise). That being said, I'm not as familiar with Doctrine caching as I'd like.

.travis.yml
```diff
@@ -34,7 +34,9 @@ before_install:
 before_script:
   - phpenv config-add .travis.php.ini
   - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then printf "\n" | pecl uninstall apcu; fi
-  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then printf "\n" | pecl install apcu-4.0.10; fi
+  - if [[ ${TRAVIS_PHP_VERSION:0:3} == "5.6" ]]; then printf "\n" | pecl install apc; fi
+  - if [[ ${TRAVIS_PHP_VERSION:0:2} == "7." ]]; then printf "\n" | pecl install apcu-4.0.10 apc_bc; fi
```
arothuis: `apcu_bc`

thijskh:
Fixed in #427

PR #400, opened 2017-03-30
Log the proxied service providers as part of the authentication log

nicwortel:
Resolves https://www.pivotaltracker.com/story/show/118569209

### DIFF
--- a/tests/unit/OpenConext/Mockery/Matcher/ValueObjectListEqualsMatcher.php
+++ b/tests/unit/OpenConext/Mockery/Matcher/ValueObjectListEqualsMatcher.php
@@ -19,9 +19,7 @@ final class ValueObjectListEqualsMatcher extends MatcherAbstract
             __METHOD__
         );
 
-        $valueObjects = array_values($valueObjects);
-
-        $first = $valueObjects[0];
+        $first = array_values($valueObjects)[0];
 
         if (!is_object($first) || !method_exists($first, 'equals')) {
             throw new InvalidArgumentException($message);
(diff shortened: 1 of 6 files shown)


thijskh:
@pmeulen

tests/unit/OpenConext/Mockery/Matcher/ValueObjectListEqualsMatcher.php
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @return bool
+     */
+    public function match(&$actual)
+    {
+        if (!is_array($actual) || count($actual) !== count($this->_expected)) {
+            return false;
+        }
+
+        foreach (array_values($actual) as $index => $valueObject) {
```
DRvanR: since this is a list-comparator, the keys are actually relevant, I'd remove the array_values.

tests/unit/OpenConext/Mockery/Matcher/ValueObjectListEqualsMatcher.php
```diff
@@ -0,0 +1,70 @@
...
+
+        return true;
+    }
+
+    /**
+     * @return string
+     */
+    public function __toString()
+    {
+        return sprintf('<ValueObjectListEqualsMatcher(%s)>', $this->_expected);
```
DRvanR: Seems to me this needs to be updated?

tests/unit/OpenConext/Mockery/Matcher/ValueObjectListEqualsMatcher.php
nicwortel: Are they though? I would think that the order is important, but the actual keys are not (since it's a list, not a map)?

DRvanR: In this implementation the keys are actually the index of the list and therefor very much relevant; They're not under control of the consumer as would be with a map. If you would like to discuss this further I suggest we do that in person on monday as that is easier than in a PR :wink:

PR #403, opened 2017-04-04
Log a notice before throwing UnknownIssuer / UnknownPreselectedIdP exceptions

nicwortel:


library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -319,14 +319,17 @@ public function setRemoteIdpMd5($remoteIdPMd5)
...
-                "Unable to map remote IdpMD5 '$remoteIdPMd5' to a remote entity!",
-                $remoteIdPMd5
-            );
+
+        if (isset($this->_configs['Idp'])) {
+            return $this;
         }
 
-        return $this;
+        $this->getSystemLog()->notice(sprintf('Unable to map remote IdpMD5 "%s" to a remote entity.', $remoteIdPMd5));
```
nicwortel: I used the system log, but there is also a session log. I assumed this would be the correct logger to use.

nicwortel: Because the logger is not set in the constructor, I used the `getSystemLog()` method to make sure I get a logger and not `null`.

DRvanR: :+1: Please add a tech debt ticket to the backlog to remove this confusing setup - in the end all loggers end up as a call to `EngineBlock_ApplicationSingleton::getLog` which itself resolves to the injected monolog logger. No need to keep things complex when it can be simplified :smile:

nicwortel: Done, see https://www.pivotaltracker.com/story/show/143228773

PR #405, opened 2017-04-06
Use the browser preferred languages in locale selection

nicwortel:
https://www.pivotaltracker.com/story/show/142511655


tvdijen:
Hey Nic,

Could you elaborate on this a bit more?:
https://github.com/OpenConext/OpenConext-engineblock/pull/405/commits/ff2f0cd65f3ae138ba8005b46c568fa1958ed488#diff-83a7f2880ad7a7ee600777c27091769fR34

Shouldn't secure/httponly at least be configurable instead of being hardcoded to false?

nicwortel:
@tvdijen for now we are trying to mimic the cookies set by other applications in the platform. Those are set without the "httpOnly" and "secure" flags.

Since you are asking for it, however, I will add the possibility to override these defaults using parameters in `ini_parameters.yml`.

tvdijen:
The comment triggered me by suggesting incompatibility with the rest of the suite..

Thanks for the changes! I think we should have safe defaults whenever possible, even when we're talking locale-cookies..

DRvanR:
With regards to the cookies, we share your sentiment (secure by default, configurable, etc), but that is indeed out of scope for this issue. It will be done at some point in the future as part of the larger refactoring though.

DRvanR:
@nicwortel would like to see a feature added that tests the following scenario's:

- when I make the first visit and don't have a `lang` cookie, a cookie is set with the configured default locale
- when I make a consecutive visit, the expiration of the cookie is updated (not sure we can check this, but would be nice), but the locale not
- when I visit a page, the UI reflects the locale in the cookie
- when I change the language in the UI, the cookie is updated and the UI reflects the locale change

This could be tested against a simple error page, for instance (no need for a SAML auth): https://engine.vm.openconext.org/authentication/proxy/idps-metadata?sp-entity-id=urn:example.org generates a quick and simple error page, which can be translated through UI interaction.

src/OpenConext/EngineBlockBridge/EventListener/CortoTranslationsLocaleListener.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace OpenConext\EngineBlockBridge\EventListener;
+
+use EngineBlock_ApplicationSingleton;
+use OpenConext\EngineBlockBundle\Localization\LocaleProvider;
+use Zend_Translate_Adapter;
+
+final class CortoTranslationsLocaleListener
```
DRvanR: Perhaps make a bit more explicit that this sets the locale by prepending `Set`?

src/OpenConext/EngineBlockBundle/DependencyInjection/OpenConextEngineBlockExtension.php
```diff
@@ -25,6 +25,7 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('logging.yml');
 
         $loader->load('bridge.yml');
+        $loader->load('bridge_event_listeners.yml');
```
DRvanR: 👍

src/OpenConext/EngineBlockBridge/EventListener/CortoTranslationsLocaleListener.php
DRvanR: Might also want to add a class level docblock that this must run after the LocaleListener

src/OpenConext/EngineBlockBundle/EventListener/LocaleListener.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->localeProvider = $localeProvider;
+        $this->cookieFactory = $cookieFactory;
+    }
+
+    public function onKernelRequest(GetResponseEvent $event)
+    {
+        $request = $event->getRequest();
+
+        $this->localeProvider->scopeWithRequest($request);
```
DRvanR: Might want to consider `scopeWith(Request)`, reads a bit better. Just for consideration, not a must or anything by miles 😉

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace OpenConext\EngineBlockBundle\Http\Cookies;
+
+use Assert\Assertion;
```
DRvanR: Please use [the own assertion][assert] object.

[assert]: https://github.com/OpenConext/OpenConext-engineblock/blob/master/src/OpenConext/EngineBlock/Assert/Assertion.php

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
```diff
@@ -0,0 +1,79 @@
...
+     * @param string $value
+     *
+     * @return Cookie
+     */
+    public function createCookie($value)
+    {
+        $expire = 0;
+
+        if ($this->expiry) {
+            $expire = (new DateTime())->add(DateInterval::createFromDateString($this->expiry . ' seconds'));
```
DRvanR: FYI, alternatively: `(new DateTime())->add(new DateInterval(sprintf('PT%dS', $this->expiry)));`

tests/integration/OpenConext/EngineBlockBundle/Localization/LocaleProviderTest.php
```diff
@@ -0,0 +1,142 @@
+<?php
+
+namespace OpenConext\EngineBlockBundle\Localization;
+
+use PHPUnit_Framework_TestCase;
+use Symfony\Component\HttpFoundation\Request;
+
+class LocaleProviderTest extends PHPUnit_Framework_TestCase
```
DRvanR: Would llike to see some tests that cover the priority rules: GET > POST > COOKIE > browser. They're all tested in isolation, but not together.

thijskh:
Re the cookie parameters discussion: we want the settings to be the same as they are now, because otherwise all current cookies will not be recognised. That means not secure and not httponly. This is not a security problem for the language cookie of course.

For new deployments I would indeed recommend to set the secure flag, just because it's good practice to always set it, and will reduce noise from security checkers that claim that this is a risk.

I'm not so sure about httponly, since I can imagine that at some point there would be a need to access language info from JavaScript, even though that's not done today.

All of couse only applies to the language cookie; our other cookies are already secure+httponly.

src/OpenConext/EngineBlockBundle/EventListener/LocaleListener.php
nicwortel: I had considered it, but I personally prefer this way :wink:

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
nicwortel: I have considered this method as well, but I figured that "x seconds" is a better way to communicate the intention.

tests/integration/OpenConext/EngineBlockBundle/Localization/LocaleProviderTest.php
nicwortel: Yes, I have added extra test cases.

arothuis:
> I'm not so sure about httponly, since I can imagine that at some point there would be a need to access language info from JavaScript, even though that's not done today.

For future reference, the language (potentially retrieved from a cookie) can be injected in a JavaScript configuration when rendering the page, see [discover.phtml](https://github.com/OpenConext/OpenConext-engineblock/blob/eb60495cd3bae41c145bda1a3f23fc60d897cb16/theme/material/templates/modules/Authentication/View/Proxy/discover.phtml#L140-L156). We are doing this to render `requestAccessUrl`s with the correct language.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
```diff
@@ -435,4 +443,50 @@ public function theAuthnRequestToSubmitShouldMatchXpath($xpath)
...
+     */
+    public function myBrowserIsConfiguredToPrefer($locale)
+    {
+        $this->getMainContext()->getMinkContext()->getSession()->setRequestHeader('Accept-Language', $locale);
+    }
+
+    /**
+     * @Given I don't have a cookie with a locale preference set
+     */
+    public function iDonTHaveACookieWithALocalePreferenceSet()
```
DRvanR: if it doesn't lead to an actual action, why is this a stepdefinition?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
```diff
@@ -435,4 +443,50 @@ public function theAuthnRequestToSubmitShouldMatchXpath($xpath)
...
+    /**
+     * @Then /^a cookie should be set with locale "([^"]*)"$/
+     */
+    public function aCookieShouldBeSetWithLocale($locale)
+    {
+        $cookie = $this->getMainContext()->getMinkContext()->getSession()->getCookie('lang');
+
+        if ($cookie !== $locale) {
+            throw new ExpectationException(
+                sprintf('The "lang" cookie should contain "%s", but contains "%s"', $locale, $cookie),
```
DRvanR: Might be worthwhile to make a separate, more informing test & message if there is no cookie

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
```diff
@@ -435,4 +443,50 @@ public function theAuthnRequestToSubmitShouldMatchXpath($xpath)
...
+     * @Given I don't have a cookie with a locale preference set
+     */
+    public function iDonTHaveACookieWithALocalePreferenceSet()
+    {
+    }
+
+    /**
+     * @Then /^a cookie should be set with locale "([^"]*)"$/
+     */
+    public function aCookieShouldBeSetWithLocale($locale)
```
DRvanR: This is not `a` cookie, but a specific `lang` cookie :wink: (as above, either generalize or make more specific)

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/LocaleSelection.feature
```diff
@@ -0,0 +1,27 @@
+Feature:
+  As a user,
+  I want EB to display the interface in the language that my browser sends,
+  so I can understand the interface better.
+  Where desired, I want to override the language with the EN/NL buttons in the interface.
```
DRvanR: Food for thoguht: "Also, I want to be able to set my preferred language from the available choices" seems more fitting to me, removes the NL/EN implementation details and "where desired" - it's platform wide, not just EngineBlock only.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/LocaleSelection.feature
```diff
@@ -0,0 +1,27 @@
...
+    Given I have a locale cookie containing "nl"
+    When I go to Engineblock URL "/authentication/sp/debug"
+    Then a cookie should be set with locale "nl"
+    And I should see "Gebruiksvoorwaarden"
+
+  Scenario: a user changes their locale
+    Given I have a locale cookie containing "nl"
+    When I go to Engineblock URL "/authentication/sp/debug?lang=en"
+    Then a cookie should be set with locale "en"
+    And I should see "Terms of Service"
```
DRvanR: Might want to add a scenario that tests a browser sending an unsupported locale makes EB use the default configured language.

DRvanR: Would it also be possible to send a POST somewhere? Would allow for testing that functionality as well...

Maybe post a bogus SAML Reponse, which will lead you to be redirected to an error page, which can be verified of being of the correct locale?

tests/unit/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactoryTest.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public function sets_the_expiry_time()
+    {
+        $expiry = 3600;
+
+        $cookieFactory = new CookieFactory('name', '.example.com', $expiry);
+
+        $cookie = $cookieFactory->createCookie('value');
+
+        // Allow the result to be off by one second, to compensate for time issues
```
DRvanR: this is what I mean with the DateTime  thing above.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
arothuis: Because it makes the actual functionality more explicit, I reckon. However, I think this step definition can go, as one can assume the absence of a cookie if it has not been explicitly set in the feature context. 

Otherwise, this step definition would ensure the absence of that specific lang cookie.

DRvanR: Was a question for @nicwortel hehe, I assume the same. However, unless it actually does something that needs to be done explicitly (albeit perhaps unnecessary, something like removing the lang cookie), it is not something that should be a step because no actual step is taken - it could be part of the scenario description for instance.

nicwortel: @DRvanR I don't think there is time in the scope of this ticket, but I've created a new issue: https://www.pivotaltracker.com/story/show/143502037

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/LocaleSelection.feature
nicwortel: In this case, the feature description comes right from the Pivotal feature (https://www.pivotaltracker.com/story/show/142511655).

I didn't think it is up to me to change the description into my own interpretation.

src/OpenConext/EngineBlockBridge/EventListener/SetCortoTranslationsLocaleListener.php
```diff
@@ -0,0 +1,39 @@
...
+
+use EngineBlock_ApplicationSingleton;
+use OpenConext\EngineBlockBundle\Localization\LocaleProvider;
+use Zend_Translate_Adapter;
+
+/**
+ * This listener depends on the LocaleListener in the EngineBlockBundle (which has to scope the LocaleProvider with the
+ * current request), so the priority should be set so that it is called after the LocaleListener. Otherwise, if the
+ * LocaleProvider hasn't been scoped with the request yet, it will fall back to the default locale and all user
+ * preferences are ignored.
```
DRvanR: nice!

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/LocaleSelection.feature
DRvanR: check

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
```diff
@@ -0,0 +1,79 @@
...
+     * @param string $value
+     *
+     * @return Cookie
+     */
+    public function createCookie($value)
+    {
+        $expiresAt = 0;
+
+        if ($this->expiryInSeconds) {
+            $expiresAt = (new DateTime())->add(DateInterval::createFromDateString($this->expiryInSeconds . ' seconds'));
```
DRvanR: if we cannot create a DateTime object that allows us to control time for the sake of tests within the scope of this issue, please use the [TimeProvider][1] instead (use the interface for typehints) so that we can have strict tests. This requires setting up a bridge class that makes a call to the DiContainer to retrieve the TimeProvider. 

[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/master/library/EngineBlock/TimeProvider/Default.php

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
nicwortel: @DRvanR I don't think there is time in the scope of this ticket, but I've created a new issue: https://www.pivotaltracker.com/story/show/143502037

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
nicwortel: I do not fully understand how you want to implement this, perhaps we can discuss this IRL tomorrow.

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
DRvanR: Please expand the issue a bit so that it is clear from the issue itself what should be done.

src/OpenConext/EngineBlockBundle/Http/Cookies/CookieFactory.php
DRvanR: will do

PR #406, opened 2017-04-07
Added checks for the IssueInstant of SAML messages

baszoetekouw:
Check if IssueIntant in received SAML messages is in a reasonable interval.
This should give advance warnings of NTP-issues at IdPs and SPs.

Note: this is untested as of yet, please do not merge right away.

### DIFF
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -107,13 +107,7 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
         }
 
         // check the IssueInstant against our own time to see if the SP's clock is getting out of sync
-        if ($sspRequest->getIssueInstant() <= time() - 10 || $sspRequest->getIssueInstant() >= time() + 10) {
-            $this->_logger->notice(sprintf(
-                'IssueInstant of received request is out of sync (%lu), might indicate NTP-issues at the SP',
-                $sspRequest->getIssueInstant()
-            ));
-        }
-
+        $this->_checkIssueInstant( $sspRequest->getIssueInstant() );
 
         $ebRequest = new EngineBlock_Saml2_AuthnRequestAnnotationDecorator($sspRequest);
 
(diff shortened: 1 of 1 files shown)


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -106,6 +106,15 @@ public function receiveRequest()
             ));
         }
 
+        // check the IssueInstant against our own time to see if the SP's clock is getting out of sync
+        if ($sspRequest->getIssueInstant() <= time() - 10 || $sspRequest->getIssueInstant() >= time() + 10) {
+            $this->_logger->notice(sprintf(
+                'IssueInstant of received request is out of sync (%lu), might indicate NTP-issues at the SP',
```
thijskh: Would add "at the SP (or this machine)" and replace NTP with "clock synchronization".

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -251,6 +260,15 @@ public function receiveResponse()
             );
         }
 
+        // check the IssueInstant against our own time to see if the IdP's clock is getting out of sync
+        // a more strict check is implemented elsewehere; this simply serves as an advance warning
+        if ($sspResponse->getIssueInstant() <= time() - 10 || $sspResponse->getIssueInstant() >= time() + 10) {
+            $this->_logger->notice(sprintf(
+                'IssueInstant of received response is out of sync (%lu), might indicate NTP-issues at the IdP',
```
thijskh: as above

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -106,6 +106,15 @@ public function receiveRequest()
             ));
         }
 
+        // check the IssueInstant against our own time to see if the SP's clock is getting out of sync
+        if ($sspRequest->getIssueInstant() <= time() - 10 || $sspRequest->getIssueInstant() >= time() + 10) {
```
thijskh: Open question is what a reasonable time is. In order to make this useful, it must not be too low to prevent too many false positives with clocks that are temporarily a little out of sync.

It's arbitrary, but we keep 60 seconds as the margin for validating any incoming time, so halving it to 30 seconds tolerance for the warning would give us enough advance warning and not too many FPs?

thijskh: Calculate diff once, then compare it once (with abs)? So we can use the calculated diff in the log message ("time is off by `x` sec").

DRvanR: Small but important request: please use the [EngineBlock_TimeProvider_Default](https://github.com/OpenConext/OpenConext-engineblock/blob/master/library/EngineBlock/TimeProvider/Default.php) object rather than the `time()` function - the former allows us to control time and thus test behaviour, the latter not.

DRvanR:
Now just a matter of adding tests and it's a 👍 from me 😉

baszoetekouw:
@DRvanR Could you point me to where the tests should be added?  I looked through the tests/ dir, but it wasn't entirely obvious (to me at least) where to add a functional test like this.

DRvanR:
In all honesty I doubt this is easily tested regardless...

Creating a functional test (using Behat) for this will be very complex as the wanted behavior is one where the logging is verified.

Would recommend adding a unittest here: https://github.com/OpenConext/OpenConext-engineblock/blob/master/tests/library/EngineBlock/Test/Corto/Module/BindingsTest.php

Creating a unittest where you can use a different logger is a lot easier than doing this with a functional test. It does require injecting the logger into the binding class rather than retrieving it from the DI in the constructor though, which would require even more changes...

In all honesty (but @thijskh would have to sign off on that) seeing the limited scope and impact of the changes and the impact required to make this testable I would say that adding this without a test might be the best option for now...

thijskh:
Thanks! I've merged the changes after combining the two commits into one.

PR #408, opened 2017-04-11
Remove welcome mail code

nicwortel:
https://www.pivotaltracker.com/story/show/135943729


thijskh:
What about the table `emails` and `library/EngineBlock/Mail/Mailer.php`?

DRvanR:
@thijskh There is still [debug mail][1] functionality and [request access][2] mail functionality. If those can go as well, a lot more could be cleaned up. The latter is disabled at the moment if I recall correctly (due to previous XSS attack vector), the former I'm not sure when it's used.

As an aside, searching for [mail][3] helps with identification of where it's used :wink: 

[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/1e4fa842d7d868d124514b95977a7d58cbb38782/library/EngineBlock/Corto/Module/Service/SingleSignOn.php#L423
[2]: https://github.com/OpenConext/OpenConext-engineblock/blob/366a74df01919727cffb7d7624fede2d0c07e489/src/OpenConext/EngineBlockBundle/Controller/IdentityProviderController.php#L146
[3]: https://github.com/OpenConext/OpenConext-engineblock/search?utf8=%E2%9C%93&q=mail&type=

thijskh:
> @thijskh There is still debug mail functionality and request access mail functionality. If those can go as 
> well, a lot more could be cleaned up. The latter is disabled at the moment if I recall correctly (due to 
> previous XSS attack vector), the former I'm not sure when it's used.

Those (and only those) need to stay. Fixing the latter one is on the backlog.

However, I think both of those do not use the code I mentioned in my previous comment.

DRvanR:
Something for @nicwortel to check then 😄

nicwortel:
@thijskh you are right, the `EngineBlock_Mail_Mailer` is indeed no longer being used, so I have removed it.

It looks like the `emails` table is already removed in [`patch-0020.sql`](https://github.com/OpenConext/OpenConext-engineblock/blob/master/database/patch/patch-0020.sql).

PR #409, opened 2017-04-11
Broad API functional tests

arothuis:
[112497975](https://www.pivotaltracker.com/story/show/112497975)

This PR adds:
* Tests for basic API interaction
* Small corrections in the API: throw correct Exceptions, depend on service interfaces not implementations

More extensive functional tests will be added in a separate PR (#413).


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/ApiContext.php
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * @When /^I check EngineBlock's heartbeat$/
+     */
+    public function iCheckEngineBlockSHeartbeat()
+    {
+        $this->visit('/');
+    }
+
+    private function visit($path)
```
arothuis: This should be modified to make use of explicit request methods and payloads (e.g. GET, POST).

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Api/Heartbeat.feature
```diff
@@ -0,0 +1,9 @@
+Feature:
+  In order to determine whether EngineBlock is available
+  As a client service or availability monitor
+  I want to be able to check EngineBlock's heartbeat
+
+  @WIP
+  Scenario: EngineBlock returns a heartbeat
+     When I check EngineBlock's heartbeat
```
arothuis: What level of detail shall we use when describing API functionality?

I'd say we have two options here:
1. Generalize to match HTTP/REST implementations (technical)?
1. Explicitize the intent by describing behaviors (behavioral)?

I would prefer the technical option as it regards an API, which is technical in nature.

arothuis:
Before I go in and implement all the API endpoints, I'd like to put some thought into designing this properly. Any thoughts on the comments above, @DRvanR @nicwortel

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Api/Heartbeat.feature
DRvanR: Always technical. API's are not behavioral, nor do we have any behavioral tests. With the chance of sounding like a scratched record: we leverage behat to do functional testing, *not* BDD in any way shape or form.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/ApiContext.php
DRvanR: Personally think that this could better be generalized to something like "visit URL" rather than specific naming and hiding what actually matters (the route) in an implementation detail.

DRvanR:
Even though the issue itself states Behat, has there been given thought to using for instance LiipFunctionaltTestingBundle or even plain webtestcases for API tests? Behat does not seem the be the best tool for testing API's as we have experienced now in various projects.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Api/Heartbeat.feature
nicwortel: As discussed IRL, it is probably best to go for the technical approach, using steps like
> When I send a "GET" request to "/some/path"
> Then the response status code should be 200

arothuis:
Thanks for the replies. The reason I asked is because one could describe the API (which could be functional and behavioral) or its implementation (which is technical). 

I was indeed wondering whether an existing library, bundle or context would be more suited for the job. LiipFunctionalTestBundle looks like a good candidate.

arothuis:
These tests cover the error cases of the API.

In a separate PR, the interactions with database, metadata, serviceregistry will be brought under test as well.

build.xml
```diff
@@ -113,6 +113,13 @@
         </exec>
     </target>
 
+    <target name="phpunit-functional-tests" description="Run functional tests with PHPUnit">
```
arothuis: We already had `functional-tests` as a top-level target, running `behat-travis` so I've named these differently. I've added the PHPUnit-based functional tests to that target and to the `pre-commit` as well. If the runtime would increase too much, these tests could be moved to the `pre-push`.

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyControllerTest.php
```diff
@@ -0,0 +1,156 @@
+<?php
+
+namespace OpenConext\EngineBlockBundle\Tests;
+
+use Liip\FunctionalTestBundle\Test\WebTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+class AttributeReleasePolicyControllerTest extends WebTestCase
```
arothuis: This test also exists as [a unit test](https://github.com/OpenConext/OpenConext-engineblock/blob/6874cc1700662b34659e50ffcb75bd700eea3aeb/tests/unit/OpenConext/EngineBlockBundle/Controller/AttributeReleasePolicyControllerTest.php). I'm not sure whether that test is still useful if a fully covering functional test is in place.

build.xml
arothuis: Also, this does not work as tests integrating with the database will not work if they are not run in the vm.

src/OpenConext/EngineBlockBundle/Controller/Api/ConnectionsController.php
```diff
@@ -7,12 +7,17 @@
...
+use OpenConext\EngineBlockBundle\Http\Exception\ApiMethodNotAllowedHttpException;
+use OpenConext\EngineBlockBundle\Http\Exception\ApiNotFoundHttpException;
 use OpenConext\EngineBlockBundle\Http\Exception\BadApiRequestHttpException;
 use OpenConext\EngineBlockBundle\Http\Request\JsonRequestHelper;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 
+/**
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects) Static calls, factories
```
nicwortel: Perhaps you can also mention here the fact that some checks have to be made in the controller instead of by Symfony itself

src/OpenConext/EngineBlockBundle/Controller/Api/MetadataController.php
```diff
@@ -4,16 +4,20 @@
...
 use OpenConext\EngineBlock\Service\MetadataService;
 use OpenConext\EngineBlockBundle\Http\Response\JsonHelper;
 use OpenConext\Value\Exception\InvalidArgumentException;
 use OpenConext\Value\Saml\EntityId;
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 
+/**
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects) Static calls, factories
```
nicwortel: Perhaps you can also mention here the fact that some checks have to be made in the controller instead of by Symfony itself

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyControllerTest.php
```diff
@@ -0,0 +1,156 @@
...
+     * @test
+     * @group Api
+     * @group Arp
+     * @group Profile
+     */
+    public function authentication_is_required_for_applying_arps()
+    {
+        $unauthenticatedClient = $this->makeClient();
+        $unauthenticatedClient->request('POST', 'https://engine-api.vm.openconext.org/arp');
+        $this->assertStatusCode(Response::HTTP_UNAUTHORIZED,  $unauthenticatedClient);
```
nicwortel: Is it enough to test the response status code, or should we also test that the response body does not contain data and/or that no side effects happen?

tests/unit/OpenConext/EngineBlockBundle/Controller/AttributeReleasePolicyControllerTest.php
```diff
@@ -32,7 +32,11 @@ public function access_is_denied_for_unauthorized_requests()
         $authorizationChecker = $this->mockAuthorizationCheckerDenyingAccessToProfile();
 
         $arpController = new AttributeReleasePolicyController($authorizationChecker, $metadataService, $arpEnforcer);
-        $arpController->applyArpAction(new Request);
+
+        $request = new Request;
```
nicwortel: This is a very minor, unimportant nitpick, but personally I would try to be consistent in always using parentheses in object creation, even if there are no constructor arguments.

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/MetadataControllerTest.php
```diff
@@ -0,0 +1,180 @@
+<?php
+
+namespace OpenConext\EngineBlockBundle\Tests;
+
+use Liip\FunctionalTestBundle\Test\WebTestCase;
+use OpenConext\EngineBlockBundle\Configuration\Feature;
+use OpenConext\EngineBlockBundle\Configuration\FeatureConfiguration;
+use stdClass;
```
nicwortel: This class does not seem to be used here, please remove it.

PR #410, opened 2017-04-12
New Metadata API & Metadata Storage.

DRvanR:
Story: https://www.pivotaltracker.com/story/show/111132276

Todo:
- [ ] Update metadata documentation links in `README` and `docs/Metadata/index.md` to link to master branch rather than feature/metadata-api branch.

Development remarks:
A patch file is included in the PR, this is purely intended to be used for debugging during development by using a file rather than having to push json constantly. This should be removed from the PR when no longer of use. Furthermore it is recommended to request a dump from production Janus/SR metadata to ensure full compatibility with the current metadata.

Todo for the required feature:
- [ ] Resolve all open questions and remarks (see [questions and findings][qaf]). Take care that any changes to logic/default values regarding SAMLValueObjects must be made local in EngineBlock as the value objects are conforming to the SAML specification. Changes to logic/default values for all other value objects are local to EngineBlock by nature and can thus be changed.
- [ ] Ensure that an additional value object is added to the IdentityProvider and ServiceProvider objects that allow storing of arbitrary `key => value(s)` structures to allow flexible expansion of the configuration. Should the expansion become a part of mainstream application logic, a custom value object should be built for that.
- [ ] Create an entity and repositories to persist and hydrate the models. The value objects should be stored as json blobs. each entity should also store its UUID, EntityID and EntityType.
- [ ] Create a model and accompanying repository that allows the modelling and querying of Connections. A connection is an explicit allowed connection between a ServiceProvider and an IdentityProvider. If there is no limit on which IdPs can be used for an SP, the idea is not to store any connections to reduce load (meaning introduction of the rule: no connection means all connections are allowed)
- [ ] Parse the connections from the JSON and create connections between the correct metadata models.
- [ ] Create the entity and repository to persist and load connections.
- [ ] Create a service that allows the retrieval of metadata and connections (read only - reading should be used everywhere but in the the ConnectionsController, writing should only be used in the ConnectionsController
- [ ] Create a service that allows the updating of the metadata and connections (write only). For connections cheapest would be wipe and replace, for the metadata it should be checked, updates might be better, but are expensive in a cluster, therefor wipe/replace all might be best.
- [ ] Use the new services and storage to persist the metadata and connections in the new structure. This could be done in an intermediate release with a feature toggle so that temporarily both exists and the new structure is only used for writes, the old for both writes and reads. This allows for verification of compatibility with production data.
- [ ] Adapt all usages of metadata to the new metadata structures, effectively implementing the usage of the new metadata structures on the read side. If necessary/effective, bridge functionality can be created that allows transformation of the new structure to the old structure to minimize the required changes in the `library` (EB4) code. All new code must use the new structures.
- [ ] Remove any and all code that is no longer needed.

[qaf]: https://github.com/OpenConext/OpenConext-engineblock/blob/24f2f9b21edac9b12f54fff89e0ede195d95701d/docs/Metadata/questions-and-findings.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ Also, the following documentation can be found in the [docs][docs] directory:
 1. [Release Procedure][docs-release]
 1. [EngineBlock Input and Output Command Chains][docs-filter]
 1. [Release notes for releases < 5.0.0][docs-release-notes]
+1. [Metadata Documentation][docs-metadata]
 
 [travis-build]: https://travis-ci.org/OpenConext/OpenConext-engineblock
 [license]: LICENSE-2.0.txt
@@ -216,3 +217,4 @@ Also, the following documentation can be found in the [docs][docs] directory:
 [docs-release]: https://github.com/OpenConext/OpenConext-engineblock/tree/master/docs/release_procedure.md
 [docs-filter]: https://github.com/OpenConext/OpenConext-engineblock/tree/master/docs/filter_commands.md
 [docs-release-notes]: https://github.com/OpenConext/OpenConext-engineblock/tree/master/docs/release_notes
+[docs-metadata]: https://github.com/OpenConext/OpenConext-engineblock/blob/feature/metadata-api/docs/Metadata/index.md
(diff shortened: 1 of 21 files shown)


MKodde:
Closing this PR as these changes have been addressed in the upcoming 5.7 release.

PR #411, opened 2017-04-13
Remove LDAP backend

nicwortel:
https://www.pivotaltracker.com/story/show/142941799


thijskh:
README.md still references ldap als a requirement. Should probably also verify that EB indeed works without php-ldap module installed.

PR #412, opened 2017-04-14
Move ConsentRepository to EngineBlockBundle

nicwortel:
https://www.pivotaltracker.com/story/show/117679053

### DIFF
--- a/src/OpenConext/EngineBlock/Service/ConsentService.php
+++ b/src/OpenConext/EngineBlock/Service/ConsentService.php
@@ -2,7 +2,7 @@
 
 namespace OpenConext\EngineBlock\Service;
 
-use Doctrine\DBAL\DBALException;
+use Exception;
 use OpenConext\EngineBlock\Authentication\Dto\Consent;
 use OpenConext\EngineBlock\Authentication\Dto\ConsentList;
 use OpenConext\EngineBlock\Authentication\Model\Consent as ConsentEntity;
@@ -46,7 +46,7 @@ final class ConsentService
     {
         try {
             $consents = $this->consentRepository->findAllFor($userId);
-        } catch (DBALException $e) {
+        } catch (Exception $e) {
             throw new RuntimeException(
                 sprintf('An exception occurred while fetching consents the user has given ("%s")', $e->getMessage()),
                 0,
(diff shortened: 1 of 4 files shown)


nicwortel:
Please note that the `ConsentService` (https://github.com/OpenConext/OpenConext-engineblock/blob/c96243ad945a5900cc84656cef2d8db6ba3d5a2d/src/OpenConext/EngineBlock/Service/ConsentService.php) still has a `catch` for `DBALException`. We should decouple the `ConsentService` from DBAL (which is an implementation detail) but then we'd have to create a generic `catch` for the base `Exception`.

We could also introduce our own exception and have the `DbalConsentRepository` throw that instead of a `DBALException`.

What do you think?

DRvanR:
> We could also introduce our own exception and have the DbalConsentRepository throw that instead of a DBALException.

This :)

Nice refactoring!

arothuis:
The throwing/catching of the correct exception should be fixed in this PR as well.

Normally, I prefer throwing and catching an aptly named custom Exception. However, if you were to solely catch those Exceptions in the Service, you would have to force all implementations of the repository to use those Exceptions  (which cannot be done technically). You could throw a custom (runtime) Exception in the repository and catch a generic Exception in the Service, so proper context can be given to what has gone wrong.

src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
```diff
@@ -0,0 +1,68 @@
...
+                consent
+            WHERE
+                hashed_user_id=:hashed_user_id
+        ';
+
+        try {
+            $statement = $this->connection->executeQuery($sql, ['hashed_user_id' => sha1($userId)]);
+            $rows = $statement->fetchAll(PDO::FETCH_ASSOC);
+        } catch (DBALException $exception) {
+            throw new RuntimeException('Could not fetch user consents from the database', 0, $exception);
```
arothuis: This could also be a child of EngineBlock RuntimeException to have an even more specific context, but I'm not sure that offers any extensive benefits.

src/OpenConext/EngineBlock/Service/ConsentService.php
```diff
@@ -46,7 +46,7 @@ public function findAllFor($userId)
     {
         try {
             $consents = $this->consentRepository->findAllFor($userId);
-        } catch (DBALException $e) {
+        } catch (Exception $e) {
```
arothuis: As you cannot technically enforce implementations of the ConsentRepository to throw a certain exception and there are multiple things that could go wrong when fetching consents, catching all possible exceptions seems right to me.

PR #419, opened 2017-05-16
Remove unused vendor library

DRvanR:
It seems that the openid  library is not in use in the project. It does require a php extension that is not installed on the dev env (php-gmp). By removing the library the extension is no longer required. This solves the first half of https://github.com/OpenConext/OpenConext-deploy/issues/133


thijskh:
This was a dependency of simpleSAMLphp which has been dropped at some point.

PR #423, opened 2017-06-15
Make DROP TABLE migrations more robust to when a table is already gone.

thijskh:
It's simple: if it's for some reason already away, the goal has been accomplished, no need to fail.


thijskh:
@MKodde @jorissteyn

PR #424, opened 2017-06-15
Support the new attribute aggregator

jorissteyn:
This change allows the serviceregistry to specify sources for attributes in the ARP. The sources are stored in engineblocks metadata and passed to the attribute aggregator. The API of the aggregator has been changed to support this feature and this PR updates the client code to the new aggregator API format.

Related PRs:

 - [PR for engineblock-metadata](https://github.com/OpenConext/OpenConext-engineblock-metadata/pull/13)
 - PR for service registry (work in progress)

Another PR will deal with showing attribute sources on the consent screen. Until then, the aggregated attributes will be shown indistinguishable from regular attributes.

Tasks done:

- [X] Refactor AA client to use new fw components and support the new API
- [X] Create behat scenario's using a client stub
- [x] ARP configuration with a source should be sent to the AA endpoint
- [x] Upgrade to version of engineblock-metadata supporting attribute sources
- [X] Figure out why the aggregator client doesn't work on PHP7 ( -> not related to this PR)
- [x] Handle server errors in the aggregator endpoint
- [x] Replace attributes configured for aggregation (never allow from IdP if ARP specifies a source)


library/EngineBlock/Corto/Filter/Command/AttributeAggregator.php
```diff
@@ -27,30 +63,82 @@ public function execute()
...
+    private function clearAttributesForAggregation(array $rules) {
+        foreach ($rules as $rule) {
+            unset($this->_responseAttributes[$rule->name]);
+        }
+    }
+
+    /**
+     * Replace the attributes sent by the IdP with aggregated attributes.
+     *
+     * TODO: keep track of attribute sources.
```
jorissteyn: Has been removed/implemented in #426

composer.json
```diff
@@ -17,7 +17,7 @@
         "kriswallsmith/assetic": "~1.2.0",
         "monolog/monolog": "~1.13",
         "mrclay/minify": "~2.2",
-        "openconext/engineblock-metadata": "^2.0",
+        "openconext/engineblock-metadata": "dev-feature/attribute-aggregation",
```
jorissteyn: Let's update that when we merge all related PRs.

PR #428, opened 2017-06-26
Merged ProxyServer loggers

MKodde:
The two loggers (session and system) have been merged to a single
logger. The three fields have been merged into a new `_logger` field.
All usages in the EngineBlock module have been updated to use the new
logger.

Pivotal tracker: #143228773

### DIFF
--- a/library/EngineBlock/Corto/Adapter.php
+++ b/library/EngineBlock/Corto/Adapter.php
@@ -207,7 +207,7 @@ class EngineBlock_Corto_Adapter
                 );
             }
             else {
-                $this->_getSessionLog()->warning(
+                $this->_getLogger()->warning(
                     "Unable to apply RequesterID '$requesterId' to sub-scope the available IdPs as we don't know this SP!"
                 );
             }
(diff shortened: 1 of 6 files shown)


library/EngineBlock/Corto/Adapter.php
```diff
@@ -351,8 +351,8 @@ protected function _initProxy()
 
     protected function _configureProxyServer(EngineBlock_Corto_ProxyServer $proxyServer)
     {
-        $proxyServer->setSystemLog($this->_getSystemLog());
-        $proxyServer->setSessionLogDefault($this->_getSessionLog());
+
+        $proxyServer->setLogger($this->_getSystemLog());
```
jorissteyn: Why not remove all references to "session" and "system" logs, and just call it logger?

MKodde: At first I deliberately left this out of scope as the story asks for the merge of the ProxyService loggers. This change was the only file remaining in the project with a session/server log combination. So changing the Adapter was the proper thing to do. Thanks for bringing this up.

jorissteyn:
I found one remaining reference to the session logger in [EngineBlock_Test_Corto_Module_BindingsTest](https://github.com/OpenConext/OpenConext-engineblock/blob/feature/merge-proxy-server-loggers/tests/library/EngineBlock/Test/Corto/Module/BindingsTest.php#L18). Makes me wonder why the build doesn't break...

PR #429, opened 2017-06-27
Display unconnected IdPs in the WAYF

jorissteyn:
When the feature "display_unconnected_idps" is enabled for a service provider:

 - all unconnected IdPs are shown in a separate block on the WAYF
 - the block is searchable, just like connected IdPs
 - next to each unconnected IdP is a button "request"access" -> opens modal dialog
 - if no unconnected IdP matches the search query, a generic "request access" link is shown

Todo:

 - [x] prefill Institution field in the dialog with the selected IdP


src/OpenConext/EngineBlockFunctionalTestingBundle/Fixtures/ServiceRegistryFixture.php
```diff
@@ -66,6 +66,14 @@ public function spRequiresPolicyEnforcementDecision($entityId)
     public function requireAttributeAggregation($entityId)
     {
         $this->data[$entityId]['coin:attribute_aggregation_required'] = true;
+
+        return $this;
+    }
+
+    public function displayUnconnectedIdpsForSp($entityId, $displayUnconnected = true)
+    {
+        $this->data[$entityId]['coin:display_unconnected_idps_wayf'] = $displayUnconnected;
```
jorissteyn: Right, we're using the verb "display" everywhere.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/DisplayUnconnectedIdpsWayf.feature
```diff
@@ -0,0 +1,36 @@
...
+    And an Identity Provider named "Connected IdP1"
+    And an Identity Provider named "Connected IdP2"
+    And an Identity Provider named "Unconnected IdP1"
+    And an Identity Provider named "Unconnected IdP2"
+    And IdP "Connected IdP1" uses a blacklist for access control
+    And IdP "Connected IdP2" uses a blacklist for access control
+    And IdP "Unconnected IdP1" uses a whitelist for access control
+    And IdP "Unconnected IdP2" uses a whitelist for access control
+
+  Scenario: As a user for an SP I sees only connected IdPs without request form
```
jorissteyn: Fixed.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/DisplayUnconnectedIdpsWayf.feature
```diff
@@ -0,0 +1,37 @@
+@WIP
```
jorissteyn: @WIP prevents the scenario from running in CI - I've removed the tag.

jorissteyn:
> When all IdP's with access are in the 'Previously chosen:' section. The 'Identity Providers with access' section shows the text: 'Your search did not return any results.' even when I did not perform a search query. This might be unrelated to this story. It is not very user friendly in my opinion.

This is the current behaviour of the WAYF, let's discuss that with a PO.

> The 'Identity Providers without access' section do not always have the 'Request access' button visible. The story and the screenshot in the story says this should be the case. For now the button in only visible on hover. Maybe the requester of this feature can shed it's light on this. Personally I think your implementation is better from a UX standpoint.

I've interpreted the screenshot liberally. Buttons for IdPs with access are also only shown on hover, so it made sense to implement it the same way. Clicking on the row opens the dialog right away, you don't actually need the button.

thijskh:
I'm afraid it's gotten worse. Using this branch I now see no IdPs at all. The screenshot is from an SP with the display_unconnected_idps_wayf enabled. If it's disabled, the only difference is that the bottom box is omitted.

screenshot

MKodde:
While merging master to this FB, there was a merge conflict on a Javascript file. This introduced the problem in the screenshot above.

PR #431, opened 2017-07-05
Show idp logo on PEP error message

MKodde:
PEP errors are now enriched with the context of the IdP who raised the message. This is done by testing de PdP server response for the presence of the `IdPOnly` property in the `AttributeAssignment`. If it is present, the IdP logo is shown on the PEP error message. If not, or the logo is not set in Service Registry, the logo is omitted.

This is a joint venture with @oharsta. He provided the changes to the PDP server.


src/OpenConext/EngineBlockBundle/Pdp/PolicyDecision.php
```diff
@@ -78,6 +89,25 @@ public static function fromResponse(Response $response)
...
+     * Checks attributeAssignment for clues if this assignment is IdP specific. And sets the idpOnly field
+     * accordingly.
+     *
+     * @param AttributeAssignment $attributeAssignment
+     * @param PolicyDecision $policyDecision
+     */
+    private static function setIsIdpOnly(AttributeAssignment $attributeAssignment, PolicyDecision $policyDecision)
+    {
+        // When the Identifier is IdpOnly and its value is true, this is a message that's IdP specific.
+        $isIdpOnly = $attributeAssignment->attributeId === 'IdPOnly'
```
jorissteyn: See above, I believe this could be simpler if it was non-static. And, moving into the realm of personal taste, I would try to avoid one-liners and use a guard class instead for cases where the assignment value is not what we expect:

```php
private function assignmentIsIdpOnly(AttributeAssignment $assignment)
{
    if (empty($assignment->value)) {
        return false;
    }

    return $assignment->attributeId === 'IdpOnly';
}
```

src/OpenConext/EngineBlockFunctionalTestingBundle/Fixtures/FunctionalTestingPdpClient.php
```diff
@@ -140,4 +169,20 @@ public function clear()
...
     }
+
+    /**
+     * @param array $additionalData
+     * @return string
+     */
+    private function getIdpFromAdditionalData(array $additionalData)
+    {
+        $idp = '';
+        if (!empty($additionalData)) {
```
jorissteyn: Since we typehint $additionalData for being an array, this check is redundant.

app/Resources/views/modules/Authentication/View/Feedback/AuthorizationPolicyViolation.phtml
```diff
@@ -19,6 +21,13 @@ if (isset($_SESSION['feedbackInfo']['error_authorization_policy_decision'])) {
     } elseif ($policyDecision->hasStatusMessage()) {
         $policyDecisionMessage = $policyDecision->getStatusMessage();
     }
+
+    // Show the IdP logo for status messages
+    if ($policyDecision->isIdpSpecificMessage() && $policyDecision->hasIdpLogo()) {
+        /** @var Logo $logo */
+        $logo = $policyDecision->getIdpLogo();
```
MKodde: Agreed, results in a better separation of concerns.

app/Resources/views/modules/Authentication/View/Feedback/AuthorizationPolicyViolation.phtml
```diff
@@ -35,7 +44,15 @@ $layout->wide = true;
     <?php
         if (isset($policyDecisionMessage)) {
     ?>
+        <?php if ($logo !== false) : ?>
+            <img src="<?php echo $logo->getUrl(); ?>"
```
MKodde: Thanks for pointing this out. Fixed this and did some additional boy scouting.

src/OpenConext/EngineBlockBundle/Pdp/PolicyDecision.php
MKodde: I have improved readability of the method. This method is in the static context because it's used by the (static) `fromResponse` method.

app/Resources/views/modules/Authentication/View/Feedback/AuthorizationPolicyViolation.phtml
```diff
@@ -21,13 +19,11 @@ if (isset($_SESSION['feedbackInfo']['error_authorization_policy_decision'])) {
...
+}
 
-    // Show the IdP logo for status messages
-    if ($policyDecision->isIdpSpecificMessage() && $policyDecision->hasIdpLogo()) {
-        /** @var Logo $logo */
-        $logo = $policyDecision->getIdpLogo();
-    }
-
+$logo = null;
+if (isset($_SESSION['feedbackInfo']['idp_logo'])) {
```
jorissteyn: As discussed, is the logo properly cleared? Or do we show an irrelevant idp logo on subsequent error pages?

Plus: assign feedbackInfo to the view object instead of using the superglobal in the template.

MKodde: While implementing the resetting of the `idp_logo`, I decided to use another solution altogether. I actually went back to the previous solution. Which is reading the idpLogo directly from the `$policyDecision` variable in the template. The additional testing that was required previously is no longer required as this was moved to the `PolicyDecision` class in the previous code cleanup. 

Assigning the data directly to the view was not practical and would imply that this should be done for the other Zend templates.

PR #432, opened 2017-07-05
Uncomment mandatory settings

tvdijen:
Without them defined at all, you always immediately get the 'stuck-in-authentication-loop' error, without even being redirected to the IdP you picked.
Probably because 0 >= null evaluates to true @ https://github.com/OpenConext/OpenConext-engineblock/blob/master/src/OpenConext/EngineBlockBundle/Authentication/AuthenticationLoopGuard.php#L72

It probably needs a safe default in code as well...

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -49,8 +49,8 @@ engineblock.feature.encrypted_assertions_require_outer_signature = 1
 ; wayf.cutoffPointForShowingUnfilteredIdps = 50
 
 ;; Settings for detecting whether the user is stuck in a authentication loop within his session
-;engineblock.timeFrameForAuthenticationLoopInSeconds = 60
-;engineblock.maximumAuthenticationProceduresAllowed  = 5
+engineblock.timeFrameForAuthenticationLoopInSeconds = 60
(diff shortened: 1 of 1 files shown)


thijskh:
@MKodde Could you see if there's a way we can have these settings have a safe default in code (I'd say: if undefined, do no loop detection)?

MKodde:
@thijskh, yes this would be a possibility. The `AuthenticationLoopGuard` can be setup to disable itself when the default values are passed in its constructor. 

@tvdijen was absolutely right when assuming the null value is evaluated (casted in this case) to 0. This can be seen in [dump-required-ini-params.sh](https://github.com/OpenConext/OpenConext-engineblock/blob/master/bin/composer/dump-required-ini-params.sh#L68-L69).

Question is, what is the desired solution? I gravitate towards @thijskh solution as this provides us with a more testable solution.

tvdijen:
Fine with me, I just wanted to address this.. I ran into this when manually deploying

thijskh:
Yes let's go for the fix in the codebase.

thijskh:
Maybe good to merge this to fix the acute problem and create an issue to fix in codebase?

Issue #339, opened 2016-09-21
Pretty print compiled application.css instead of minimised version


mrvanes:
If a deployer of OpenConext wants to make fast and minimal changes to WAYF web/stylesheets/application.css, she is confronted with a minimised version of application.css. She would need to unminimise this file before making (small) changes like corporate identity colors when recompiling the metro theme would require installation of all sorts of supporting engines and compilers.

baszoetekouw:
IMO, there is no need for minimizing in the first place, as gzip compression is much better than whatever minimizing scheme is used.  (This would require enabling transparent gzip compression in the http-server, obviously.)

relaxnow:
Does appear to add little: https://css-tricks.com/the-difference-between-minification-and-gzipping/
Especially as SASS already concatenates everything.

thijskh:
I think the modern theming solves this in a better way than hacking things into the application.css directly.

PR #435, opened 2017-07-23
Unminify CSS and JavaScript

tvdijen:
As discussed earlier in https://github.com/OpenConext/OpenConext-engineblock/pull/337 minifying CSS/JS doesn't add much because of GZip-compression, and makes the code pretty much unreadable.


tvdijen:
Related: #339

MKodde:
This pull request reminds me of [a discussion](https://stackoverflow.com/questions/16196358/why-do-people-minify-css-and-javascript-when-we-have-gzip) that I read about some time ago. TLDR: Yes, GZip compression is superior when it comes to serving smaller sized files, but minifying JS and CSS offers other important features:

* Removal of possible sensitive comments and doc-blocks
* Influences [search engine ranking](https://developers.google.com/speed/docs/insights/MinifyResources)
* Possibility to concatenate several css / js files (reduction in # requests) 

For production 'builds' I would suggest utilizing both techniques. When it comes to development, [source maps](http://blog.teamtreehouse.com/introduction-source-maps) might prove very useful in this situation. Luckily this feature is built into the [Grunt contrib module](https://github.com/gruntjs/grunt-contrib-uglify/blob/master/README.md#sourcemap) we use for minification.

tvdijen:
>    Removal of possible sensitive comments and doc-blocks
Not really useful on open source projects
>    Influences search engine ranking
Not really useful in this case;  we always get redirected to EB by an SP
>    Possibility to concatenate several css / js files (reduction in # requests)
We can still do that without minifying

MKodde:
>>    Removal of possible sensitive comments and doc-blocks

> Not really useful on open source projects

Why not? The original sources will keep their comments and doc-blocks. Contributors can still read and write comments while end users aren't.

>>    Influences search engine ranking

> Not really useful in this case;  we always get redirected to EB by an SP

I'm with you on this one.

>>    Possibility to concatenate several css / js files (reduction in # requests)

> We can still do that without minifying

True that.

PR #438, opened 2017-07-26
Remove redeclaration of $application and fix indentation

tvdijen:
$application is already defined/initialized on the first line of this function. No need to do it twice


thijskh:
Thanks!

PR #440, opened 2017-07-31
Add documentation for trusted proxy

pmeulen:


docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
+# Trusted Proxy in EngineBlock
+
+In version 4.3.3 support for "trusted proxies" was added to Openconext. A "trusted proxy" is a SAML SP that uses engine as it's IdP. The difference between a normal SP and a SP is that is a trusted proxy is that the latter is allowed to impersonate other service providers. This allows a SP proxy to be connected to engineblock while using the ACL, ARP, consent and pseudonymisation functions of enginblock.
```
thijskh: its

docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
...
+## Engineblock Trusted Proxy behaviour
+
+Engine block will only enable trusted proxy processing for a SAML AuthnRequest that it receives from a SP when all of the following conditions are met:
+* The AuthnRequest has a valid signature
+* "coin:trusted_proxy" is enabled in the SP Entity configuration in engineblock. This SP is identified by the value of the /AuthnRequest/Issuer element in the SAML AuthnRequest.
+* The SAML AuthnRequest contains at least one /AuthnRequest/Scoping/RequesterID element.
+
+### Trusted proxy processing
+
+The imange below show a proxy (a trusted proxy) that is connected to engineblock :
```
thijskh: image, shows

docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
...
+In version 4.3.3 support for "trusted proxies" was added to Openconext. A "trusted proxy" is a SAML SP that uses engine as it's IdP. The difference between a normal SP and a SP is that is a trusted proxy is that the latter is allowed to impersonate other service providers. This allows a SP proxy to be connected to engineblock while using the ACL, ARP, consent and pseudonymisation functions of enginblock.
+
+
+## Enabling
+
+Trusted proxy is enabled per SP. To enable trusted proxy behaviour for a SP, set "coin:trusted_proxy" to true for the configuration of the SP. The default for this setting is false.
+
+## Engineblock Trusted Proxy behaviour
+
+Engine block will only enable trusted proxy processing for a SAML AuthnRequest that it receives from a SP when all of the following conditions are met:
```
thijskh: Engineblock

docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
...
+* The AuthnRequest has a valid signature
```
thijskh: Must `redirect.sign` be set? Or is that not relevant?

docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
...
+* The __SP being proxied__ is the SP behind the trusted proxy.
+
+![trusted_proxy](trusted_proxy.png)
+
+When processing a AuthnRequest from a trusted proxy engineblock performs some actions as if the SP being proxied sent the AuthnRequest directly. This is what differentiates trusted proxy processing from the normal processing in engineblock, which is also proxy aware but will never allow impersonation of another SP or reveal pseudonyms of another SP. For trusted proxies engineblock:
+* Generates the NameID for the SP being proxied and adds this to the eduPersonTargetedID attribute. This makes the persistent NameID (which is a pseudonym targeted at a SP) of the SP being proxied available to the trusted proxy. The NameID for the trusted proxy is available in the Subject in the Assertion. This way the trusted proxy can have an unique identifier for the user. When generating the Assertion for the SP being proxied the trusted proxy must copy the NameID from the eduPersonTargetedID to the Subject.
+* Consent is asked and remembered for the SP being proxied
+* The attribute manipulations of the SP being proxies are run
```
thijskh: + The ARP of the original SP is being applied?

thijskh: proxied

thijskh: What about PEP, AA if configured for the SP?

docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Trusted proxy processing details
+
+The SAML AuthnRequest below is en example of the AuthnRequest that a trusted proxy sends to engineblock. In this request:
```
thijskh: an

docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
...
+* The trusted proxy SP has an entityID of "https://trusted-proxy.example.com/metadata". This is the value of the `/AuthnRequest/Issuer` element in the SAML AuthnRequest to engineblock.
+* The SP being proxied has en entityID of "https://sp-being-proxied.example.net/metadata". Engineblock uses the value of the _last_ `/AuthnRequest/Scoping/RequesterID` element in the SAML AuthnRequest to engineblock. This means that a trusted proxy must ensure that its RequesterID element is the last RequesterID in the list.
```
thijskh: What happens with RequesterIDs other than the first one? Are they ignored?

thijskh:
Looks like a good explanation of the phenomenon.
Was wondering if we'd not place this in the [OpenConext wiki](https://github.com/OpenConext/OpenConext-engineblock/wiki) instead.

docs/trusted_proxy.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+Processing of the request:
+* Both the trusted proxy SP and the SP being proxied must be known to engineblock
+* Both the trusted proxy SP and the SP being proxied must have the same workflow state
+
+* The ACL of both the trusted proxy SP and the SP being proxied are verified. Only IdPs are allowed access to both SPs are allowed to login
+* The ARPs of both the trusted proxy SP and the SP being proxied are applied. Only attributes and attribute values that are allowed by both ARP are included in the response
+* The attribute manipulations (AMs) of both the trusted proxy SP and the SP being proxied are run. The AMs of the trusted proxy SP are run first.
+
```
quartje: The PEP and AA are run as well I assume?

pmeulen: That is a good question. I'd have to look into that. What behaviour do we want?
- You'd want the AA to be run for both  proxy SP and the SP being proxied, order should be specified.
- It makes sense to run the PEP for both the proxy SP and the SP being proxied, order does not matter

thijskh: Would indeed run both if applicable. Both are guarded by an enable-setting so there's no cost when not using it.

In any case of course the doc should describe what happens now - the desired state can be a backlog story if it deviates currently.

thijskh: Note that AA is closely related to ARP in our current architecture, so if both ARPs are applied then that implies any AA in either of these ARP's shiuld be run.

quartje: Some trusted proxies are SP's as well: In the case of the OpenConext OIDC gateway, the proxy is the same entity as the admin interface of that proxy. In that particular use case, it makes more sense to run the PEP for only the SP being proxied (which it currently does if I am not mistaken).

docs/trusted_proxy.md
pmeulen: For this `redirect.sign` is not relevant

quartje: I seemed to remember that I had some issues with a trusted proxy when the redirect.sign is not set for that trusted proxy. I just checked, when the redirect.sign is not set for the trusted proxy, the ARP is not applied for the proxied SP. However, the ACL is correctly applied. redirect.sign should thus be set for the trusted proxy.

docs/trusted_proxy.md
pmeulen: Some filter commands will take them into account, others will not. If scoping is enabled, the RequestID of engineblock will be prependend and send in the AuthnRequest to the IdP.

pmeulen:
Because this behaviour is version dependent I think it's better to leave this description close to the code. It is also more likely to be updated/seen during development. We can link to it from the wiki of course.

PR #441, opened 2017-08-01
Add documentation for Attribute Aggregation

thijskh:


docs/attribute_aggregation.md
```diff
@@ -0,0 +1,69 @@
...
+It will return the resulting attributes to Engineblock.
+
+Engineblock will always _replace_ the contents of any
+attribute that has a source specified with the contents
+from the attribute aggregator. That means that if
+aggregation is enabled for attribute _x_, any value
+received from the IdP for _x_ is discarded and
+only what is received from the AA for _x_ is sent on
+to the IdP (even when the result from the AA for _x_
+is empty, the SP will hence also not receive any value
```
pmeulen: How an attribute for which EB does not receive a value from the AA expressed in the SAML AttributeStatement. Is the Attribute omitted, will it have an empty value?

thijskh: Will add the information that it is then omitted.

PR #448, opened 2017-09-12
Drop obsolete group_provider database tables.

thijskh:


MKodde:
The proposed change looks good!

There however seems to be an issue with the behat tests (the build is failing). Some changes might be needed in that department.

In addition the security checker also notifies us that the currently installed version of simplesamlbundle has got an XSS vulnerability in the consentAdmin module. If this module is in use, this should be fixed in another PR. @thijskh, do you know whether or not this module is used throughout the openconext product range?

jorissteyn:
Merged. Did require a fix though: the name (timestamp) of the migration was already taken, and did not match the name of the file.

PR #461, opened 2018-01-09
Expose the /read-arp endpoint on EB API

MKodde:
The arp configuration of entities can now be read from the EB API using
the /read-arp endpoint. The endpoint expects to be called using POST.
An array of entityIds must be provided to specify what Arp config should
be read.

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
@@ -104,4 +104,35 @@ final class AttributeReleasePolicyController
 
         return new JsonResponse(json_encode($releasedAttributes));
     }
+
+    /**
+     * @param Request $request
+     * @return JsonResponse
+     */
+    public function readArpAction(Request $request)
+    {
+        if (!$request->isMethod(Request::METHOD_POST)) {
+            throw ApiMethodNotAllowedHttpException::methodNotAllowed($request->getMethod(), [Request::METHOD_POST]);
+        }
+
+        if (!$this->authorizationChecker->isGranted('ROLE_API_USER_PROFILE')) {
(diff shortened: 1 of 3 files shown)


src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
```diff
@@ -104,4 +104,35 @@ public function applyArpAction(Request $request)
...
+            throw ApiMethodNotAllowedHttpException::methodNotAllowed($request->getMethod(), [Request::METHOD_POST]);
+        }
+
+        if (!$this->authorizationChecker->isGranted('ROLE_API_USER_PROFILE')) {
+            throw new ApiAccessDeniedHttpException('Access to the ARP API requires the role ROLE_API_USER_PROFILE');
+        }
+
+        $body = JsonRequestHelper::decodeContentAsArrayOf($request);
+
+        if (!isset($body['entityIds'])) {
```
jorissteyn: Later on we assume this is an array, so why not is_array()?

src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
```diff
@@ -104,4 +104,35 @@ public function applyArpAction(Request $request)
...
         return new JsonResponse(json_encode($releasedAttributes));
     }
+
+    /**
+     * @param Request $request
+     * @return JsonResponse
+     */
+    public function readArpAction(Request $request)
+    {
+        if (!$request->isMethod(Request::METHOD_POST)) {
```
jorissteyn: POST is not appropriate here!

MKodde: This has been adressed

PR #463, opened 2018-01-12
Code quality

tvdijen:
I went through Scrutinizer and fixed quite a bunch of issues.

Mostly:
- Phpdoc
- Explicit type checking
- PSR-2
- Unused code


library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -47,7 +47,7 @@ public function serve($serviceName)
         $log->notice(sprintf(
             'Received Assertion from Issuer "%s" with signature method algorithm "%s"',
             $receivedResponse->getIssuer(),
-            $receivedResponse->getAssertion()->getSignatureMethod()
+            $receivedResponse->getSignatureMethod()
```
jorissteyn: Just as a side-note: in theory, the signature method of the assertion is not necessarily the same as the signature method of the message.

jorissteyn:
Thanks @tvdijen, especially like the strict type checking. Merging.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
tvdijen: Perhaps I've missed something, but CI was complaining about a non-existing method:

afbeelding

thijskh: This is indeed something different. It now logs the algorithm used for signing the Response, while it used to log the method in the assertion.

Scrutinizer is mistaken, it should not look in SAML2_Response but in EngineBlock_Saml2_ResponseAnnotationDecorator.

Ideally we should log both, btw.

tvdijen: Well, the return-type for `receiveResponse()` is:
`@return bool|EngineBlock_Saml2_ResponseAnnotationDecorator|SAML2_Response` ...

The whole trace is quite complex, however, I was able to trace it back to https://github.com/OpenConext/OpenConext-engineblock/blob/fc752ac77796d77baa363c3bb04ddd14faec14d3/library/EngineBlock/Corto/Module/Bindings.php#L565 which allows `$receivedResponse` to be of type `EngineBlock_Saml2_MessageAnnotationDecorator`...

The Scrutinizer-warning may be inaccurate, but I think it's right about the fact that `getAssertion()` does not exist for all paths leading up to this point.

thijskh: Scrutinizier is right from a static analysis point of view: the method can return three different types and we assume one specific in that method. However, in this place in the code we know it will always have an object of type `EngineBlock_Saml2_MessageAnnotationDecorator` because the other two return types logically do not happen at this point. This is not something you can know from static analysis, but from analyzing when it is called in the EB process.

The root cause is that this is rather spaghetti-esque code with a method with three different return types that is called in several places with different types of input. So the proper solution is to refactor this entirely and create methods that by their signature are much more predictable. Perhaps we can do this at some point in the future but there's no quick fix for that.

But in a practical sense there is no bug in the current code (we've never seen any fatal errors occur here). And the change changes the behaviour, because it used to log the hash algorithm of the Assertion and now logs that of the response.

thijskh: In #464 I've proposed a change that will at least log both sigalgs when present so it's more explicit that the algorithm can be in two places.

PR #464, opened 2018-01-17
Log sigalg for both Response AND Assertion.

thijskh:
A SAML response can have a signature in the Response, in the Assertion
contained therein, or both. Log all information we have.


jorissteyn:
Great, merging!

PR #465, opened 2018-01-18
Upgrade of simplesamlphp

MKodde:
Pear class names have been renamed/replaced to use the PSR2 naming
convention.

XMLSecurityKey is now loaded from the RobRichards\XMLSecLibs namespace.

When possible SimpleSAML_ classes have been loaded
from the appropriate namespace. Not all SimpleSAML
classes where moved to a PSR namespace. For example
the SimpleSAML_Configuration class still exists.


jorissteyn:
As discussed, I've updated some docblocks in EngineBlock_Ssp_sspmod_saml_Message and rebased against master (to include recent changes in composer.lock on master).

Attribute manipulators are not compatible with this change: if they read or write name IDs, they should be changed to used the new NameID object in the SAML2 library.

PR #467, opened 2018-01-19
Create signatures using a configurable signature method

jorissteyn:
A signature method can be configured per SP and IDP using the
'coin:signature_method' metadata property. If this property is not
set, RSA_SHA1 is used in order to maintain backwards compatibility.

Depends on https://github.com/OpenConext/OpenConext-engineblock-metadata/pull/19 (already merged).

### DIFF
--- a/database/DoctrineMigrations/Version20180118115853.php
+++ b/database/DoctrineMigrations/Version20180118115853.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace OpenConext\EngineBlock\Doctrine\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180118115853 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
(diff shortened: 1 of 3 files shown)


jorissteyn:
Build is failing, but works on my machine. Needs more work!

database/DoctrineMigrations/Version20180118115853.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 ADD signature_method VARCHAR(255) NOT NULL');
```
MKodde: The new column is not nullable. Shouldn't you update the existing entries?

jorissteyn: Didn't think of that, a push was required after deployment to make this work. I'm now doing this in the migration:

            UPDATE sso_provider_roles_eb5 SET signature_method = "http://www.w3.org/2000/09/xmldsig#rsa-sha1"

So no need to push metadata in production after deploy.

PR #468, opened 2018-01-19
Disallow authn requests signed with forbidden signature methods

jorissteyn:
EngineBlock now allows configuring one or more 'forbidden signature
methods'. If an authentication request is signed with such a method,
an error is logged and the generic error page is shown to the user.

Before this commit, the notice 'Received signed AuthnRequest from
Issuer...' was only logged when the signature was contained inside the
SAML message, but not when the signature was sent as a query
parameter. This commit changes this behaviour and logs the used
signature method regardless of how the signature was sent.

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
+++ b/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
@@ -44,12 +44,7 @@ class EngineBlock_Corto_Module_Service_AssertionConsumer implements EngineBlock_
         $application = EngineBlock_ApplicationSingleton::getInstance();
         $log = $application->getLogInstance();
 
-        $log->notice(sprintf(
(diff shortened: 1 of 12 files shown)


library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -126,4 +121,56 @@ public function serve($serviceName)
             $this->_server->sendResponseToRequestIssuer($receivedRequest, $newResponse);
         }
     }
+
+    /**
+     * Log and verify the signature methods used.
+     *
+     * The signatures are not validatd here, but the used methods (algorithms)
```
thijskh: typo 'validatd'

jorissteyn: Fixed.

PR #470, opened 2018-01-25
Show correct name of SP in WAYF

MKodde:
When dealing with trusted proxies, the incorrect SP name is shown in the
WAYF. The name of the proxy is shown instead of the name of the original
service.

By reading the RequesterId's from the AuthnRequest, we are able to
ascertain the first requester in the chain.

@thijskh is my solution to read the original SP from the requesterId array correct? Or is there a more obvious solution?

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
+++ b/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
@@ -3,6 +3,7 @@
 use OpenConext\Component\EngineBlockMetadata\Entity\IdentityProvider;
 use OpenConext\Component\EngineBlockMetadata\Entity\ServiceProvider;
 use OpenConext\EngineBlock\Logger\Message\AdditionalInfo;
+use Psr\Log\LoggerInterface;
 use SAML2\AuthnRequest;
 use SAML2\Response;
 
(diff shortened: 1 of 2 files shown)


library/EngineBlock/Corto/Module/Service/SingleSignOn.php
```diff
@@ -402,6 +404,37 @@ protected function _showWayf(EngineBlock_Saml2_AuthnRequestAnnotationDecorator $
...
+
+        // If requesterIDs are present in the scoping of the AuthnRequest, get the first requesterId and treat it as
+        //the service provider the user is authenticating for.
+        if (!empty($authnRequest->getRequesterID())) {
+            $originalRequesterId = reset($authnRequest->getRequesterID());
+            try {
+                $serviceProvider = $this->_server->getRepository()->fetchServiceProviderByEntityId(
+                    $originalRequesterId
+                );
+            } catch (EntityNotFoundException $e) {
```
tvdijen: Would we even end up at the WAYF in the event of an unknown entity?

MKodde: If this is the case, the SpProxy.feature scenario: `User logs in via untrusted proxy accesses discovery for unknown SP` does not reflect the reality.

```behat
  Scenario: User logs in via untrusted proxy accesses discovery for unknown SP
    Given SP "Step Up" is authenticating and uses RequesterID "https://example.edu/saml2/metadata"
     When I log in at "Step Up"
     Then I should see "AlwaysAuth"
      And I should see "CombinedAuth"
      And I should see "StepUpOnlyAuth"
      And I should not see "LoaOnlyAuth"
```

@thijskh can you shed your light on this?

thijskh: I guess what is caught earlier are unknown issuers (error message "Unknown Issuer").

tvdijen: I was refering to the 'Unknown Service' error..

MKodde:
@thijskh 
Yes, using the `EngineBlock_SamlHelper` is a good idea in this context thanks for the tip. This eliminated the need for testing all kinds of error conditions. 

@tvdijen, did this change answer your concern?

PR #472, opened 2018-01-29
Make Profile url configurable in application.ini

MKodde:
A new configuration parameter was added to the application.ini. This config is read in the ProvideConsent service and injected into the consent view. The profile url is then fed into the translation helper who in turn sprintf's it into the translation string.

When no application ini is not configured with the profile url, '#' is used as a fallback value, rendering the link useless.

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -196,7 +196,14 @@ pdp.password = "secret"
 pdp.client_id = "EngineBlock"
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;;;;;;;;;;; ATTRIBUTE AGGREGATION SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;; PROFILE SETTINGS ;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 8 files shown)


application/configs/application.ini
```diff
@@ -196,7 +196,13 @@ pdp.password = "secret"
 pdp.client_id = "EngineBlock"
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;;;;;;;;;;; ATTRIBUTE AGGREGATION SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;; PROFILE SETTINGS ;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+profile.baseUrl = "https://profile.surfconext.nl/"
```
thijskh: I'd prefer `profile.example.org` while we're at it (we should clean up other occurrences of SURFconext in config at some point, but let's not add new ones to it).

MKodde: Consider it done.

MKodde:
OpenConext deploy was also instructed to provision this parameter to the engineblock.ini file in this [PR](https://github.com/OpenConext/OpenConext-deploy/pull/163)

PR #474, opened 2018-01-30
Prevent scrolling to Consent button

MKodde:
When the consent button is out of view (smaller screens) the focus on
the button caused the screen to scroll down. Cutting of the top of the
page. This change tests if the provide consent button is visible on the
page, if it is. The focus is placed on it. Otherwise it isn't.


jorissteyn:
Works fine, merging (and I've changed let to const).

MKodde:
Note that the build for this PR failed due to the security checker notifiying us to update SimpleSAMLPHP. This is addressed in #475 .

PR #476, opened 2018-02-02
Use release identifier as cache buster

jorissteyn:
The makeRelease.sh script replaces a placeholder in config.yml when
creating a tarball, setting the 'asset_version' parameter to the tag,
branch or commit hash used to create the release. This value is used
as cache buster in the html template.

See https://www.pivotaltracker.com/story/show/154839009

### DIFF
--- a/app/Resources/views/layouts/scripts/default.phtml
+++ b/app/Resources/views/layouts/scripts/default.phtml
@@ -20,7 +20,7 @@
   <meta content="yes" name="apple-mobile-web-app-capable">
   <meta content="translucent-black" name="apple-mobile-web-app-status-bar-style">
   <link href="/favicon.ico" rel="shortcut icon">
-  <link href="/stylesheets/application.css?v=5.5.2" rel="stylesheet" type="text/css" />
+  <link href="/stylesheets/application.css?v=<?= EngineBlock_View::htmlSpecialCharsAttributeValue($this->layout()->assetVersion) ?>" rel="stylesheet" type="text/css" />
   <!-- Firefox CSS -->
   <style>
     @-moz-document url-prefix() {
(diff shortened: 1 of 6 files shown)


thijskh:
So it's not the commit hash, but the tag (e.g. "5.5.2"), right?

tvdijen:
Do I really have to set this manually for every install from source??

thijskh:
No, `bin/makeRelease.sh` will set it so any tarball will come with the tag set to the right value.

tvdijen:
That's why I ask... I never use the official tarballs but build from source

thijskh:
If you do that then you already have to run commands post-checkout like ‘composer’, so you could add some command to it.

But strictly speaking you’re not required to do anything new, you’ll always get the value “dev” (retaining the old behaviour).

tvdijen:
I don’t dev, I get #ASSET_VERSION# ...   shouldn’t this default to the .ini’s version setting?

MKodde:
I'd like to join this discussion. @tvdijen, is using (or manually building) tarballs not an option? If it is not, we can try to meet in the middle. For example, using a more sensible default value for the asset version replacement token. 

I'm not aware of a version number in the `application.ini`, maybe I'm overlooking something here that could be taken advantage of. Please explain.

tvdijen:
Tarballs are not really an option, because I have too many customizations to apply..
I always clone the release-tag and merge in all the customizations, regenerate theme assets, etc..

The version-number I was referring to:
https://github.com/OpenConext/OpenConext-deploy/blob/master/roles/engineblock/templates/engineblock.ini.j2#L2

MKodde:
I will try to look into using this version number instead of the one used to build a release with. 

A limitation with the usage of the `engineblock.ini`'s `version` is that we'll/you'll have to update it every time a new release is installed (as I assume you do not reprovision eb-deploy for each release), otherwise cache busting will still fail.

tvdijen:
I don't use eb-deploy at all, because unfortunately I can't use ansible here...
Anyway, I already change the version number in `engineblock.ini` as I was under the impression that it is already required to do so...

PR #479, opened 2018-02-05
Remove unused metadata backends

jorissteyn:


src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
```diff
@@ -27,7 +27,7 @@ class ServiceProvider extends AbstractRole
      *
      * @ORM\Column(name="attribute_release_policy", type="array")
      */
-    protected $attributeReleasePolicy;
+    public $attributeReleasePolicy;
```
MKodde: Why make this property `public` when there is a dedicated getter? All other fields are publically available, in that light I see the logic in this change. But then I would expect the removal of the `getAttributeReleasePolicy` method.

src/OpenConext/EngineBlockBundle/Controller/Api/ConnectionsController.php
```diff
@@ -77,11 +71,7 @@ public function pushConnectionsAction(Request $request)
 
         $assembler = new JanusPushMetadataAssembler();
```
MKodde: This Controller needs to be cleaned up, and after looking at it for a second, might be a candidate for total removal.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
```diff
@@ -177,8 +176,6 @@ public function anEngineblockInstanceConfiguredWithJsonData()
             ->addSpsFromJsonExport($this->spsConfigUrl)
```
MKodde: `addSpsFromJsonExport` does not exist? Or is invoked magically. Please investigate this.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
```diff
@@ -128,72 +128,6 @@ public function anUnregisteredServiceProviderNamed($name)
         return $mockSp;
     }
 
-    /**
-     * @Given /^SP "([^"]*)" may only access "([^"]*)"$/
-     */
-    public function spMayOnlyAccess($spName, $idpName)
```
MKodde: * Remove `LogChunkParser` usages (class no longer exists) 
* Remove `IdFixture` usages
* Cleanup unused imports

src/OpenConext/EngineBlockBundle/Controller/Api/ConnectionsController.php
```diff
@@ -39,16 +33,16 @@ class ConnectionsController
...
     public function __construct(
         AuthorizationCheckerInterface $authorizationChecker,
         FeatureConfiguration $featureConfiguration,
-        EngineBlock_ApplicationSingleton $engineBlockApplicationSingleton
+        DoctrineMetadataRepository $repository
     ) {
-        $this->engineBlockApplicationSingleton = $engineBlockApplicationSingleton;
         $this->authorizationChecker            = $authorizationChecker;
         $this->featureConfiguration            = $featureConfiguration;
+        $this->repository                      = $repository;
```
MKodde: We should not rely on a dynamically created field here. I propose the addition of a private repository field.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
MKodde: I ended up cleaning up the `anEngineblockInstanceConfiguredWithJsonData` method altogether as it was never used and would pose problems when it might be used in the future.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
MKodde: Fixed this by removing the no longer used `spMayRunInTransparentModeIfIndicatedIn` method.

PR #483, opened 2018-02-07
Move arp field declaration to AbstractRole

MKodde:
The $attributeReleasePolicy field declaration was set on the
 ServiceProvider implementation. The IdentityProvider also exposed this
 field. PHP generously dynamically published this field publicly. To
 get a consistent result, the field was moved from ServiceProvider to
 the AbstractRole.


jorissteyn:
As discussed, it might be better to just remove arp from IdentityProvider, since idps don't use arp.

jorissteyn:
Closing this one, the actual problem (unable to statically detect type errors) is addressed in #487.

PR #486, opened 2018-02-08
Remove find/fetchEntityByEntityId methods

MKodde:
These methods are bugs waiting to happen. An entityID should only be
 unique on a role basis. Technically an IdP and an SP could have the
 same EntityID.

This PR removes these two methods and replaces any instance in the code
 with a call to the relevant findSP or findIdP counterpart.

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/Metadata.php
+++ b/library/EngineBlock/Corto/Module/Service/Metadata.php
@@ -1,6 +1,7 @@
 <?php
 
 use EngineBlock_Corto_Module_Service_Metadata_ServiceReplacer as ServiceReplacer;
+use OpenConext\EngineBlock\Metadata\MetadataRepository\EntityNotFoundException;
 
 class EngineBlock_Corto_Module_Service_Metadata extends EngineBlock_Corto_Module_Service_Abstract
 {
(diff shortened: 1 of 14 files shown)


library/EngineBlock/Corto/Module/Service/Metadata.php
```diff
@@ -11,7 +12,12 @@ public function serve($serviceName)
         $engineEntityId = $this->_server->getUrl($serviceName);
         $this->_server->unsetProcessingMode();
 
-        $engineEntity = $this->_server->getRepository()->fetchEntityByEntityId($engineEntityId);
+        // TODO: Are we dealing with an SP or IdP at this point? The previous solution suggests it can be both
```
MKodde: I'm going to investigate this in more detail. But if anybody can point me in the right direction, that would be great.

@thijskh

thijskh: I guess it can be both? This code is about producing the feeds as listed on EB's "homepage".

MKodde: Thanks!

MKodde: I've removed the TODO comment and made the method more robust in this commit: 3ab3188

PR #488, opened 2018-02-13
Reverse name/displayName order

tvdijen:
DisplayName should have preference over the technical name


tvdijen:
Ah crap, let me fix that

tvdijen:
Fixed

thijskh:
Don't think JS should be changed...

tvdijen:
Well, that's what happens when I run npm install && grunt...
Changes to the file seem syntactical only, so I suspect npm installed a newer version of a dependency..

thijskh:
Personally I'd prefer to not "git add" that change then. Keeps it more clean.

tvdijen:
As you wish

PR #493, opened 2018-02-19
Remove dbpatch: the complete SQL schema is managed with doctrine

jorissteyn:
All installations up-to-date with dbpatch can now run
doctrine:migrations:diff and see no changes. All changes from here on
forward can be managed with doctrine.
    
WARNING: This commit breaks the upgrade path from EB4.x. So make sure
your installation is upgraded to any 5.x version before upgrading to
5.7.
    
See: https://www.pivotaltracker.com/story/show/148324779


jorissteyn:
https://github.com/OpenConext/OpenConext-deploy/pull/166

PR #499, opened 2018-02-22
Remove unused simplesamlphp configuration

jorissteyn:
The authsources and metadata configuration files were in use when
profile was still part of EngineBlock. It should now be possible to
remove them.
    
The config.php file is required for some basic configuration like
logging and timezone configuration, but most settings are not
applicable for how EngineBlock uses simplesamlphp. All configuration
related to this file (auth.simplesamlphp, engine.simplesamlphp) are
removed to make it clear they have no effect in EngineBlock.


jorissteyn:
This needs some more thought!

PR #502, opened 2018-02-23
Use symfony DI parameters for all configuration

jorissteyn:
There are currently three ways of accessing configuration options:

 1. Some INI configuration is converted to yaml and accessed through the symfony DI container
 2. All INI configuration is loaded into Zend_Config and accessed through the application singleton
 3. The "engineblock bridge" config component as a way of accessing INI configuration directly from within the symfony DI container

This PR removes the 2nd and 3rd way of accessing configuration, leaving the symfony DI container the single source for configuration. As a result it is now very transparent which configuration options are used by what component in EngineBlock. Unused configuration options are removed from the default application.ini.

It is now very easy to stop using the INI files altogether, and leave the YAML format as the only supported configuration format. The INI files are only used when preparing the environment ("prepare-env") and no longer parsed runtime. After running `prepare-env`, the file  `app/config/ini_parameters.yml` now shows all options used by EngineBlock.

All changes are backwards compatible, no changes to application.ini or /etc/openconext/engineblock.ini are required when upgrading to this version of EngineBlock. The file `UPGRADING.md` contains a list of options which can be removed from INI files in order to clean them up.

https://www.pivotaltracker.com/story/show/155358923


application/configs/application.ini
```diff
@@ -211,12 +211,6 @@ email.idpDebugging.subject    = "IdP debug info van %1$s"
 ; terms of use surfconext
 openconext.termsOfUse = "https://wiki.surfnet.nl/display/conextsupport/Terms+of+Service+%28EN%29"
 
-; Which LDAP attribute to use as the primary identifier.
-; EXPERT use only!
-; Recommend only using one of the following possible values:
-; 'collabpersonid', 'collabpersonuuid', 'edupersonprincipalname'
-subjectIdAttribute = "collabpersonid"
```
thijskh: This is used in PR #383. So we might want to keep it if we want to merge that PR (sometime), or drop it and close the PR

jorissteyn: Sharp! I've reverted the commit.

PR #505, opened 2018-02-26
Apply security update to paragonie/random_compat

jorissteyn:
See: https://github.com/paragonie/random_compat/issues/96

This vulnerability is actually two years old, but has recently been
added to the sensio security checker database. EngineBlock was not
affected because the insecure openssl function is only used when all
other providers (like mcrypt) are not installed.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1617,16 +1617,16 @@
         },
         {
             "name": "paragonie/random_compat",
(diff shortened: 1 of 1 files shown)


kaystrobach:
will there be a new 1.4.x version which makes it possible to disable the openssl stuff?

PR #510, opened 2018-02-28
Do not validate EngineBlocks own hostname

jorissteyn:
It is not EngineBlocks responsibility to validate it's own
hostname. The webserver should be configured to allow specific
hostnames under which EngineBlock is served.

### DIFF
--- a/library/EngineBlock/ApplicationSingleton.php
+++ b/library/EngineBlock/ApplicationSingleton.php
@@ -325,13 +325,7 @@ class EngineBlock_ApplicationSingleton
             return $configHostname;
         }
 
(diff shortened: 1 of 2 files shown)


thijskh:
I think this is not really defense-in-depth. You're right in theory but e.g. Apache does not do validation of this header. Before we use such an externally obtained value, I'd prefer a basic sanity check.

jorissteyn:
I'm working on a new PR to address your issue. Simply restoring the validation removed here is not ideal because in some places the HTTP_HOST value is used directly without validation.

Proposal: https://www.pivotaltracker.com/story/show/155617973

PR #497, opened 2018-02-21
Installation and implementation of Twig in EngineBlock

MKodde:
This PR is part of the changes described in [Pivotal](https://www.pivotaltracker.com/story/show/155358923)

In this PR:
* [x] Install Twig
* [x] Update the EB welcome page with a new Twig template as proof of concept.
* [x] Rewrite of all Zend View based templates to Twig.
* [x] Removal of 'cached' views from the `/app/Resources/views/*` folder
* [x] Added unit tests for some of the Twig extensions
* [x] Fully functional language switcher.
* [x] Updated UPGRADING.md and [Development guidlines wiki page](https://github.com/OpenConext/OpenConext-engineblock/wiki/Development-Guidelines)

Not in this PR:
* Support for the `transChoise`. This might be addressed in an upcoming PR.

PR #514, opened 2018-03-01
Remove inline styles from Twig

MKodde:
Further cleanup after the Twig migrations of #497 . This PR aims to remove any inline styling from the Twig templates.

More info in the commit messages and the Pivotal story
https://www.pivotaltracker.com/story/show/155582419


MKodde:
Thanks for your review @tvdijen! The rogue whitespace has been taken care of.

PR #515, opened 2018-03-01
Require the 'hostname' INI setting

jorissteyn:
You can already configure this option in earlier versions of EB in
order to be forward compatible.

See PR #510 and https://www.pivotaltracker.com/story/show/155617973

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -12,6 +12,11 @@
 ;;;
 [base]
 
+base_domain = example.org
+
+; Set a fixed hostname for OpenConext EngineBlock to use.
+hostname = engine.example.org
+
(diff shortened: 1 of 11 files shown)


application/configs/application.ini
```diff
@@ -16,6 +16,9 @@
 ;;;;;;;;;;;; PHP SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+; Set a fixed hostname for OpenConext EngineBlock to use.
+hostname = engine.example.org
```
MKodde: This might not be the optimal place to put this setting as it is not a php setting in the sense that it overrides a php.ini setting.

jorissteyn: Fixed.

PR #516, opened 2018-03-01
Allow only POST requests on readArpAction

MKodde:
Previously we only allowed GET requests on the readArp endpoint. This
 limited the number of entity id's that could be passed to the endpoint
 as the query string is limited ot a certain length before it is cut off

By moving from a GET to a POST request we remedy this problem.

This change is made to be compatible with EB 5.5, hence the merge target: `release/5.5`

This PR is part of the fix for the following [bug](https://www.pivotaltracker.com/story/show/155497459)

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/Api/AttributeReleasePolicyController.php
@@ -111,18 +111,32 @@ final class AttributeReleasePolicyController
      */
     public function readArpAction(Request $request)
     {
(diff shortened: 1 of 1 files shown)


MKodde:
@jorissteyn the release branch: `release/5.5` is not configured to trigger Travis builds.

MKodde:
`.travis.yml` was updated to create PR builds on the current release branches (5.5, 5.6)

jorissteyn:
Merged into: 5.5.2, 5.5, 5.6 and master. Tagging releases next week when all profile changes are merged.

PR #511, opened 2018-02-28
Replace Zend_Translate with Symfony translator

jorissteyn:
This commit contains two important changes to make the translations
compatible with the Symfony translator:

 - translation files are renamed from languages/[lang].php to
   languages/messages.[lang].php

 - message placeholders are now using symfony-style named
   parameters ('%name%') instead of sprintf placeholders (%s, %d,
   %1$s)

EngineBlock users with a custom nl.php / en.php use instructions in
UPGRADING.md to easily convert the parameter style from the old to the
new format.

PR #518, opened 2018-03-02
Allow translation overrides and remove references to SURFconext

jorissteyn:
See the upgrade instructions for a detailed description of the
changes. In summary, the following words are dynamically injected into
the translations in order to have a sane set of default translations
not specifically targeted to the educational world:
    
 - %suiteName%:
   - OpenConext (default)
   - SURFconext, ACMEconext (example overrides)

 - %organisationNoun%:
   - organisation/organisatie (default)
   - institution/instelling (example overrides)

 - %organisationNounPlural%:
   - organisations/organisaties (default)
   - institutions/instellingen (example overrides)

 - %accountNoun%:
   - organisation account/organisatieaccount (default)
   - institutional account/instellingsaccount (example overrides)
    
The WAYF help page is a special case because the questions and answers
all have their own translation key, and those questions can be
considered specific to SURFconext. So the help page is now simplified
to just render one single "translatable" blob of HTML. The default
content for the help page is "No help content available" - this
content can of course be overridden in overrides.[lang].php.
    
See: https://www.pivotaltracker.com/story/show/155467582

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ bin/ignore_me.php
 /app/config/ini_parameters.yml
 /app/config/config_local.yml
 .idea
+/languages/overrides.*.php
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -64,6 +64,14 @@ configuration files and will be ignored otherwise.
     symfony.logPath
     symfony.cachePath
 
+### New configuration parameter openconext.supportUrl
+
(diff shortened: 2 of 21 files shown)


jorissteyn:
We've discussed this in #511, @tvdijen feedback is appreciated!

tvdijen:
By the looks of it, this is amazing!!
I'll properly test it once a new release is available

jorissteyn:
Sure! We're creating a "pre-release" next week.

PR #520, opened 2018-03-02
Send support URLs of SP to profile in the consent API

jorissteyn:
See: https://www.pivotaltracker.com/story/show/107474726


jorissteyn:
Merged into all current releases (5.5.2, 5.5, 5.6, 5.7).

PR #521, opened 2018-03-07
Support old and new NameId style in AM

MKodde:
The old style (array format) and new style (object of type NameID) are
 now supported in the ResponseAnnotationDecorator.

For more context see [Pivotal](https://www.pivotaltracker.com/story/show/155705429)

:information_source: Note that the build fails on the Security Checker. `paragonie/random_compat` can be patched (which is done in version 5.7) but is postponed for version 5.6 as we are not vulnerable.

### DIFF
--- a/library/EngineBlock/Saml2/ResponseAnnotationDecorator.php
+++ b/library/EngineBlock/Saml2/ResponseAnnotationDecorator.php
@@ -3,6 +3,7 @@
 use SAML2\Assertion;
 use SAML2\EncryptedAssertion;
 use SAML2\Response;
+use SAML2\XML\saml\NameID;
 
 /**
  * Annotate the response with our own metadata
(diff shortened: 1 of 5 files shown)


library/EngineBlock/Saml2/ResponseAnnotationDecorator.php
```diff
@@ -180,14 +192,20 @@ public function getOriginalNameId()
...
+                'Value' => $customNameId->value,
+            ];
+        }
         $this->customNameId = $customNameId;
         return $this;
     }
 
     /**
-     * @return array
+     * @return NameID
```
jorissteyn: As discussed: this is incorrect.

MKodde: Corrected. And added some extra unit tests as a bonus.

PR #525, opened 2018-04-23
Do not show all IDPs in WAYF when using experimental caching

jorissteyn:
This commit upgrades engineblock-metadata to 3.1.2, which provides a
fix for a bug where all IDPs where shown on the WAYF, unfiltered, when
the experimental caching metadata backend was used.

See:
 - https://github.com/OpenConext/OpenConext-engineblock-metadata/pull/24
 - https://www.pivotaltracker.com/n/projects/1425900/stories/156995448
 - OpenConext/OpenConext-engineblock#524

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,7 +1,7 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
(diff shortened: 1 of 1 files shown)


jorissteyn:
Ignoring unpatched random-compat , as described earlier, the problem has been there for several years, and only happens when not using openssl. We use the pached version of random-compat starting from EB5.7.

PR #527, opened 2018-04-30
Remove usages of SimpleSAMLphp

jorissteyn:
This pull request is a work-in-progress. The goal is to get rid of the dependency on SSP as wel as removing all code forked from SSP.


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -335,7 +326,7 @@ public function receiveResponse()
             }
 
             try {
-                $assertions = $className::processResponse($sspSpMetadata, $sspIdpMetadata, $sspResponse);
+                $assertions = EngineBlock_Ssp_sspmod_saml_SymfonyRequestUriMessage::processResponse($sspSpMetadata, $sspIdpMetadata, $sspResponse);
```
MKodde: Thanks, that makes it much clearer what class is actually used!

PR #528, opened 2018-05-14
Load initialize() when the page has been completely loaded

tvdijen:
Occasionally, users get the JavaScript-error "Uncaught TypeError: Cannot read property 'className' of undefined".
Some research showed that this occurs when the DOM-structure has not been loaded completely at script execution. This PR should prevent this (admittely somewhat edgecase) from happening.

js_error


tvdijen:
Somehow an upgrade to 5.7,0 fixed this...

PR #529, opened 2018-05-21
Update redirect.html.twig

tvdijen:
Fixes:
engineblock.ERROR: Unexpected token "string" of value "content" ("name" expected).

was encountered while `debug=true`


tvdijen:
That's the one I've touched in this PR :D Coffee?

tvdijen:
I do see however another issue in the `authorization-policy-violation` template...
I see a translation tag and the page is being cutoff (no footer) and when I click on a language (language bar), I get a completely different error page

jorissteyn:
Merging, I'm creating a new PR to address the issue you mentioned above.

PR #531, opened 2018-05-24
Fall back to name or entity ID for display name in consent API

jorissteyn:
If no diplay name is present in the metadata of a service provider,
use the name field or the entity ID instead.

See https://www.pivotaltracker.com/story/show/157775822

### DIFF
--- a/src/OpenConext/EngineBlock/Authentication/Dto/Consent.php
+++ b/src/OpenConext/EngineBlock/Authentication/Dto/Consent.php
@@ -52,10 +52,6 @@ final class Consent
 
         $serviceProvider = [
             'entity_id'    => $this->serviceProvider->entityId,
-            'display_name' => [
-                'en' => $this->serviceProvider->displayNameEn,
-                'nl' => $this->serviceProvider->displayNameNl,
-            ],
             'support_url' => [
                 'en' => $this->serviceProvider->supportUrlEn,
                 'nl' => $this->serviceProvider->supportUrlNl,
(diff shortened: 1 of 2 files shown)


src/OpenConext/EngineBlock/Authentication/Dto/Consent.php
```diff
@@ -67,10 +63,33 @@ function (ContactPerson $contact) {
...
+    {
+        if (!empty($this->serviceProvider->displayNameEn)) {
+            $fields['display_name']['en'] = $this->serviceProvider->displayNameEn;
+        } else if (!empty($this->serviceProvider->nameEn)) {
+            $fields['display_name']['en'] = $this->serviceProvider->nameEn;
+        } else {
+            $fields['display_name']['en'] = $this->serviceProvider->entityId;
+        }
+
+        if (!empty($this->serviceProvider->displayNameEn)) {
```
thijskh: This should be `displayNameNl`, also on line 87.

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/ConsentControllerTest.php
```diff
@@ -172,6 +172,7 @@ public function a_consent_listing_for_a_given_user_is_retrieved_from_the_consent
         $serviceProvider = new ServiceProvider($spEntityId);
         $serviceProvider->displayNameEn = 'My Test SP';
         $serviceProvider->displayNameNl = 'Mijn Test SP';
+        $serviceProvider->nameNl = 'Mijn Test SP';
```
thijskh: Is this a good test if displayname and name are the same string?

src/OpenConext/EngineBlock/Authentication/Dto/Consent.php
jorissteyn: Fixed and I've added a unit test covering every line in this file.

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/ConsentControllerTest.php
jorissteyn: True, but I've reverted the change since we test the specific rules of serialization in ConsentTest.php.

PR #532, opened 2018-05-24
Remove duplicate 'support_url' in consent API response

jorissteyn:


jorissteyn:
Sorry for the noise, at least there is absolutely no doubt I want to remove the duplicate support_url field.

PR #533, opened 2018-05-25
Store the support URLs to sso_provider_roles

jorissteyn:
See: https://www.pivotaltracker.com/story/show/157872470


jorissteyn:
I've updated manage using:

    cd OpenConext-deploy && git pull && ansible-playbook -i  environments/vm/inventory -u vagrant provision-vm.yml --extra-vars="secrets_file=environments/vm/secrets/vm.yml" --tags=manage

And confirmed this works as expected.

PR #534, opened 2018-05-29
Fix requirements

tvdijen:
When running bin/composer.phar install --no-dev:

PHP Fatal error:  Class 'Liip\FunctionalTestBundle\LiipFunctionalTestBundle' not found in /apps/components/engineblock/OpenConext-engineblock-5.7.0/app/AppKernel.php on line 37

It is actually a requirement

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         "ramsey/uuid": "^3.3.0",
         "sensio/generator-bundle": "^3.0",
         "guzzlehttp/guzzle": "^6.0",
+        "liip/functional-test-bundle": "^1.7",
         "openconext/monitor-bundle": "^1.0",
         "swiftmailer/swiftmailer": "^5.4",
         "symfony/swiftmailer-bundle": "^2.6",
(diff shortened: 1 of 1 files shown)


jorissteyn:
Hi Tim,

see `AppKernel.php`:

```
        if (in_array($this->getEnvironment(), array('dev', 'test'))) {
            $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
            $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
            $bundles[] = new Liip\FunctionalTestBundle\LiipFunctionalTestBundle();

            // own bundles
            $bundles[] = new OpenConext\EngineBlockFunctionalTestingBundle\OpenConextEngineBlockFunctionalTestingBundle();
        }
```

When running `composer install` without `--no-scripts`, the post-install scripts require a functioning EngineBlock. So if you use `--no-dev`, you must make sure you export `SYMFONY_ENV=prod`. For example, this should work:

    SYMFONY_ENV=prod composer install --no-dev

tvdijen:
Now that's why I shouldn't do upgrades at 1am :(
Thanks @jorissteyn !

PR #536, opened 2018-05-30
Remove autoloader bootstrap.php.cache

jorissteyn:
The file is automatically created by composer and should not be under
version control.


thijskh:
Add to `.gitignore` then?

jorissteyn:
I've now added the file to .gitignore.

PR #539, opened 2018-06-06
Fix error in PDP policy decision error page

jorissteyn:
PR #529 describes a bug in the policy-decision violation page.

EB5.7 streamlined all error pages: they now all show the 'feedback
info' table. But for PDP errors, the feedbackInfo session entry
contains an object holding the PDP failure - which cannot be cast to
string and should not be shown in the table.

This PR fixes the problem by not storing the PDP object into
'feedbackInfo', so the table can be safely rendered on the PDP error
page.

### DIFF
--- a/src/OpenConext/EngineBlockBridge/ErrorReporter.php
+++ b/src/OpenConext/EngineBlockBridge/ErrorReporter.php
@@ -75,10 +75,7 @@ class ErrorReporter
         if ($exception instanceof EngineBlock_Corto_Exception_ReceivedErrorStatusCode) {
             $feedback = array_merge($feedback, $exception->getFeedbackInfo());
         } elseif ($exception instanceof EngineBlock_Corto_Exception_PEPNoAccess) {
-            $feedback = array_merge(
-                $feedback,
-                ['error_authorization_policy_decision' => $exception->getPolicyDecision()]
-            );
+            $_SESSION['error_authorization_policy_decision'] = $exception->getPolicyDecision();
         }
 
         $_SESSION['feedbackInfo'] = array_merge(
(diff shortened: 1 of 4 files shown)


jorissteyn:
@tvdijen

tvdijen:
Tested it and it works like a charm!

thijskh:
So does this make the PEP error page look different? Previously it didn't have the table on it. I think that was on purpose (since it's not from a functional perspective an "error" in the sense that we need to be able to debug something - a decision was reached that you may not continue. More an information page, really?).

tvdijen:
I agree that this should probably not show a table-formatted error...
Either way, the previous behaviour was broken, so this is an improvement

MKodde:
I've updated the PR with the feedback of @thijskh, to test this: visit the PDP error page at: `https://engine.example.com/authentication/feedback/authorization-policy-violation`
And make sure you add some PDP response to your `error_authorization_policy_decision` session bag:

```
        $_SESSION['error_authorization_policy_decision'] = PolicyDecision::fromResponse(\OpenConext\EngineBlockBundle\Pdp\Dto\Response::fromData([
            'Response' => [
                [
                    'Status' => [
                        'StatusCode' => 500,
                        'StatusMessage' => 'Status message from the PDP',
                    ],
                    'PolicyIdentifier' => 'foobar',
                    'Decision' => 'go on',
                ]
            ],
        ]));
```

jorissteyn:
Almost there:

 * commit 1370d6a removed the message error description from the template (`git log -S error_authorization_policy_violation_desc`)
 * there was no back button pre 5.7

You can take a look at https://engine.surfconext.nl/authentication/feedback/authorization-policy-violation to spot the two differences.

MKodde:
Thanks @jorissteyn, fixed this in these two commits: ef8ef1b & 0f5c21d.

PR #541, opened 2018-06-15
Array of value types and array of values have different size

MKodde:
The Pivotal story describes the issue that is being fixed here in greater detail. In short, the attribute value type array needed to be updated to prevent the raising of an exception in the saml2 library.

The first commit adds behat tests that touch this problem. The second commit provides the bugfix for the issue.

The sec checker fails on the recent Symfony CVE's. I'm investigating the impact/benefits of a Symfony upgrade. And might open a new PR with the upgrade.

See: https://www.pivotaltracker.com/story/show/1581841791

### DIFF
--- a/library/EngineBlock/Arp/AttributeReleasePolicyEnforcer.php
+++ b/library/EngineBlock/Arp/AttributeReleasePolicyEnforcer.php
@@ -33,7 +33,9 @@ class EngineBlock_Arp_AttributeReleasePolicyEnforcer
                 continue;
             }
 
-            foreach ($attributeValues as $attributeValue) {
+            // The attributeIndex is kept to preserve the original keys in $newAttributes
+            // this to be able to remove the matching attribute value types
+            foreach ($attributeValues as $attributeIndex => $attributeValue) {
                 if (!$arp->isAllowed($attributeName, $attributeValue)) {
                     EngineBlock_ApplicationSingleton::getLog()->info(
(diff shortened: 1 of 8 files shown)


thijskh:
I think this solution is correct.

Am wonding though why testsuite is now failing.

Was also wondering, more as a sidenote, if you could not also solve it with array_intersect_key, e.g.
```php
function updateAttributeValueTypes(array $attributes, array $attributeValueTypes) {
    $newAttributeValueTypes = [];
    $attributeValueTypes = array_intersect_key($attributeValueTypes, $attributes);
    foreach($attributeValueTypes as $attributeIndex => $attributeValue) {
        $newAttributeValueTypes[$attributeIndex] = array_intersect_key($attributeValue, $attributes[$attributeIndex]);
    }
    return $newAttributeValueTypes;
}
```
but I guess that does not add much value over the method employed now.

MKodde:
Thanks for the review @thijskh. Feedback on your questions:

Regarding the failing build:
> The sec checker fails on the recent Symfony CVE's. I'm investigating the impact/benefits of a Symfony upgrade. And might open a new PR with the upgrade.

Upgrading Symfony in the current release did not seem to make much sense. The 5.7 release will, of course, have the latest security patches.

`array_intersect_key` could have been used. Thanks for noting this more elegant solution. Amended it to my commit.

PR #544, opened 2018-06-19
Implement a minimal variant of the consent screen

jorissteyn:
Manage before 2.0.18 sends a list of service provider entity IDs for
each identity provider indicating when not to ask for
consent. In newer versions of 2.0.18 manage sends configuration
objects for each service provider:

    "disable_consent_connections": [
    {
        "name": "https://serviceregistry.acc.surfconext.nl/simplesaml/module.php/saml/sp/metadata.php/default-sp",
        "type": "minimal_consent",
        "explanation:nl": "Explanation EN",
        "explanation:en": "Explanation NL"
    }

This commit changes the metadata storage backend to persist this
configuration and make it available to the ProviderConsent corto
module. A stripped version of the consent screen is shown when the
consent type is set to 'minimal_consent'. The differences between the
default and minimal consent screens are small, but important for legal
reasons.

See https://www.pivotaltracker.com/story/show/158145547


tests/unit/OpenConext/EngineBlock/Metadata/ConsentSettingsTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->assertEquals(
+            [
+                'https://example.org/test1',
+                'https://example.org/test2',
+            ],
+            $settings->getSpEntityIdsWithoutConsent()
+        );
+    }
+
+    public function testReadsAllConsentTypeOptions()
```
MKodde: Hypothetical the type could also be `null`. I'm not sure manage would ever output this as a type, but the json schema allows for the consent setting to be null. See: https://github.com/OpenConext/OpenConext-manage/blob/43e33c58714becbbc898f26ecf1bf607afc8a104/manage-server/src/main/resources/metadata_configuration/saml20_idp.schema.json#L121

theme/material/templates/modules/Authentication/View/Proxy/consent.html.twig
```diff
@@ -116,13 +118,21 @@
...
+                           {% else %}
                            value="{{ 'consent_buttons_ok'|trans({'%arg1%': spName}) }}"
+                           {% endif %}
                     />
                 </form>
 
                 <!-- NO -->
                 <a id="decline-terms" class="reject c-button c-button-link" href="#" data-slidein="reject">
-                    {{ 'consent_buttons_nok'|trans }}
+                    {%  if minimalConsent %}
```
MKodde: Why not create a new template for minimal consent? This way you will not need to constantly check for the `{% if not minimalConsent %}`. Which would become even more cumbersome when a new consent type arrives (if ever).

src/OpenConext/EngineBlock/Metadata/Entity/IdentityProvider.php
```diff
@@ -64,11 +66,11 @@ class IdentityProvider extends AbstractRole
...
 
     /**
-     * @var string[]
+     * @var ConsentSettings
      *
-     * @ORM\Column(name="sps_entity_ids_without_consent", type="array")
+     * @ORM\Column(name="consent_settings", type="json_array")
      */
-    public $spsEntityIdsWithoutConsent = array();
+    public $consentSettings;
```
MKodde: Can't we make this a private field? And always be sure the getter is used, ensuring the application always uses a ConsentSettings object?

src/OpenConext/EngineBlock/Metadata/ConsentSettings.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace OpenConext\EngineBlock\Metadata;
+
+/**
+ * Value object for IDP consent settings.
+ *
+ * @package OpenConext\EngineBlock\Metadata
+ */
+class ConsentSettings implements \JsonSerializable
```
MKodde: Nitpick, consider importing the JsonSerializable interface instead of yanking it from the \ namespace.

MKodde:
In the meantime, I've been able to get the right version of manage configured in my dev VM and I've been functional testing the changeset. No problems found so far.

tests/unit/OpenConext/EngineBlock/Metadata/ConsentSettingsTest.php
jorissteyn: I've added a new test case to assert we interpret null as default consent.

theme/material/templates/modules/Authentication/View/Proxy/consent.html.twig
jorissteyn: As discussed, abstracting the templates brings more complexity than hiding or changing two labels.

src/OpenConext/EngineBlock/Metadata/Entity/IdentityProvider.php
jorissteyn: Good idea, altough it's not in line with other similar properties (like logo). Fixed in new commit.

PR #546, opened 2018-06-20
Prevent all keypresses on the IDP list when modal is open

jorissteyn:
When the request access modal is open, keypresses in the input fields
should not interfer with the IDP list. Instead of only preventing
enter keypresses, prevent all keypresses from bubbling.

See https://www.pivotaltracker.com/story/show/155516160


theme/material/javascripts/modules/KeyboardListener.js
```diff
@@ -6,6 +6,12 @@ export class KeyboardListener {
     }
 
     handle(keyCode) {
+        if (this.requestAccessModalHelper.modalIsOpen()) {
+            // Don't interfer with the IDP selection list when
+            // the request-access modal dialog is open.
+            return;
+        }
+
         if (keyCode === ESCAPE) {
```
MKodde: This block becomes unreachable, should be in the `if (this.requestAccessModalHelper.modalIsOpen())` context.

PR #549, opened 2018-06-20
Prevent unexpected behaviour when using new manage on old EB

jorissteyn:
Before manage 2.0.18, consent should be disabled for all service
providers listed in 'disable_consent_connections'. Starting from the
new version of manage, the objects listed contain a type indicating
default, minimal or no consent.

In order to prevent unexpected behaviour when configuring default or
minimal consent in manage, we now check if the type is present and if
so, read the value to see if it's set to 'no_consent'.

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/JanusPushMetadataAssembler.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/JanusPushMetadataAssembler.php
@@ -411,13 +411,21 @@ class JanusPushMetadataAssembler
             return array();
         }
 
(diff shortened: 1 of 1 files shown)


jorissteyn:
@thijskh do you expect to use manage >=2.0.18 on EB 5.6? We could backport this.

MKodde:
For future reference, I've described this feature in the consent types Pivotal story. See: https://www.pivotaltracker.com/story/show/158145547/comments/190919225

thijskh:
@jorissteyn We are going to use the newer manage with EB 5.6 for a short while but can live with the consent gotcha for that time.

PR #551, opened 2018-06-20
Log the workflow state to the auth logs

MKodde:
The workflow state of the SP was chosen to serve as the log value. This
value should always be the same for the IdP. The state is stored in the
'workflow_state' key of the log context.

See: https://www.pivotaltracker.com/story/show/158283204

To test this in the dev vm. Tail  `/var/log/messages` grepping on `EBAUTH`. After authentication you should receive something like:

```
Jun 20 14:53:23 apps EBAUTH[5365]: {"channel":"authentication","level":"INFO","message":"login granted","context":{"login_stamp":"2018-06-20T14:53:23+02:00","user_id":"urn:collab:person:example.com:admin","sp_entity_id":"https:\/\/teams.vm.openconext.org\/shibboleth","idp_entity_id":"http:\/\/mock-idp","key_id":null,"proxied_sp_entity_ids":[],"workflow_state":"prodaccepted"},"extra":{"session_id":"5Y53SRgSPB6yuv1N9UWoQpLjjF2","request_id":"5b2a4e42cdb83"}}
```

### DIFF
--- a/src/OpenConext/EngineBlock/Logger/AuthenticationLogger.php
+++ b/src/OpenConext/EngineBlock/Logger/AuthenticationLogger.php
@@ -31,6 +31,7 @@ class AuthenticationLogger
      * @param Entity         $identityProvider
      * @param CollabPersonId $collabPersonId
      * @param array          $proxiedServiceProviders
+     * @param string         $workflowState
      * @param KeyId|null     $keyId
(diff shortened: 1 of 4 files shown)


jorissteyn:
Fine work!

PR #554, opened 2018-06-27
Use the first array value when single attribute value is shown on consent screen

MKodde:
Since the bugfix of the 'uneven' attribute value and type arrays was
introduced in the codebase. ARP filtering started introducing some
side effects on the consent screen.

This because attribute value and type arrays are filtered, taking extra
care to keep the keys intact. This might result in filtering away value
0, 2 and 3, leaving index 1.

The previous twig solution was not robust in that aspect and always
showed index 0, as that previously always was the first (and only) entry
in the case of a single value attribute.

Now the first filter is used to get the first array entry from the array
regardless of index, fixing the problem.

See: https://www.pivotaltracker.com/story/show/158184179/comments/191135246


MKodde:
Will release in 5.7.3 and merge it to master

PR #558, opened 2018-06-29
Remove eduGAIN section from front page

MKodde:
Just removes the section from the Twig template. Removal of the actual
metadata endpoints will be picked up at a later stage in the refactoring
of EB metadata generation.

See: https://www.pivotaltracker.com/story/show/158439992


jorissteyn:
@tvdijen I believe this change will save you one more patch in your implementation of EngineBlock :)

PR #559, opened 2018-07-02
RequestAccess: there's absolutely no need for system hostname in email subject

thijskh:


MKodde:
@jorissteyn, remember task 3 from https://www.pivotaltracker.com/story/show/149554887 ?

PR #563, opened 2018-07-04
Prevent twig extension XSS vulnerabilities

MKodde:
During the introduction of twig in engineblock. Some XSS vulnerabilities have been overlooked. This PR aims to fix all of these issues.

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/Feedback.php
+++ b/src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/Feedback.php
@@ -38,8 +38,8 @@ class Feedback extends Twig_Extension
     public function getFunctions()
     {
         return [
-            new TwigFunction('feedbackInfo', [$this, 'getFeedbackInfo'], ['is_safe' => ['html']]),
-            new TwigFunction('flushLog', [$this, 'flushLog'], ['is_safe' => ['html']]),
+            new TwigFunction('feedbackInfo', [$this, 'getFeedbackInfo']),
+            new TwigFunction('flushLog', [$this, 'flushLog']),
         ];
     }
 
--- a/theme/material/templates/layouts/scripts/language-switcher.html.twig
+++ b/theme/material/templates/layouts/scripts/language-switcher.html.twig
@@ -2,7 +2,7 @@
     {% if postData is not empty %}
     <form method="post" action="">
         {% for name, value in postData %}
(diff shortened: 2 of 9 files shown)


src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/Feedback.php
```diff
@@ -38,7 +38,7 @@ public function __construct(EngineBlock_ApplicationSingleton $application)
     public function getFunctions()
     {
         return [
-            new TwigFunction('feedbackInfo', [$this, 'getFeedbackInfo'], ['is_safe' => ['html']]),
+            new TwigFunction('feedbackInfo', [$this, 'getFeedbackInfo']),
             new TwigFunction('flushLog', [$this, 'flushLog'], ['is_safe' => ['html']]),
```
jorissteyn: If this function does not output anything, why mark it as "safe"?

theme/material/templates/layouts/scripts/language-switcher.html.twig
```diff
@@ -16,10 +16,10 @@
     {% else %}
         <ul class="comp-language">
             <li>
-                <a{% if currentLanguage == 'en' %} class="selected"{% endif %} href="{{ queryStringFor('en') }}">EN</a>
+                <a{% if currentLanguage == 'en' %} class="selected"{% endif %} href="{{ queryStringFor('en')|escape('url') }}">EN</a>
```
jorissteyn: This is not correct, see https://twig.symfony.com/doc/2.x/filters/escape.html:

    url: escapes a string for the URI or parameter contexts. This should not be used to escape an entire URI; only a subcomponent being inserted.

In HTML, hrefs shouldnt' be url encoded but must be escaped for HTML (attr) output. So escape('html_attr') is the only valid way of escaping here. Applies to other usages of `|escape('url')` as well, of course.

jorissteyn: Only the query parameters should be URL-encoded, and the twig function takes care of that. The attribute value should be escaped just like any other attribute value: using `html_attr`.

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/Feedback.php
MKodde: My rationale was: it's safe if it doesn't output anything. We're talking about the `flushLog` method here right?

PR #564, opened 2018-07-05
Actually show the collab name id on consent. Part II

MKodde:
There was a typo in the constant name checking for the nameid format.


thijskh:
Quick test confirms it does the trick!

PR #572, opened 2018-07-25
Remove attribute_aggregation_required metadata setting

pablothedude:
The attribute aggregator is now only queried when the metadata setting
coin:attribute_aggregation_required is set. This is simplified and
therefore an SP does not require this extra setting to be enabled
anymore.


jorissteyn:
We ought to drop this line too: https://github.com/OpenConext/OpenConext-engineblock/blob/master/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/JanusPushMetadataAssembler.php#L193

pablothedude:
Thanks. I've removed these line too.

pablothedude:
I've removed the paragraph as requested.

pablothedude:
Your changes make perfectly sense to me

PR #575, opened 2018-07-26
Fix wayf unconnected idp button hover

pablothedude:
The unconnected IdP rows should show the white button with text
"request access" to make sure it's clear what happens when you click it.

https://www.pivotaltracker.com/story/show/149554887


MKodde:
@arnoutt, as a WAYF expert, what do you thing of this change?

**Current**
screenshot from 2018-07-30 14-05-17

**New**
screenshot from 2018-07-30 14-03-03

tvdijen:
@MKodde I think you're missing the 'Previously chosen' header in the new situation?

arnoutt:
@MKodde is the 'Press enter to select' now always visible for IdPs with access? That is a bit confusing, since obviously when you press enter, only 1 IdP (the 1 that is currently selected) will be chosen :)

Same question for the 'Request access' button: is it always visible or only when hovered?

MKodde:
@tvdijen thanks for pointing that out! That issue was fixed in another PR and this work will probably be rebased on that once it is merged.

@arnoutt thanks for your feedback. The 'press enter to select' text was previously only visible on hover. But should of course only be set on the first/selected option. And yes, the request access buttons will now always be visible as advertised in the screenshot. When hovering these items they will revert to their current hover state.

This is our interpretation of the request by Thijs in the pivotal story mentioned in the description of this PR.

> The unconnected IdP rows should show the white button with text "request access" to make sure it's clear what happens when you click it.

arnoutt:
@MKodde Ok I understand, but would like to see that the 'Press enter to select' is only visible when hovered. The 'Request access' button should indeed always be visible. Thanks!

arnoutt:
@MKodde After some internal discussion, would it be possible to implement as follows:

- Currently, it is possible to highlight 2 IdPs: 1 with the keyboard (arrows), another with the mouse. Could this be made into 1 possible selection? So mouse overrides keyboard and vice versa.
- Remove the 'Press enter to select' button entirely (keep the 'Request access' button). (Keep the 'Request access' button.)

Thanks!

pablothedude:
The requested adjustements by @arnoutt are implemented. @MKodde could you verify if all requirements are met.

MKodde:
The IdP selection list works more intuitive now! 

One thing that might not yet be as intended is when mouse hovering over the 'Identity providers without access' entries, they are not lit up. Which seems logical to me, but @arnoutt might think otherwise? At the time of taking the screenshot, my cursor was on the location where I drew the red arrow. No additional hover state is applied to the NWO entry.  What do you think Arnout?

screenshot-engine vm openconext org-2018-08-02-09-08-31

arnoutt:
@MKodde the 'IdPs without access' list should behave the same as those with access, which means the selection should also light up when hovered. Also, it should be possible to select them with the arrow keys and enter.

Rest looks nice, thanks!

MKodde:
And one little thing in IE11: When returning to the WAYF after cancelling authentication for an IdP, all IdP's are selected in the overview.

screenshot from 2018-08-02 09-22-00

pablothedude:
@MKodde I've fixed the IE11 behaviour as requested
@arnoutt I've created a ticket in the backlog with your request becasue it's kinda out-of-scope so your request could be implemented later on:
https://www.pivotaltracker.com/story/show/159595638

theme/material/javascripts/modules/MouseListener.js
```diff
@@ -0,0 +1,26 @@
...
+            this.searchBarElement.blur();
+            return;
+        }
+
+        var index = target.getAttribute('data-idp-list-index');
+        if (index !== null) {
+            this.idpPicker.focusOnIdpByIndex(index);
+        }
+    }
+}
```
MKodde: Nittiest of nitpickings: Maybe on your next rebase also add a white space to the bottom of the file.

PR #576, opened 2018-07-26
Show SP name on "No connection between institution and service" page

pablothedude:
When a user wants to log into a service that the IdP does not allow,
and the user is presented with an error page "No connection between
institution and service", the entityID of the SP is listed. Now also
the SP name is shown in this message because that's more meaningful.

Also see the [Pivotal issue](https://www.pivotaltracker.com/story/show/149064131)

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
@@ -7,7 +7,6 @@ use OpenConext\EngineBlockBundle\Pdp\PolicyDecision;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 use Twig_Environment;
 
--- a/README.md
+++ b/README.md
@@ -103,15 +103,14 @@ the `composer` part above with `bin/composer.phar`. This version is regularly up
 being outdated.
 
 
-### Install database schema
+### Install database schema updates
 
-To install the initial database, just call the 'migrate' script in *bin/*, followed by migration tool introduced in 5.x,
-like so:
+To install possible database updates, call doctrine migrations by using the following console command:
 
-    (cd bin && ./migrate && cd .. && app/console doctrine:migrations:migrate --env=prod)
+    app/console doctrine:migrations:migrate --env=prod
 
 _**Note**:
-EngineBlock requires database settings, without it the install script will not function. Furthermore, this assumes that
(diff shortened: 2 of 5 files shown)


src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
```diff
@@ -309,6 +317,11 @@ public function unknownPreselectedIdpAction(Request $request)
     {
         // Add feedback info from url
         $customFeedbackInfo['Idp Hash'] = $request->get('idp-hash');
+
+        $serviceProvider = $this->getServiceProviderById();
+        $name = $serviceProvider->getDisplayName($request->getLocale());
+        $customFeedbackInfo['Service Provider Name'] = $name;
```
MKodde: Using 'Service Provider Name' as the title here results in a missing translation error (see screenshot)

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
```diff
@@ -340,4 +353,16 @@ private function setFeedbackInformationOnSession(array $customFeedbackInfo)
         }
         $_SESSION['feedbackInfo'] = array_merge($customFeedbackInfo, $_SESSION['feedbackInfo']);
     }
+
+    /**
+     * @SuppressWarnings(PHPMD.Superglobals) This is required to mimic the existing functionality
```
MKodde: As we discussed IRL, not the most elegant solution to get the current SP but I think this is the easiest way to get it at this point.

Alternatively, you could fish the required information from the `SAMLRequest` index in the SESSION superglobal.

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
```diff
@@ -31,19 +32,26 @@ class FeedbackController
...
     ) {
         $this->translator = $translator;
         $this->twig = $twig;
         $this->logger = $logger;
+        $this->engineBlockApplicationSingleton = $engineBlockApplicationSingleton;
 
         // we have to start the old session in order to be able to retrieve the feedback info
         $server = new EngineBlock_Corto_ProxyServer($twig);
         $server->startSession();
+
```
MKodde: Empty line on the last line of the constructor may be removed

#nitpick

pablothedude:
I've added the missing translations and fixed the empty line. I've looked into the option to get the name of the service provider form the `SAMLRequest` but I wasn;t able to do so.

Also I removed the logic from the FeedbackController and moved it to the ApplicationSingleton. As a result the name of the service provider is always set if an error report is being generated. I think that's not an issue and it's always helpful on an error report when a service provider id is set but makes sure the name is updated after a new error report was generated. I will report this behavior on a ticket so @thijskh is ware of this change.

PR #577, opened 2018-07-27
Create a DateTime value object that allows to control time for tests

pablothedude:
Also removed a lot death code only used for testing?


MKodde:
As far as I could see, the dead code you removed was not yet in use. And was simply copy/pasted from Stepup-Middleware. Removing it seems the right choice here.

PR #578, opened 2018-07-31
Prevent collapse issues on consent screen for IE11

MKodde:
The application.min.js was generated afresh causing problems on the
consent screen before. This was reported by the test team in ticket:
159198595

While at it, a new revision of the bootstrap.php.cache is generated and
included in this commit.

To test this issue in Browser stack from a 5.7+ development environment, 
please come to me for instructions or a live demo on my system.

See https://www.pivotaltracker.com/story/show/159198595


pablothedude:
Seems valid to me

PR #581, opened 2018-08-01
Rename Janus references

MKodde:
Instead of using Janus or Manage as an indication for the push API
feature that EngineBlock features, a more generic 'metadata push' nomer
is used.

Also see Pivotal: https://www.pivotaltracker.com/story/show/157912446

### DIFF
--- a/ci/travis/files/engineblock.ini
+++ b/ci/travis/files/engineblock.ini
@@ -17,8 +17,8 @@ engineApi.user = serviceregistry
 engineApi.password = test
 
 ;; New version that allows for multiple different API users
-engineApi.users.janus.username = serviceregistry
-engineApi.users.janus.password = test
+engineApi.users.metadataPush.username = serviceregistry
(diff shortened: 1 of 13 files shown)


ci/travis/files/engineblock.ini
```diff
@@ -17,8 +17,8 @@ engineApi.user = serviceregistry
 engineApi.password = test
 
 ;; New version that allows for multiple different API users
-engineApi.users.janus.username = serviceregistry
-engineApi.users.janus.password = test
+engineApi.users.metadataPush.username = serviceregistry
```
pablothedude: The standard used for non root configurations seems to be snakecase,  but now lower-camel-case is used? Should this be changed for clarity?

MKodde: I could not realy find a standard, so I've chosen the camelCase variant. If you're okay with this I'd like to keep it this way.

PR #583, opened 2018-08-02
Add CodeStyle fixer to the project and remove explicit PHP 5.3 compatibility

pablothedude:
Add extra codestyle commands to ant
https://www.pivotaltracker.com/story/show/115227481

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -80,6 +80,16 @@
         </exec>
     </target>
 
+    <target name="phpcbf"
+            description="Fix coding standard violations using PHP_CodeSniffer">
+        <exec dir="${basedir}" executable="vendor/bin/phpcbf" failonerror="true">
+            <arg value="--standard=ci/travis/phpcs.xml" />
+            <arg value="--extensions=php" />
+            <arg value="--no-patch" />
+            <arg path="src" />
+        </exec>
+    </target>
+
     <target name="security-checker" description="Check for vulnerable dependencies with SensioLabs Security Checker">
         <exec dir="${basedir}" executable="vendor/bin/security-checker" failonerror="true">
             <arg value="security:check" />
--- a/build.xml
+++ b/build.xml
@@ -90,6 +90,26 @@
         </exec>
     </target>
 
+    <target name="phpcompat"
+            description="Check code compatibilty using PHP_CodeSniffer">
+        <exec dir="${basedir}" executable="vendor/bin/phpcs" failonerror="true">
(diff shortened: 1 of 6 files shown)


build.xml
```diff
@@ -90,6 +90,26 @@
...
     </target>
 
+    <target name="phpcompat"
+            description="Check code compatibilty using PHP_CodeSniffer">
+        <exec dir="${basedir}" executable="vendor/bin/phpcs" failonerror="true">
+            <arg value="--standard=ci/travis/phpcompat.xml" />
+            <arg value="--extensions=php" />
+            <arg value="--runtime-set" />
+            <arg value="testVersion" />
+            <arg value="7.0-" />
```
MKodde: Officially we support only PHP 5.6. Although It won't be long before upgrading to 7.x. So let's leave this value at 7.0 for now?

ci/travis/phpcompat.xml
```diff
@@ -0,0 +1,20 @@
...
+        This is used by the Ibuildings QA tools to wrap the coding standard of your choice.
+        By default it is less stringent about long lines than other coding standards
+    </description>
+
+    <exclude-pattern>src/*/*Bundle/Resources</exclude-pattern>
+    <exclude-pattern>src/*/*Bundle/Tests</exclude-pattern>
+    <exclude-pattern>src/*/Bundle/*Bundle/Resources</exclude-pattern>
+    <exclude-pattern>src/*/Bundle/*Bundle/Tests</exclude-pattern>
+    <exclude-pattern>*/Tests/*</exclude-pattern>
+
```
MKodde: Have you considered adding the `tests` folder to the excludes list? The `*/Tests/*` entry could be replaced with this in my opinion.

build.xml
pablothedude: I've done this on purpose so we could see on forehand what is going to break. So all new code is 7.0 compatible. Maybe it's even better to force >7.2. What's your view on this?

ci/travis/phpcompat.xml
pablothedude: This is a copy/paste action from the defaul QA-tools `/ci/travis/phpcs.xml`. This is done in order support default bundle setups. When the command is called, as last parameter a folder is given and that folder is at the moment `src` or `library` so excluding `test` wouldn't be necessary.

build.xml
MKodde: I think checking for 7.2 compatibility is a good proposal. Amending this to the `build.xml`

PR #589, opened 2018-08-15
Make database health query configurable

MKodde:
The monitor_database_health_check_query ini config is used to set the
Symfony yml param, which is included in the monitor config that was
added to the services.yml file.

By default a sensible query is set in the application.ini, this can be
overridden if so desired by the end user.

Todo:
 * [x] Update composer.json / composer.lock once the monitor-bundle is approved & released

### DIFF
--- a/src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
+++ b/src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
@@ -1,0 +1,78 @@
+<?php
+
+/**
+ * Copyright 2018 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlockBundle\HealthCheck;
+
+use Doctrine\ORM\EntityManager;
+use Exception;
+use OpenConext\EngineBlock\Assert\Assertion;
+use OpenConext\MonitorBundle\HealthCheck\HealthCheckInterface;
+use OpenConext\MonitorBundle\HealthCheck\HealthReportInterface;
+use OpenConext\MonitorBundle\Value\HealthReport;
+
+/**
+ * Test if there is a working database connection.
+ */
+class DoctrineConnectionHealthCheck implements HealthCheckInterface
+{
+    /**
+     * @var EntityManager|null
+     */
+    private $entityManager;
+
+    /**
+     * @var string
+     */
+    private $query;
+
+    public function __construct($query)
+    {
+        Assertion::nonEmptyString($query, 'health check query');
+        $this->query = $query;
+    }
+
+    /**
+     * Run the doctrine connection health check.
+     *
+     * @param HealthReportInterface $report
+     * @return HealthReportInterface
+     */
+    public function check(HealthReportInterface $report)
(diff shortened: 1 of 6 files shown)


src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
```diff
@@ -0,0 +1,78 @@
...
+    private $query;
+
+    public function __construct($query)
+    {
+        Assertion::nonEmptyString($query, 'health check query');
+        $this->query = $query;
+    }
+
+    /**
+     * @return HealthReportInterface
```
pablothedude: Missing parameter phpdoc or inherit

src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function check(HealthReportInterface $report)
+    {
+        // Was the entityManager injected? When it is not the project does not use Doctrine ORM
+        if (!is_null($this->entityManager)) {
+            try {
+                $connection = $this->entityManager->getConnection();
+                $connection->exec($this->query);
+                $connection->close();
+                return HealthReport::buildStatusDown('Unable to execute a query on the database.');
```
pablothedude: If the query succeeds it still returns an invalid status

src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
```diff
@@ -0,0 +1,78 @@
...
+     * @return HealthReportInterface
+     */
+    public function check(HealthReportInterface $report)
+    {
+        // Was the entityManager injected? When it is not the project does not use Doctrine ORM
+        if (!is_null($this->entityManager)) {
+            try {
+                $connection = $this->entityManager->getConnection();
+                $connection->exec($this->query);
+                $connection->close();
```
pablothedude: Is there a reason why the close is here and it's not handled by Symfony itself?

MKodde: Doctrine fails to close the connection resulting in many sleeping mysql connections, resulting in nasty log messages.

The problem is described in this thread: https://stackoverflow.com/questions/7134469/doctrine-2-close-connection

src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
MKodde: :fearful:

src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
MKodde: Fixed

src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php
tvdijen: Hah, Ik think this is exactly what I'm experiencing:
https://github.com/OpenConext/Monitor-bundle/issues/1

I am seeing similar issues with health checks from StepUp (middleware/gateway).. Are those on your radar too?

MKodde: The new Monitory bundle release could now be easily installed in the other applications. And I'm going to!

MKodde: Just installed the upgrade to GW, MW, and some other relevant OpenConext applications. They have been pushed to develop, so in the next release of these applications, the update should be shipped.

tvdijen: I just did a `composer update` on all OC apps to pull in monitor-bundle 1.0.3, but so far I'm still seeing these errors.. Actually I can just trigger them by going to /health on my engineblock

PR #590, opened 2018-08-20
Symfony 3.4 upgrade

MKodde:
This is a proposal PR to upgrade Symfony 2.8 to 3.4.

In the process, quite some deprecation warnings had to be addressed, mainly autoloader related. In the future, we could take a critical look at utilizing the autowire option that is now at our disposal.

See #159790180


pablothedude:
I get an error while testing this upgrade. This is because of the `app_dev.php` which doesn't get updated after this change while it should. Is it possible to add a `dev_php.dist` to the repo and add a PR to engineblock-deploy where this file gets copied in the development environment so we wouldn't have this issue anymore?

pablothedude:
@thijskh is my proposal acceptable for you? The main concern of @MKodde is that the `app_dev.php.dist` was in the root of the production project. I don't think hat should be an issue but I wanted to verify that with you to be sure.

thijskh:
I don't think that's a problem per se. Maybe you could skip/remove that file when building a release tarball?

MKodde:
Sounds like a good suggestion to me!

MKodde:
Holding back the merge of this PR until release 5.8.x is tagged. This change is targetted for 5.9.0.

PR #593, opened 2018-09-07
Rolling update for release 5.8

pablothedude:
This PR is to allow rolling updates and is intended to release 5.7.5 to 5.8
https://www.pivotaltracker.com/story/show/159983435

There are two features which could possibly break backwards compatibility:
- Allow enforcing requesterid on trusted proxy
- Remove attribute_aggregation_required metadata setting

The first should be backwards compatible because only a column is added so there shouldn't be any problems at all. The second removes an old metadata column and therefore needs some extra attention to be backwards compatible.

Thats why I've created three separate commits:
- The first commit removes the migration where the `attribute_aggregation_required` is dropped
- The second commit keeps the old code which changed the `attribute_aggregation_required` attribute so this attribute is still modified on data push. 
- The third commit removes the `attribute_aggregation_required` because this column isn't needed after a successful rolling update.

:warning: This PR and it's branch should never be merged back to master! We will cherry pick the commits manually. This is meant to be a place for discussion and review.


MKodde:
Great work so far! But I think you missed the most challenging part in your proposal. And that's the `consent_settings` column that is updated in the `Version20180724175453`

Not only the structure of the database column has changed, but also the contents of the column changed quite drastically. For more details see #544.

Do you propose we use this branch to cherry pick the commits from? In that case, we might need to put some more effort in the individual commits (messages could be more descriptive) I've got some ideas for that. 

Otherwise, this might be a very nice way to prepare a release!

UPGRADING.md
```diff
@@ -3,8 +3,8 @@
 ## 5.7 -> 5.8
 
 ### Stored metadata incompatibility
-Metadata pushed to EngineBlock in earlier versions (EB<5.8) is not compatible with this version. A metadata push is
-required after upgrading to EB 5.8.
+In order to upgrade from 5.7 to 5.8 you need to go to 5.8.3 to prevent backwards compatibility
+breaking changes. From version 5.8.3 rolling updates are supported.
```
MKodde: Great addition! But why remove the notion of the need for a metadata push? This still is mandatory right?

database/DoctrineMigrations/Version20180618134736.php
```diff
@@ -6,6 +6,8 @@
 use Doctrine\DBAL\Schema\Schema;
 
 /**
+ * From now on rolling updates are supported.
```
MKodde: Possibly elaborate here that the `sps_entity_ids_without_consent` column is kept for rollback purposes and will be cleaned up in the next release.

database/DoctrineMigrations/Version20180618134736.php
```diff
@@ -18,7 +20,7 @@ public function up(Schema $schema)
         // this up() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
 
-        $this->addSql('ALTER TABLE sso_provider_roles_eb5 CHANGE sps_entity_ids_without_consent consent_settings LONGTEXT DEFAULT NULL COMMENT \'(DC2Type:array)\'');
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 ADD sps_entity_ids_without_consent LONGTEXT DEFAULT NULL COMMENT \'(DC2Type:array)\'');
```
MKodde: When migrating from a 5.7 version, wouldn't this `sps_entity_ids_without_consent` column already be present? Or am I mixing things up here?

I expected the addition of the `consent_settings` column here.

database/DoctrineMigrations/Version20180618134736.php
```diff
@@ -29,6 +31,6 @@ public function down(Schema $schema)
         // this down() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
 
-        $this->addSql('ALTER TABLE sso_provider_roles_eb5 CHANGE consent_settings sps_entity_ids_without_consent LONGTEXT DEFAULT NULL COLLATE utf8_unicode_ci COMMENT \'(DC2Type:array)\'');
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 DROP sps_entity_ids_without_consent');
```
MKodde: Again I think we need to keep this column when rolling back to 5.7.x

UPGRADING.md
pablothedude: That's true, I thought that pushing the meta data again should be very obvious. But it is certainly better not to drop that line.

database/DoctrineMigrations/Version20180618134736.php
pablothedude: That's a good suggestion but maybe it's even better to have a specific place to document these kind of scheme changes presumably in the docs folder. Maybe we could discuss IRL what the best option should be?

database/DoctrineMigrations/Version20180724175453.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace OpenConext\EngineBlock\Doctrine\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180724175453 extends AbstractMigration
```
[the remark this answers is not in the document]
pablothedude: Personally I would prefer to do this per feature, I think that should be more maintainable in the future and you could merge everything a lot easier. I did change the order to change the order of  the migrations to be more sequential.

database/DoctrineMigrations/Version20180618134736.php
pablothedude: Changed

database/DoctrineMigrations/Version20180618134736.php
MKodde: Yup: `UPGRADING.md` is a perfect place for this in my opinion. We will discuss this IRL later on!

database/DoctrineMigrations/Version20180724175453.php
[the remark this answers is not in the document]
MKodde: :+1: makes perfect sense.

thijskh:
So for my understanding, the first commit would be part of 5.8
and the second commit would be part of >=5.9

src/OpenConext/EngineBlock/Metadata/Entity/IdentityProvider.php
```diff
@@ -65,6 +65,13 @@ class IdentityProvider extends AbstractRole
      */
     public $schacHomeOrganization = null;
 
+    /**
+     * @var string[]
+     *
+     * @ORM\Column(name="sps_entity_ids_without_consent", type="array")
+     */
+    public $spsEntityIdsWithoutConsent = array();
```
MKodde: I've been testing the change with a functionally. And we need a to revert more code in the IdentityProvider entity.

Consider the change introduced in d56d64f. You will need to add the `spsEntityIdsWithoutConsent` parameter back into the constructor and make sure it will be saved correctly to the `sso_roles` table! Now I get empty arrays at the places where should be a listing of the SP's that should not provide consent.

MKodde: Additionally, the `attribute_aggregation_required` column is also not updated! Might this be the same problem?!

pablothedude:
Yes @thijskh, if we're done testing we will cherrypick the desired commits. You only should make sure that the release of versions >5.8.1 <= 5.8.2 should not be used on production. Because there are migrations in those versions which don't support rolling updates.

PR #594, opened 2018-09-11
 Prevent 'Aborted connection' warnings from the DoctrineConnectionHealthCheck

pablothedude:
The shutdown function in the monitor bundle isn't always called
after a request. So we need to manually close the connection to
prevent the connections from staying open.


pablothedude:
We've tested the issue on the test environment and could reproduce the problem there. The problem was that the wrong method was used and therefore the connections were not properly closed. This fix should now address this  issue.

MKodde:
See: https://www.pivotaltracker.com/story/show/158906723

MKodde:
@pablothedude, would you be so kind to install this change in:
* Stepup-*
* SP Dashboard
* User lifcecycle
* Engine (even tho this change was already fixed in the overloaded Checker)

tvdijen:
Well that exec/query was a needle in a haystack!  Great job!! I can confirm that this fixes the issue...

PR #597, opened 2018-10-03
Improve the error message from MultilingualValue

MKodde:
In some cases this error message is raised in the logs, and the generic
error message does not help to identify the actual problem.

Before: 'Value "<NULL>" expected to be string, type NULL given.'
After: 'The 'value' of a MultilingualValue should be a string. "%s" given'

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/MultilingualValue.php
+++ b/src/OpenConext/EngineBlock/Metadata/MultilingualValue.php
@@ -18,8 +18,20 @@ class MultilingualValue
      */
     public function __construct($value, $language)
     {
-        Assertion::string($value);
-        Assertion::string($language);
+        Assertion::string(
+            $value,
+            sprintf(
+                'The \'value\' of a MultilingualValue should be a string. "%s" given',
+                var_export($value, true)
+            )
+        );
+        Assertion::string(
+            $language,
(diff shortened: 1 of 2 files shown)


MKodde:
@quartje, this is the change I mentioned in the slack channel. The fuzzy error message has been made more verbose. If any other changes are required to fix the actual problem you encoutered, please let me know!

src/OpenConext/EngineBlock/Metadata/MultilingualValue.php
```diff
@@ -18,8 +18,20 @@ class MultilingualValue
      */
     public function __construct($value, $language)
     {
-        Assertion::string($value);
-        Assertion::string($language);
+        Assertion::string(
+            $value,
+            sprintf(
+                'The \'value\' of a MultilingualValue should be a string. "%s" given',
+                var_export($value, true)
```
pablothedude: Is it safe to to show the invalid value itself? Probably only the message itself would be sufficient?

MKodde: This message ends up in the log, I thought it beneficial for debugging. But maybe you are right. The mention of `MultilingualValue` will help enough.

PR #598, opened 2018-10-11
Filter ACS location URI scheme to prevent XSS vulnerability

MKodde:
There was a possible XSS vulnerability where EB would take the ACS location from a signed AuthnRequest and use that location to send the SAMLResponse. The URI scheme was prone to XSS, as the JS URI scheme was allowed.

The proposed change makes the allowed URI schemes configurable. By default only allowing the `http` and `https` schemes.

The ACS locations that are received from Manage are also validated.

This PR is first targeted at the 5.8 branch.

:warning: Disclaimer, the build fails on the security checker. This is a well-considered decision.

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -235,6 +235,11 @@ ui.return_to_sp_link.active = false
 ; Signature methods explicitly forbidden by EngineBlock, comma-separated.
 ; forbiddenSignatureMethods = "http://www.w3.org/2000/09/xmldsig#rsa-sha1"
 
+; Configure which ACS location URI schemes are allowed for incomming AuthnRequests. The settings is configured with
+; with the 'http' and 'https' schems by default. Which are considered safe to use.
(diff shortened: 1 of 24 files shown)


thijskh:
It is not related to trusted proxy mode btw

application/configs/application.ini
```diff
@@ -237,8 +237,8 @@ ui.return_to_sp_link.active = false
 
 ; Configure which ACS location URI schemes are allowed for incomming AuthnRequests. The settings is configured with
 ; with the 'http' and 'https' schems by default. Which are considered safe to use.
```
tvdijen: schems = schemas

MKodde: Thanks, fixed this!

thijskh:
Yes, needs `redirect.sign` to be on for the SP. But obviously we don't mean with that settings to give SP's control over engineblock.

PR #604, opened 2018-10-22
Prevent crash when OriginalIssuer is null

MKodde:
When OriginalIssuer can not be read from the response, EB used to crash.
This is prevented by reporting on this issue and skipping the additional
logging that might have been requested.

This is a workaround to an issue that is described in greater detail in:
https://www.pivotaltracker.com/story/show/161346289

The actual solution to this problem is yet to be found.
Thanks @thijskh for collaborating on this fix!


pablothedude:
Looks good, only travis isn;t working ATM to see the build pass:
https://blog.github.com/2018-10-21-october21-incident-report/

PR #606, opened 2018-10-24
Replace un needed polyfills

BackEndTea:
This project runs php 5.6, so polyfills providing php 5.4 to 5.6 functionality dont add anything. Instead, they only serve to clog up the autoloader as they are autoloaded at the start of any process.

Mb string can be replaced as the extension is installed, see:
* https://github.com/OpenConext/OpenConext-deploy/blob/17d1d5c3782f1d8d0f20370f210d47634bf394d8/roles/php56fpm/tasks/main.yml#L33

See also:
* https://github.com/symfony/polyfill/pull/153


tvdijen:
I think `ext-mbstring` should stay.. It's a requirement, so it should be listed as such

PR #607, opened 2018-10-24
Add an editorconfig

BackEndTea:
The editorconfigs reflects current styling


MKodde:
It took a while, but we are merging this. Thanks @BackEndTea

PR #608, opened 2018-11-14
Fix validity checking of shibmd:scopes in regexp format.

thijskh:
In Metadata these are stored without the delimiters. We need to add the
delimiters around these values when testing for validity or using the
regexp. saml-value-objects already does this correctly when applying
the scope values to attributes. But this code was not triggered because
of the validity test in EB that considers each regexp scope invalid.


thijskh:
The build failure is only the security checker for a PHP 5.5 only issue in code we do not use.

MKodde:
Why the [VerifyShibMdScopingAllowsEduPersonPrincipalNameTest](https://github.com/OpenConext/OpenConext-engineblock/blob/f897c224ee5ff52a9c1413bcaf5a454038ff4101/tests/library/EngineBlock/Test/Corto/Filter/Command/VerifyShibMdScopingAllowsEduPersonPrincipalNameTest.php) and [VerifyShibMdScopingAllowsSchacHomeOrganisationTest](https://github.com/OpenConext/OpenConext-engineblock/blob/f897c224ee5ff52a9c1413bcaf5a454038ff4101/tests/library/EngineBlock/Test/Corto/Filter/Command/VerifyShibMdScopingAllowsSchacHomeOrganisationTest.php) did not fail is not 100% clear to me. The input for the regex tests do not add the delimiters but passed previous tests. I'm spinning up my box to further investigate.

thijskh:
@MKodde Thanks for the analysis. Indeed it logs the warning but does not do what it claims ("Ignoring..."). I've opted to remove it entirely then, since the regexp validity check is already done when constructing the value object and an exception will be thrown if it's invalid.

PR #609, opened 2018-11-15
Remove unused config

tvdijen:


tvdijen:
It is not unused, by bad

PR #611, opened 2018-12-17
ProtocolBinding is not a mandatory field. Assume HTTP-POST if it's lacking

thijskh:
It is not right to disallow authnrequests without the protocolbinding,
since it's optional in the standard. EB only supports HTTP-POST, so that
seems a sane default. Further strengthened by saml2int's statement that
"The ProtocolBinding attribute, if present, MUST be set to
urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST."

### DIFF
--- a/library/EngineBlock/Corto/ProxyServer.php
+++ b/library/EngineBlock/Corto/ProxyServer.php
@@ -610,28 +610,35 @@ class EngineBlock_Corto_ProxyServer
         /** @var AuthnRequest $request */
 
         // Ignore requests for bindings we don't support for responses.
-        if ($request->getProtocolBinding() && ($request->getProtocolBinding() !== Constants::BINDING_HTTP_POST)) {
+        $protocolBinding = $request->getProtocolBinding();
+        // ProtocolBinding is not mandatory, seems safe to assume HTTP-POST since it's
+        // the only thing EB supports anyway.
+        if (empty($protocolBinding)) {
+            $protocolBinding = Constants::BINDING_HTTP_POST;
+        }
+
+        if ($protocolBinding !== Constants::BINDING_HTTP_POST) {
             $this->_server->getLogger()->notice(
-                "ProtocolBinding '{$request->getProtocolBinding()}' requested is not supported, ignoring..."
+                "ProtocolBinding '{$protocolBinding}' requested is not supported, ignoring..."
             );
             return false;
         }
 
         // Custom ACS Location & ProtocolBinding goes first
-        if ($request->getAssertionConsumerServiceURL() && $request->getProtocolBinding()) {
+        if ($request->getAssertionConsumerServiceURL()) {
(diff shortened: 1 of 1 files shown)


library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -610,28 +610,35 @@ public function getCustomAssertionConsumer(
...
+        $protocolBinding = $request->getProtocolBinding();
+        // ProtocolBinding is not mandatory, seems safe to assume HTTP-POST since it's
+        // the only thing EB supports anyway.
+        if (empty($protocolBinding)) {
+            $protocolBinding = Constants::BINDING_HTTP_POST;
+        }
+
+        if ($protocolBinding !== Constants::BINDING_HTTP_POST) {
             $this->_server->getLogger()->notice(
                 "ProtocolBinding '{$request->getProtocolBinding()}' requested is not supported, ignoring..."
```
MKodde: I would use `$protocolBinding` here for good measure, if in the future you would ever get the protocol binding from another resource you would then only have to change it in 1 place.

library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -657,18 +664,6 @@ public function getCustomAssertionConsumer(
             }
             return false;
         }
-        else {
-            if ($request->getAssertionConsumerServiceURL() || $request->getProtocolBinding()) {
-                // Note that an SP is not actually required to supply both a URL and a Binding.
-                // But what should we do if we don't have both? Pick out a random counterpart from the metadata?
-                // Seems a little hard to predict for the SP, so we go with the default endpoint.
-                $this->_server->getLogger()->notice(
```
MKodde: Can you explain why you removed this log entry? Your modifications do not exclude the possibility for a situation where there would be no ACS url, but there is a protocol binding.

library/EngineBlock/Corto/ProxyServer.php
thijskh: Agreed, just forgot that one...

library/EngineBlock/Corto/ProxyServer.php
thijskh: I could indeed change it to `if $protocolBinding) then log notice "protocol binding specified but no acs"`. Seems exceptionally rare since it semantically does not mean anything if you do that.

MKodde: You be the judge, it might be interesting to have in the logs if this ever occurs. The fact that it was logged might be an indication of this.

thijskh: Because we already set protocolBinding to a default value if it was empty, it will always be set at this point.

The notice existed precisely because the behaviour was different before. I'm fine with leaving it out.

PR #615, opened 2019-01-16
Add Portuguese language support

domgon:
add the Portuguese language messages and add the two tag for "done" and "edit" for previews selected IDP

### DIFF
--- a/languages/messages.nl.php
+++ b/languages/messages.nl.php
@@ -53,6 +53,8 @@ return $overrides + [
     // WAYF
     'search'                    => 'Zoek een %organisationNoun%...',
     'our_suggestion'            => 'Onze suggestie:',
+    'edit'                      => 'bewerk',
+    'done'                      => 'gedaan',
     'idps_with_access'          => '%organisationNounPlural% met toegang',
     'idps_without_access'       => '%organisationNounPlural% zonder toegang',
     'log_in_to'                 => 'Selecteer een %organisationNoun% en login bij',
--- a/languages/messages.pt.php
+++ b/languages/messages.pt.php
@@ -1,0 +1,279 @@
+<?php
+
+$overrides = [];
+$overridesFile = __DIR__ . '/overrides.pt.php';
+if (file_exists($overridesFile)) {
+    $overrides = require $overridesFile;
+}
+
+return $overrides + [
+    // Values used in placeholders for other translations
+    // %suiteName%: OpenConext (default), SURFconext, ACMEconext
+    'suite_name' => '{{ instance_name }}',
+
+    // Example translation message:
+    //     'Find an %organisationNoun%'
+    //
+    // Becomes:
+    //     'Find an organisation' (default)
+    // or: 'Find an institution' (when overridden)
+    'organisation_noun' => 'organização',
+    'organisation_noun_plural' => 'organizações',
+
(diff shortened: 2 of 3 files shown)


MKodde:
Thanks for your translation efforts @domgon! I'm going to review your proposal in a functional manner as my Portugese is rusty at best :smile:

languages/messages.pt.php
```diff
@@ -0,0 +1,279 @@
+<?php
+
+$overrides = [];
+$overridesFile = __DIR__ . '/overrides.pt.php';
```
MKodde: You might already be formiliar with the possibility of adding an 'override' translation file. For more information see [this section](https://github.com/OpenConext/OpenConext-engineblock/blob/master/UPGRADING.md#changes-to-translation-files) of the upgrade notes.

languages/messages.nl.php
```diff
@@ -53,6 +53,8 @@
     // WAYF
     'search'                    => 'Zoek een %organisationNoun%...',
     'our_suggestion'            => 'Onze suggestie:',
+    'edit'                      => 'bewerk',
```
MKodde: Good catch! these translations were left untranslated as for both the Dutch and English languages we simply used `edit` and `done`. 

Your translation to the Dutch language is admirable but not linguistically sound :sweat_smile: in my opinion.

For now, I suggest changing the Dutch translations back to 'edit' and 'done'. Unless @thijskh or @arnoutt have a better suggestion? For context: these texts are displayed when the user is modifying the suggested IdP's on the WAYF.

languages/messages.pt.php
```diff
@@ -0,0 +1,279 @@
...
+if (file_exists($overridesFile)) {
+    $overrides = require $overridesFile;
+}
+
+return $overrides + [
+    // Values used in placeholders for other translations
+    // %suiteName%: OpenConext (default), SURFconext, ACMEconext
+    'suite_name' => '{{ instance_name }}',
```
MKodde: FYI the instance name is one of the translations that is usually set in the overrides file.

languages/messages.pt.php
```diff
@@ -0,0 +1,279 @@
...
+    'cookie_removal_header'     => 'Remover cookies',
+    'cookie_remove_button'      => 'Remover',
+    'cookie_remove_all_button'  => 'Remover todos',
+    'cookie_removal_description' => '<p>Em baixo poderá encontrar uma visão geral dos seus cookies e a possibilidade de os remover individualmente ou todos de uma vez.</p>',
+    'cookie_removal_confirm'     => 'O seu cookie foi removido.',
+    'cookies_removal_confirm'    => 'Os seus cookies foram removidos.',
+
+    // Footer
+    'service_by'            => 'Este serviço está ligado através da',
+    'serviceprovider_link'  => '<a href="https://www.fccn.pt/seguranca/rcts-aai/" target="_blank">%suiteName%</a>',
```
MKodde: We aim to prevent vendor specific entries in the translation files. This makes is easier for other users to adopt the product. Consider keeping this link more generic and overriding it to the correct value in the overrides.pt.php file.

languages/messages.nl.php
thijskh: I would prefer "bewerken" and "klaar" personally.

PR #617, opened 2019-01-24
Make tests Behat 3 compatible on Travis PHP 5.6

MKodde:
This PR tries to isolate (and fix) the issue with the failing functional tests on PHP 5.6 builds. Behat 3 on Travis Precice in combination with PHP 5.6 causes weird filesystem issues. Where files written to disk (in the project folder) are not always readable.

The following commits address this issue by leveraging Flysystem. Flysystem can easily be set with different locking strategies.

In addition to using Flysystem, an additional manual call to gc_collect_cycles was required to ensure clearing the fixture files does not fail.

One other task that was included in this PR was the move from Travis (precise) to Travis Xenial. Hoping the filesystem issue would not be present on that platform. This was not the case. But moving away from the already decommissioned Precise was a task that was overdue.

https://www.pivotaltracker.com/story/show/161282428

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,10 @@ branches:
     - release/5.8
     - feature/symfony-upgrade-3.4-retry
 
+after_failure:
+  - cat ~/build/OpenConext/OpenConext-engineblock/app/logs/test/test.log
+  - sudo cat /var/log/syslog
+
 addons:
   hosts:
     - engine.vm.openconext.org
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -1,26 +1,26 @@
 language: php
 
-dist: precise
+dist: xenial
```
pablothedude: Nice that this upgrade is done!

.travis.yml
```diff
@@ -1,26 +1,26 @@
...
-#  - 7.0
+  - 7.2
 
 matrix:
   fast_finish: true
-#  allow_failures:
-#  - php: 7.0
-#  - env: COMMAND="functional-tests-wip"
+  allow_failures:
+  - php: 7.2
```
pablothedude: A good decision to go to 7.2.

PR #620, opened 2019-01-25
Add Portuguese messages and missing "done" and "edit" for NL an EN

domgon:


domgon:
the security validations are still failing. Please can some one give a tip or help to solve this. Finally I understand the previews comments from an old Pull Request. The links are generic and the "done" and "edit" in Dutch are according the suggestions "bewerken" and "klaar". Hope we are starting to contribute correctly

PR #622, opened 2019-01-31
Filter AA response

MKodde:
AA response values that are non-string should be filtered out of the
AA client response. The EPTI caused this problem to surface. See the
ticket below for more information.

See: https://www.pivotaltracker.com/story/show/162197380

### DIFF
--- a/src/OpenConext/EngineBlockBundle/AttributeAggregation/Dto/Response.php
+++ b/src/OpenConext/EngineBlockBundle/AttributeAggregation/Dto/Response.php
@@ -40,7 +40,11 @@ final class Response
         $attributes = [];
 
         foreach ($jsonData as $attributeData) {
-            $attributes[] = self::parseAggregatedAttribute($attributeData);
+            $attribute = self::parseAggregatedAttribute($attributeData);
+            if (!$attribute) {
+                continue;
+            }
+            $attributes[] = $attribute;
         }
 
         $response = new self;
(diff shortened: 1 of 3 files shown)


src/OpenConext/EngineBlockBundle/AttributeAggregation/Dto/Response.php
```diff
@@ -67,9 +71,17 @@ private static function parseAggregatedAttribute(array $attributeData)
             throw new InvalidAttributeAggregationResponseException('Missing aggregated attribute source');
         }
 
+        // All values must be string. See: https://www.pivotaltracker.com/story/show/162197380
+        $values = (array) $attributeData['values'];
+        foreach ($values as $value) {
+            if (!is_string($value)) {
```
pablothedude: Are null values or integers not supported at all?

MKodde: In the story @quartje states:

> To solve this, all attributes that are not strings should be filtered out of the request to the AA

pablothedude: But are we sure that all data that till this point that should be a string is a string? The `urn:mace:dir:attribute-def:eduPersonTargetedID` attribute wasn't valid either?

MKodde: It's very strict now, and I suppose that is what we want here? If you want to pass a numeric value, it should be a string and not casted to an integer or other numeric data type.

PR #624, opened 2019-02-05
Prevent logo aspect ratio violation

MKodde:
The logo's where explicitly set with a width and height. This caused
the aspect ratio to fail. Using the max-width/max-height solution solves
this problem with relative ease.

This commit/PR addresses the first improvement described in:
https://www.pivotaltracker.com/story/show/159269295


pablothedude:
If you've fixed it feel free to merge

PR #625, opened 2019-02-05
Align logo and title on small viewports

MKodde:
Previously the logo and h2 title would be display: blocked. Causing them
to be rendered on two lines. This change will now display them next to
each other as they are now displayed inline-block.

In addition the h2 title is also positioned to vertically align with the
logo.

This fixes the second improvement described in:
https://www.pivotaltracker.com/story/show/159269295


pablothedude:
If you've fixed it feel free to merge

PR #631, opened 2019-02-07
Remove obsolete surfconextId attribute

MKodde:
This attribute is no longer used and therefor support can be dropped.
This means the attribute is no longer added explicitly to the response
attributes. All mentions of the attribute are also removed from the
project.

https://www.pivotaltracker.com/story/show/160788232


pablothedude:
These changes seems prefectly fine. I only saw these attributes also in the SmalBundle

MKodde:
> These changes seems prefectly fine. I only saw these attributes also in the SmalBundle

Yes I've opened a [PR](https://github.com/OpenConext/Stepup-saml-bundle/pull/85) there.

pablothedude:
:+1:

PR #641, opened 2019-03-13
Catch SSO endpoint [ab|mis]use earlier

MKodde:
Logging of incorrect requests on the SSO endpoint where quite cryptic. Often resulting in a generic 500 error page with a log entry reading: `Unable to find the current binding.`. 

This PR aims to improve the error reporting portion of this problem. The actual better user-facing error handling is addressed in an upcoming PR. This PR introduces a request validator that is used upon entry of the SSO controller action. It tests if a supported request method is used (GET is currently supported exclusively) and if the correct binding is used (HTTP binding is expected).

If the request does not match these criteria, a meaningfull exception is thrown. Either reflecting the use of an usupported request method. Or in the case of an usuported SAML request binding, will report which binding was attempted.

https://www.pivotaltracker.com/story/show/164121560

### DIFF
--- a/src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
+++ b/src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * Copyright 2019 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlock\Validator;
+
+use Exception;
+use OpenConext\EngineBlock\Exception\InvalidBindingException;
+use OpenConext\EngineBlock\Exception\InvalidRequestMethodException;
+use SAML2\Binding;
+use SAML2\HTTPPost;
+use SAML2\HTTPRedirect;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * The SsoRequestValidator verifies valid requests on the SSO endpoint
+ *
+ * Valid requests method / binding combinations for SSO are:
+ *
+ *  | SAML Binding     | Request method | Parameter name |
+ *  | ---------------- | -------------- | -------------- |
+ *  | HttpRedirect     | GET            | SAMLRequest    |
+ *  | HTTPPost         | POST           | SAMLRequest    |
+ */
+class SsoRequestValidator implements RequestValidator
+{
+    private $supportedRequestMethods = [Request::METHOD_GET, Request::METHOD_POST];
+
+    public function isValid(Request $request)
+    {
+        $requestMethod = $request->getMethod();
+        // Defense in depth; anything other than POST and GET are probably already rejected at routing time.
+        if (!in_array($requestMethod, $this->supportedRequestMethods)) {
+            // Only Redirect binding is supported for Single Sign On
+            throw new InvalidRequestMethodException(
+                sprintf('The HTTP request method "%s" is not supported on this SAML SSO endpoint', $requestMethod)
+            );
+        }
+        try {
+            $binding = Binding::getCurrentBinding();
+        } catch (Exception $e) {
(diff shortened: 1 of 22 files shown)


MKodde:
The build breaks on the sec-tests. They run into the Twig sandbox cve. Which is fixed in #643,

tests/unit/OpenConext/EngineBlock/Validator/SsoRequestValidatorTest.php
```diff
@@ -0,0 +1,113 @@
...
+     * @var SsoRequestValidator
+     */
+    private $validator;
+
+    public function setUp()
+    {
+        $this->validator = new SsoRequestValidator();
+
+        // PHPunit does not reset the superglobals on each run.
+        unset($_GET['SAMLRequest']);
```
pablothedude: Have you considered using the [backupGlobals](https://phpunit.de/manual/5.7/en/appendixes.annotations.html#appendixes.annotations.backupGlobals) annotation?

tests/unit/OpenConext/EngineBlock/Validator/SsoRequestValidatorTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $request = m::mock(Request::class);
+        $request
+            ->shouldReceive('getMethod')
+            ->andReturn(Request::METHOD_POST);
+        $this->assertTrue($this->validator->isValid($request));
+    }
+
+    /**
+     * @expectedException \OpenConext\EngineBlock\Exception\InvalidRequestMethodException
```
pablothedude: The `expectedException` seems to become deprecated, have you considered not using the annotations?

src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
```diff
@@ -0,0 +1,70 @@
...
+            $binding = Binding::getCurrentBinding();
+        } catch (Exception $e) {
+            throw new InvalidBindingException(
+                sprintf('No SAMLRequest parameter was found in the HTTP "%s" request parameters', $requestMethod),
+                0,
+                $e
+            );
+        }
+        if (!($binding instanceof HTTPRedirect || $binding instanceof HTTPPost)) {
+            // We only support HTTP Redirect binding
```
pablothedude: The check above suggests we also support Post binding?

src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
```diff
@@ -0,0 +1,70 @@
...
+            throw new InvalidBindingException(
+                sprintf('The binding type "%s" is not supported on this SAML SSO endpoint', get_class($binding))
+            );
+        }
+        return true;
```
pablothedude: Should we validate the existence of the `SAMLRequest` parameter?

tests/unit/OpenConext/EngineBlock/Validator/SsoRequestValidatorTest.php
MKodde: Yes, deprecated in phpunit 7 and removed in 8 if I recal correctly from the hackaton. It indeed would be good practice to start using the method expectation instead of relying on the annotations.

tests/unit/OpenConext/EngineBlock/Validator/SsoRequestValidatorTest.php
MKodde: Wow, no didn't know about that feature. Will definitely check that out!

src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
MKodde: We do actually! I was not aware of this, but we do. It is not advertised and we should document it somewhere in the wiki.

src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
MKodde: Let's discuss. We are validating the correctness of the request, and part of that is the existence of the correct request parameter (in this case the SAMLRequest). Would you move this check elsewhere? Or again rely on the SAML2 library, who would prevent us from catching a meaningful error message…

MKodde:
The last two commits are added under the boyscout rule. After Bas' feedback https://github.com/OpenConext/OpenConext-engineblock/pull/641#discussion_r265922562

tests/unit/OpenConext/EngineBlock/Validator/SsoRequestValidatorTest.php
MKodde: Reporting back: We actually use this as a PHPUnit setting in our phpunit.xml. There seems to be a bug in PHPUnit 5 where overriding this setting is not allowed using annotations. So I'm sticking with the previous solution for now. In newer PHPUnit versions this seems fixed.

PR #642, opened 2019-03-13
Improve sso error feedback

MKodde:
1. The RedirectToFeedbackPageExceptionListener catches the InvalidRequestMethodException and InvalidBindingException
2. The original exception message is levered into the error page as it  is very descriptive.
3. Finally the error message is rendered with a translated title (the  exception message is always displayed in English.).

This PR was created on top of #641 
See task 2 of: https://www.pivotaltracker.com/story/show/164121560

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
@@ -366,6 +366,33 @@ class FeedbackController
     }
 
     /**
+     * @return Response
+     */
+    public function noAuthenticationRequestReceivedAction(Request $request)
+    {
+        // The exception message is used on the error page. As mostly developers or other tech-savvy people will see
+        // this message. The ExceptionListener is responsible for setting the message on the feedback_custom field.
+        $session = $request->getSession();
+        if ($session->has('feedback_custom')) {
+            $message = $session->get('feedback_custom');
+        } else {
+            // This should never occur, when it does, this error page is called from outside the application context
+            // or the exception that shows this page was triggered elsewhere in code without a message.
+            $message = 'More elaborate error details could not be found..';
+        }
+
+        return new Response(
+            $this->twig->render(
+                '@theme/Authentication/View/Feedback/no-authentication-request-received.html.twig',
+                [
+                    'message' => $message,
+                ]
+            ),
+            400
+        );
+    }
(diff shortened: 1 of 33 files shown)


src/OpenConext/EngineBlockBundle/EventListener/RedirectToFeedbackPageExceptionListener.php
```diff
@@ -159,6 +161,10 @@ public function onKernelException(GetResponseForExceptionEvent $event)
         } elseif ($exception instanceof StuckInAuthenticationLoopException) {
             $message         = 'Stuck in authentication loop';
             $redirectToRoute = 'authentication_feedback_stuck_in_authentication_loop';
+        } elseif ($exception instanceof InvalidRequestMethodException | $exception instanceof InvalidBindingException) {
```
pablothedude: Did you deliberately use a bitwise operator?

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
```diff
@@ -365,6 +365,25 @@ public function stuckInAuthenticationLoopAction()
...
 
+    /**
+     * @return Response
+     */
+    public function noAuthenticationRequestReceivedAction(Request $request)
+    {
+        // The exception message is used on the error page. As mostly developers or other tech savy people will see this
+        // message. The ExceptionListener is responsible for setting the message on the feedback_custom field.
+        $session = $request->getSession();
+        $message = $session->get('feedback_custom');
```
pablothedude: Could we safely assume that the 'feedback_custom' parameter is always from the requestId we would expect?

src/OpenConext/EngineBlockBundle/EventListener/RedirectToFeedbackPageExceptionListener.php
MKodde: No, I did not intend using that! Well spotted :ok_hand:

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
MKodde: When used correctly, yes we can assume this message is set. But you are right. We cannot assume correct usage. I'm adding a safeguard.

PR #644, opened 2019-03-18
Put variables in exceptions between quotes

MKodde:
This is required to generate a good 'Art' code. The specificity of the messages has not been reviewed.

This PR covers task one of: https://www.pivotaltracker.com/story/show/164076266

PR #645, opened 2019-03-18
Show an error code on error page

MKodde:
This PR levers the ART implementation of StepUp and implements it into Engineblock.

The ART code is generated in the `collectFeedbackInfo`, and is based on the exception message of the exception that was raised. In order to generate consistent ART codes, #644 was opened to ensure variable texts in exception messages are put between quotes. These quoted variable strings are then ignored during ART generation.

Note that the commits of #644 are also in this commit.

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/X509/X509CertificateFactory.php
+++ b/src/OpenConext/EngineBlock/Metadata/X509/X509CertificateFactory.php
@@ -22,13 +22,13 @@ class X509CertificateFactory
         $pemString = file_get_contents($filePath);
 
         if (!$pemString) {
-            throw new RuntimeException("Unable to read file at path '$filePath'.");
+            throw new RuntimeException(sprintf('Unable to read file at path "%s".', $filePath));
         }
 
         try {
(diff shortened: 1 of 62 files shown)


src/OpenConext/EngineBlock/Metadata/X509/X509CertificateFactory.php
```diff
@@ -22,13 +22,13 @@ public function fromFile($filePath)
...
-            throw new RuntimeException("Unable to read file at path '$filePath'.");
+            throw new RuntimeException(sprintf('Unable to read file at path "%s".', $filePath));
         }
 
         try {
             $certificate = $this->fromString($pemString);
         } catch (Exception $e) {
-            throw new RuntimeException("File at '$filePath' does not contain a valid certificate.", 0, $e);
+            throw new RuntimeException(sprintf('File at "%s" does not contain a valid certificate.', $filePath), 0, $e);
         }
```
mroest: Having the filename in the log could potentially be a security problem? Not sure how it is handle by the app. Its just that you are aware

PR #646, opened 2019-03-20
Render 405 error pages more elegantly

MKodde:
The http 405 (request method not allowed) error handling has been made
more verbose for the users of engine block. A custom error screen was
added where the user is informed of the error situation.

On this screen we do not show the error table as we do during the
authentication process.

Task 1 of: https://www.pivotaltracker.com/story/show/164121663

### DIFF
--- a/languages/messages.pt.php
+++ b/languages/messages.pt.php
@@ -142,6 +142,8 @@ TXT
     // Error screens
     'error_404'                         => '404 - Página não encontrada',
     'error_404_desc'                    => 'Esta página não foi encontrada.',
+    'error_405'                         => 'Método HTTP não permitido',
+    'error_405_desc'                    => 'O método HTTP "%requestMethod%" não é permitido para o endereço "%uri%". Os métodos suportados são: %allowedMethods%.',
     'error_help_desc'               => '<p></p>',
     'error_no_consent'              => 'Não é possivel continua para o serviço',
     'error_no_consent_desc'         => 'Esta aplicação só pode ser utilizada quando a informação mencionada for partilhada.<br /><br />
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -160,6 +160,8 @@ TXT
     // Error screens
     'error_404'                         => '404 - Page not found',
     'error_404_desc'                    => 'This page has not been found.',
+    'error_405'                         => 'HTTP Method not allowed',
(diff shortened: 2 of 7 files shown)


MKodde:
@domgon would you be so kind to verify 'my' Portugese translations? I highly doubt the online translator created perfect translations :sweat_smile:

languages/messages.pt.php
```diff
@@ -142,6 +142,8 @@
     // Error screens
     'error_404'                         => '404 - Página não encontrada',
     'error_404_desc'                    => 'Esta página não foi encontrada.',
+    'error_405'                         => 'Método HTTP não permitido',
+    'error_405_desc'                    => 'O método HTTP "%requestMethod%" não é permitido para o local "%uri%". O método suportado são: %allowedMethods%.',
```
MKodde: The Portuguese 405 description was translated using Google translate.

domgon: Hi MKodde,

it's almost perfect :)

The message **error_405** is correct but the message for **error_405_desc** not.

So it should be:
'error_405_desc'                    => 'O método HTTP "%requestMethod%" não é permitido para o endereço "%uri%". Os métodos suportados são: %allowedMethods%.',

Regards

MKodde: Thanks for your quick reponse, Domingos! I've updated the translation.

MKodde:
Correct, as mentioned in the PR description. This PR only deals with the first task. And restricting the ACS endpoint to only allow POST requests is added in a successive PR. #647 to be precise.

pablothedude:
1. I saw that #647 will prevent POST methods but I'm unable to test this functionality in both PR's and I want to review / functional test a PR before approving it. Could you please explain when I could test all the tasks of the issue at once before merging them back?
2. If a user will post data to the  `/authentication/sp/consume-assertion` endpoint and an unexpected error occurs. If a user then refreshes the page the user will then be presented with the `MethodNotAllowed`-exception because of the `get` method which will be disallowed.

MKodde:
1. You could update your routing request method constraints effectively applying the change from #647. https://github.com/OpenConext/OpenConext-engineblock/pull/647/files#diff-70adaa3376f7527efc0344926e0839ec
2. Ah, yes I now catch your drift. In that case, you are right. Coming back to your previous statement that the first uncaught error page is more useful than the successive 405 page is not entirely true in my opinion. That error page shows us the `Oops something is wrong` page with minimal/no additional feedback. The real problem here would be to implement a redirect to a custom user-facing error page. Many occurrences of this problem would end up on the error log stats.

PR #647, opened 2019-03-20
Allow only POST requests on /sp/consume-assertion

MKodde:
We stated that GET was also allowed in order to render more user
friendly error reports. That was not at all the case. In a recent effort
error reporting regarding disallowed request methods was improved.

So we can now start enforcing POST requests on the ACS endpoint. Any
other request method will result in routing exception that in turn will
be handled by the improved error reporting.

Task 2 & 4 of: https://www.pivotaltracker.com/story/show/164121663
Also see: #646


MKodde:
The build fails on the expectation added in the Behat test. This will not yet work until we are rebased onto #646

PR #648, opened 2019-03-20
Allow repost of the SAMLResponse on consent

MKodde:
This provides a proof of concept of the possibility to repost the SAMLResponse and thus enabling us to reload the consent screen without yielding an error. See the discussion in Pivotal for more info on how I got to this POC. 

@pablothedude, do you agree with the chosen solution and how this was implemented in this PR? And what could/should be improved on?

https://www.pivotaltracker.com/story/show/164244128


pablothedude:
Great work, the only thing I'm wondering is why the `completeCurrentProcedure` call seems to be removed from the SPController and added to the IDPController. For me it's not clear why that should be done. Could you please explain?

src/OpenConext/EngineBlockBundle/Resources/config/services.yml
```diff
@@ -29,6 +29,11 @@ services:
         arguments:
             - "engine-api.%domain%"
 
+    engineblock.service.authentication_state_helper:
+        class: OpenConext\EngineBlock\Service\AuthenticationStateHelper
+        arguments:
+            - "@session"
+            -
```
pablothedude: Empty param typo?

MKodde: Good catch!

MKodde:
> The only thing I'm wondering is why the `completeCurrentProcedure` call seems to be removed from the SPController and added to the IDPController. For me it's not clear why that should be done. Could you please explain?

The commit message of 22eb97a should further explain this.

PR #649, opened 2019-03-21
Support multiple authentications

MKodde:
This progresses on #648 so this solution can also be seen as a proof of concept. 

We should discuss if the intermingling of new and old EB code is the way forward, or if we should bite the bullet and start pulling more of the Corto features to the Symfony namespace.

This PR consists of multiple changes.

1. In order to test the new feature, a Behat feature was provided which uses the Selenium2 stand-alone driver in combination with headless Chrome. 
2. Travis can run this headless Behat feature. Support was added in commit 'Run headless Chrome Behat tests on Travis '
3. Finally, the main change was made in commit: ec91e45 see commit message for more info on that one!

The next action will be to further mature this POC:
1. Ensure Travis runs the headless chrome tests correctly.
2. The Chrome tests are annotated with a Behat tag that is used to run these tests in their own suite.
3. OC Deploy installs ChromeDriver and Selenium2 (stand alone) in the dev env
4. The [enable|dsiable]-test-env ant tasks starts/stops selenium


src/OpenConext/EngineBlockBundle/Authentication/AuthenticationProcedureMap.php
```diff
@@ -0,0 +1,113 @@
...
+    {
+        $newAuthenticationProcedures = $this->authenticationProcedures;
+        $newAuthenticationProcedures[$requestId] = $authenticationProcedure;
+
+        return new self($newAuthenticationProcedures);
+    }
+
+    /**
+     * @param string $requestId
+     * @return mixed|AuthenticationProcedure|null
```
pablothedude: Mixed could presumably be omitted?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/MultipleSingleSignOn.feature
```diff
@@ -0,0 +1,38 @@
+@WIP
```
pablothedude: I would like a tag to allow running selenium tests separately or not at all.

MKodde:
Thanks for the review @pablothedude 

> I personally would prefer to use a VO as RequestId but I understand that refactoring currently won't yield enough benefits.

That crossed my mind, that crossed my mind and I think, if we go down this road, we should take the effort of making that update.

> Shouldn't the maximum number of concurrent authentication procedures be limited?

This was previously not limited either, but a good question none the less. Let's discuss this later today on the phone.

> Should an exception be thrown if a session with the same requestId is added to the map?

Yes, that is a good addition, although the chance of this happening is VERY small. Will fix this on this PR.

> I saw that the RequestId still uses an old sha1 hash, should it be changed?

Do you mean the `$prefix . sha1(uniqid(mt_rand(), true));` line in `EngineBlock_Saml2_IdGenerator_Default`? Other than SHA1 being unsafe for password hashing, I don't think we are at risk here using it? Don't you agree?

pablothedude:
Regarding the generated `ID` I don't think that this is a direct security risk but I thought it was used to compare the ID with the `InResponseTo` attribute. I don't know the saml2 spec that well to conclusively conclude that this shouldn't be changed on the long run.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationProcedureMap.php
MKodde: Yes, it can

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/MultipleSingleSignOn.feature
MKodde: Leaving at is for now. Feel free to run them in a separate suite in your upcoming PR!

PR #651, opened 2019-03-26
Move consent services from Corto to Symfony

MKodde:
Proof of concept of moving two Corto services into the Symfony application. 

The two consent services have been moved. Some findings:
1. There is a lot of coupling left to Corto classes. 
2. The benefit of this move might not outweigh the benefit (reviewer, I would appreciate your input)
3. The two tests are marked as skipped, awaiting approval or reversal of this POC.


pablothedude:
Great work done regarding the refactoring. But personally I don't think with the current knowledge that the work done completely justifies the time it will cost to move only small parts of the codebase from Corto without solving that much coupling. The POC certainly makes much more clear where the challenges regarding the maintainability of Corto lie.

Maybe it would be more valuable to make the Corto code more future proof by only removing the global variables and singletons by injecting them in the constructor by using the di-container (service manager)  for the moment. This so we could gradually implement DI in Corto. If we have DI implemented we are much more flexible and then we could go lesser 'big bang' in lesser time when the time is ripe. But that is something we should discuss.

Thanks for the obtained insights.

MKodde:
> I don't think with the current knowledge that the work done completely justifies the time it will cost to move only small parts of the codebase from Corto without solving that much coupling.

I agree with this.

> Maybe it would be more valuable to make the Corto code more future proof by only removing the global variables and singletons by injecting them in the constructor by using the di-container (service manager) for the moment. This so we could gradually implement DI in Corto. If we have DI implemented we are much more flexible and then we could go lesser 'big bang' in lesser time when the time is ripe. But that is something we should discuss.

Yes, that might be more helpful when we really start refactoring the legacy code. Which moving from Corto to Symfony also achieves. But at too high a cost in my opinion.

@thijskh Our shared conclusion is to not go down this route for now. I do share Bas' statement that trying to improve the current state of Corto (for example: removing use of superglobals and start implementing DI) is more valuable short, and long term. Short term as we make the code less error-prone due to the use of modern standards. And long term by making the code easier to understand and re-use when we are going to deprovision Corto.

For now, I suggest we do not merge this PR, but close it and let it be a valuable lesson learned.

PR #652, opened 2019-03-27
Optimize local selenium tests

pablothedude:
A sperate tag is added in behat to allow running the webdriver
tests separately from the Goutte tests. Also a seperate task is added
to ant to facilitate this change. Now we start selenium before running
the webdriver tests and stop the selenium service when we are done. This
is needed for OC-deploy.

Also configured Puppeteer to use the same headless Chrome instance and installed babel preset-env to auto detect the used js version.
https://babeljs.io/docs/en/babel-preset-env

https://www.pivotaltracker.com/story/show/164121029


README.md
```diff
@@ -27,7 +27,9 @@ In short, themes require compilation which can be done by running the following
 To setup the required tooling on the VM, the following steps might be useful:
 
     cd /opt/openconext/OpenConext-engineblock/theme
-    sudo yum install npm
+    sudo yum install gcc
+    sudo curl --silent --location https://rpm.nodesource.com/setup_8.x | sudo bash -
```
MKodde: This line, even though it is instructional, explicitly instructs the installation of node 8. Is that a hard constraint? Letting yum choose the current (probably even older) npm version is no longer an option?

README.md
```diff
@@ -27,7 +27,9 @@ In short, themes require compilation which can be done by running the following
 To setup the required tooling on the VM, the following steps might be useful:
 
     cd /opt/openconext/OpenConext-engineblock/theme
-    sudo yum install npm
+    sudo yum install gcc
```
MKodde: Some thoughts:
Maybe add an explanation of why `gcc` is needed in the first place? And why it no longer is installed in our VM?
We could also change our dev OC-deploy scripts to include the installation of `gcc`.

.travis.yml
```diff
@@ -58,9 +59,9 @@ before_script:
   - sudo a2dissite 000-default
   - sudo a2ensite default-ssl
   - sudo service apache2 restart
-  - ./ci/bin/1-install-chromedriver.sh
-  - ./ci/bin/2-install-selenium-standalone.sh
-  - java -jar /var/selenium/selenium-server-standalone-2.53.1.jar -Dwebdriver.chrome.driver=/usr/bin/chromedriver &
+  - sudo mkdir -p /opt/selenium && wget http://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.1.jar && sudo mv selenium-server-standalone-2.53.1.jar /opt/selenium/
```
MKodde: What are your thoughs on the installation of selenium and chromedriver? Installing explicit versions (who might become unavailable) seems a bit harder on maintenance. And make the Travis tests more fragile. I'm asking this because I was hoping we could improve on this after my initial setup. 

Maybe installing chrome driver from the yum repos would help?

build.xml
```diff
@@ -187,6 +188,21 @@
...
 
+    <target name="behat-webdriver-dev">
+        <echo message="Starting a new selenium instance" />
+        <antcall target="selenium-stop"/>
+        <antcall target="selenium-start"/>
+        <echo message="Preparing the dev env to run in functional testing mode" />
+        <antcall target="prepare-env"/>
+        <echo message="${line.separator}${line.separator}Running Behat"/>
+        <antcall target="run-behat">
+            <param name="suite" value="webdriver"/>
```
MKodde: I personally don't like the term 'webdriver'. It does not convay the actual nature of the tests. Other Behat tests we run could also be seen as tests run through a webdriver as Goutte, like all other drivers supported by Mink, are 'webdrivers' right?

As we are already explicit about the use of Selenium, I suggest we call these tests `behat-selenium-*` tests. And possibly renaming the original behat targets to include `goutte` in their names.

Have you experimented with running all tests through selenium2 with headless chrome? If performance is slightly worse, we can also consider migrating all tests to that driver.

tests/behat.yml
```diff
@@ -33,6 +33,36 @@ default:
...
                     mockIdpRegistry: '@engineblock.mock_entities.idp_registry'
                 - OpenConext\EngineBlockFunctionalTestingBundle\Features\Context\MinkContext
+        webdriver:
+            paths:
+                - '%paths.base%/../src/OpenConext/EngineBlockFunctionalTestingBundle/Features'
+            formatter:
+                name: progress
+            filters:
+                tags: "@webdriver&&~@SKIP&&~@WIP"
+            contexts:
```
MKodde: The contexts for the Goutte and Selenium drivers are one-to-one copies, right? Could we extract that config section, and include it on both `contexts` configuration sections? It might be a bad idea, as we will probably start using different configurations in the future, but for now, it might reduce some duplication.

theme/puppeteer/setup.js
```diff
@@ -12,8 +12,9 @@ module.exports = async function() {
         '--ignore-certificate-errors'
     ];
     const options = {
+        executablePath: '/usr/bin/google-chrome',
```
MKodde: :100:

build.xml
```diff
@@ -3,8 +3,9 @@
     <target name="build" depends="test-suites,code-quality,functional-tests,security-tests" />
     <target name="pre-commit" depends="test-suites,api-functional-tests,code-quality" />
     <target name="pre-push" depends="test-suites,code-quality,behat-dev-wip,security-tests" />
-    <target name="functional-tests" depends="api-functional-tests, behat-travis" />
+    <target name="functional-tests" depends="api-functional-tests,behat-travis" />
     <target name="functional-tests-wip" depends="behat-travis-wip" />
+    <target name="functional-tests-webdriver" depends="behat-travis-webdriver"/>
```
MKodde: On line 12, for the `js-smoke-tests`, doesn't this test require an explicit start/stop of selenium? To prevent cross-pollination? Or does Puppeteer take care of this?

README.md
pablothedude: The older node yielded in not being able to run the tests. It seems that a later version of npm is needed in order to detect the used js version in scripts.

build.xml
```diff
@@ -187,6 +188,21 @@
...
+        <antcall target="selenium-stop"/>
+        <antcall target="selenium-start"/>
+        <echo message="Preparing the dev env to run in functional testing mode" />
+        <antcall target="prepare-env"/>
+        <echo message="${line.separator}${line.separator}Running Behat"/>
+        <antcall target="run-behat">
+            <param name="suite" value="webdriver"/>
+        </antcall>
+        <echo message="${line.separator}${line.separator}Reverting the dev env to run in normal mode"/>
+        <antcall target="prepare-env"/>
```
MKodde: The second prepare-env is not required in my opinion. Previously this was run to ensure the dev cache was warmed up again, but now we just re-cache-clear the test cache. 

Is my assumption correct? Or am I overlooking something important here?

README.md
pablothedude: Gcc was dropped in production, so we need it for Ruby and Npm to compile dependencies:
https://github.com/OpenConext/OpenConext-deploy/commit/0539376c2c9b21e2902ccb7fbabb7cbd32aeb131

I've created a story to cleanup Ruby as dependency in our dev box and possibly update the dependencies in OC-deploy. Feel free to comment.
https://www.pivotaltracker.com/story/show/165016259

.travis.yml
pablothedude: Generally it would be safer to pinpoint versions so you always know which version is being used. If a version isn't available you'll know directly what the issue is. And f it's because of changes in your code you'll also know that because you will likely have touched the code.
If a version was changed without you knowing it you could run into some obscure exceptions which are often very hard to debug.

tests/behat.yml
pablothedude: I hoped that `default` could be overridden but that wasn't possible.

From the behat documentation:
> If you have multiple suites that all should use the same context, you will have to define that specific context for every specific suite:

build.xml
pablothedude: Puppeteer uses Chrome directly without the use of Selenium.

tests/behat.yml
MKodde: You might be able to use a node anchor? I have no previous experience with this YAML feature, but it might be worth a shot?

https://yaml.org/spec/1.2/spec.html#id2785586

build.xml
MKodde: Excellent!

build.xml
pablothedude: A good suggestion, I've changed the name of the tag from `webdriver` to `selenium`.

pablothedude:
1. Please update your PR description with additional information about the changes you applied
2. Squash your three commits into one, or squash merge this PR.
3. Add link to the Pivotal story in the PR and vice versa

1. Done
2. I would like to have them apart to be able to describe precisely which work was done. I did however update the titles
3. Done

tests/behat.yml
pablothedude: Tried, and seems to work as expected, thanks for the suggestion!

build.xml
pablothedude: It was a residue from earlier refactoring, which I didn't dare to remove. But after some research and IRL discussion I think I can assume it's safe to remove.

MKodde:
While starting a development task that also includes some front-end work. I noticed that one of your changes broke the Grunt 'pipeline'. The addition of the `preset-env` preset in the `.babelrc` file (8691e7d8) is not compatible with the `browserfy` task we run.

Removing the preset you added seems to fix the problem, but might interfere with the reason why you added it? Can you further explain why this preset is required?

PR #653, opened 2019-03-27
Honor explicit keyid in unsolicited single sign on request

thijskh:
When creating a new authnrequest based on a call to the
unsolicited-single-sign-on endpoint, do not forget to set
the keyid in this new request if it was specified in the
u-s-s-o URL.

This fixes the bug that assertions to SPs in response to unsolicited sso
calls would always be signed with the default key, regardless of any
explicit keyslugs in the unsolicited sso endpoint call.

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
+++ b/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
@@ -261,6 +261,10 @@ class EngineBlock_Corto_Module_Service_SingleSignOn extends EngineBlock_Corto_Mo
         }
 
         $request = new EngineBlock_Saml2_AuthnRequestAnnotationDecorator($sspRequest);
+        if ($keyid = $this->_server->getKeyId()) {
+            $request->setKeyId($keyid);
+        }
+
         $request->setUnsolicited();
 
         return $request;


library/EngineBlock/Corto/Module/Service/SingleSignOn.php
```diff
@@ -261,6 +261,10 @@ protected function _createUnsolicitedRequest()
         }
 
         $request = new EngineBlock_Saml2_AuthnRequestAnnotationDecorator($sspRequest);
+        if ($keyid = $this->_server->getKeyId()) {
```
MKodde: Aside from being no fan of combining an assignment with a conditional statement, this seems like a solid bugfix.

Can we further verify this behavior by adding a Behat test to the `UnsolicitedSingleSignOn` Behat feature?

Also, let's not forget to add this bugfix in the changelog. Am I correct this bugfix is targetted at release 5.10?

thijskh: 1) Might be a matter of preference, but it happens the same in another place where we do this.
2) Would be nice, but our tests do not currently have any infrastructure for testing keyslugs in any context, unsolicited or not. So it's for me a bit out of scope for the bugfix.
3) I think it might be nice if we could also cherry pick it for 5.9.

MKodde: 1. And of coding style/standard. For new EB code, we explicitly enforce PSR-2 compliance. And this construction kind of violates '[PSR 2: 2.3 - lines](https://www.php-fig.org/psr/psr-2/#23-lines): There MUST NOT be more than one statement per line.'. But I'm totally fine with keeping it like this.
2. Agreed, it still might be a nice feature to our testing toolchain. Adding it to the icebox.
3. No problem, dont forget updating the changelists!

PR #654, opened 2019-03-28
Fix unexpected SSO behaviour for trusted proxy

MKodde:
This bugfix adds a Behat scenario the reproduces the behavior described in the bug report. It then proceeds to fix the issue by preventing IdP scope limiting based on previously selected IdPs.

More details about this bug can be found in:
https://www.pivotaltracker.com/story/show/164069793

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
+++ b/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
@@ -25,6 +25,13 @@ class EngineBlock_Corto_Module_Service_SingleSignOn extends EngineBlock_Corto_Mo
         $log->info(sprintf("Fetching service provider matching request issuer '%s'", $request->getIssuer()));
         $sp = $this->_server->getRepository()->fetchServiceProviderByEntityId($request->getIssuer());
 
+        // When dealing with an SP that acts as a trusted proxy, we should perform SSO on the proxying SP and not the
+        // proxy itself.
+        if ($sp->isTrustedProxy) {
+            // Overwrite the trusted proxy SP instance with that of the SP that uses the trusted proxy.
+            $sp = $this->findOriginalServiceProvider($request, $log);
+        }
+
         // Exposing entityId to be used when tracking the start of an authentication procedure
         $application->authenticationStateSpEntityId = $sp->entityId;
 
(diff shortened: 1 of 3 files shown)


MKodde:
@pablothedude the last commit is the experiment we discussed over the landline with @quartje. Can you review this critically? If you see any problems with that commit, I'll drop it and go with 4ae4893

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
```diff
@@ -295,6 +295,26 @@ public function spIsAuthenticatingForSp($spName, $spDestinationName)
 
         $this->mockSpRegistry->save();
     }
+
+    /**
+     * @Given /^SP "([^"]*)" is authenticating for SP "([^"]*)" resetting the RequesterID chain$/
```
pablothedude: I find it hard to understand what this rule does. Maybe it could be more descriptive if you see fit.

library/EngineBlock/Corto/Module/Service/SingleSignOn.php
```diff
@@ -67,18 +74,13 @@ public function serve($serviceName)
         // The scope is further limited to the previously used IDP that is
         // allowed for the current issuer, this way the user does not need to
         // go trough the WAYF on subsequent logins.
-        //
-        // This scoping is not applied when the SP is a trusted proxy, as we
-        // can not be certain the SP that is proxied for had the same IdP settings
-        // as the previous one.
-        $scopedIdps = [];
-        if (!$sp->isTrustedProxy) {
```
pablothedude: Is the IDP for a trusted proxy SP the same IDP as the SP being proxied? Or why was that comment there?

MKodde: No the allowed IdP's for the proxied SP's behind a trusted proxy can differ, see the Pivotal story for more information.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
MKodde: It is kind of utilitarian, and that makes it harder to understand. When using the normal `SP "([^"]*)" is authenticating for SP "([^"]*)"` step, that would result in adding a new SP to the requester chain. We need to reset the chain, in order to simulate the behavior described in the bug report. I was not able to think of a better, more descriptive step definition. Do you have a suggestion?

MKodde:
FYI: I dropped the initial commit, containing the IdP scoping quick fix. That solution was dropped in favor of the more elegant solution that was proposed by @quartje

PR #655, opened 2019-03-28
Improve error reporting for clock related SAMLResponse issues

MKodde:
This PR adds a custom error page for when the issuer instant of the assertion is before/after the allowed time determined by the SAML2 response validator.

The individual commits should give additional information.

Test-tip:
Either manipulate your IdP to send out of time saml responses, or update the response before validating it in the bindins module.

1. Go to: `\EngineBlock_Corto_Module_Bindings::receiveResponse` at line 264
2. Override the NotOnOrAfter of the first assertion by adding line: `$sspResponse->getAssertions()[0]->setNotOnOrAfter(0);` to mimick an idp from the past
3. Override the `$sspResponse->getAssertions()[0]->setNotBefore(strtotime(date('Y-m-d H:i:s', strtotime('+1 year'))));` to mimick an idp from the future

This fixes tasks 1 and 2 of: https://www.pivotaltracker.com/story/show/164076668
~~The remaining task is to set a meaningful error title and text.~~

### DIFF
--- a/languages/messages.pt.php
+++ b/languages/messages.pt.php
@@ -240,6 +240,8 @@ A sua %organisationNoun% negou-lhe acesso a este serviço. Terá de entrar em co
     'error_attribute_validator_availability'        => '\'%arg3%\' é um schacHomeOrganization reservado para outro Fornecedor de Identidade',
     'error_unknown_service'         => 'Erro - Serviço desconhecido',
     'error_unknown_service_desc'    => '<p>O serviço solicitado não foi encontrado.</p>',
+    'error_clock_issue_title' => 'Erro - A asserção ainda não é válida ou expirou',
+    'error_clock_issue_desc' => '<p>Por favor, verifique se a hora no IdP está correta.</p>',
     'attributes_validation_succeeded' => 'Autenticação com sucesso',
     'attributes_validation_failed'    => 'Alguns atributos falharam na validação',
     'attributes_data_mailed'          => 'Os dados dos atributos foram enviados',
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -414,6 +414,19 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
             ]);
         }
         catch (ResponseProcessingFailedException $e) {
+            // The clock on the IdP side must be checked
+            // @see \SAML2\Assertion\Validation\ConstraintValidator\NotBefore::validate
+            // @see \SAML2\Assertion\Validation\ConstraintValidator\NotOnOrAfter::validate
+            if (strpos($e->getMessage(), 'Received an assertion that is valid in the future') !== false ||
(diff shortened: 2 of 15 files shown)


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -414,6 +414,19 @@ public function receiveResponse()
             ]);
         }
         catch (ResponseProcessingFailedException $e) {
+            // The clock on the IdP side must be checked
+            // @see \SAML2\Assertion\Validation\ConstraintValidator\NotBefore::validate
+            // @see \SAML2\Assertion\Validation\ConstraintValidator\NotOnOrAfter::validate
+            if (strpos($e->getMessage(), 'Received an assertion that is valid in the future') !== false ||
```
pablothedude: I don't think this construction is really future proof because we're relying on error messages in third party code. I understand that it could possibly be hard to circumvent this solution. If that's not possible then I would suggest adding a test in our codebase to check for these messages. So if they do change we'll know.

MKodde: > I don't think this construction is really future proof because we're relying on error messages in third party code. 

That is my biggest concern too, your suggestion:

> I would suggest adding a test in our codebase to check for these messages. So if they do change we'll know.

Seems like a smart thing to do!

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -414,6 +414,19 @@ public function receiveResponse()
             ]);
         }
         catch (ResponseProcessingFailedException $e) {
+            // The clock on the IdP side must be checked
```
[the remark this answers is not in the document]
MKodde: For all I know, the specific problem I'm solving here does not have a relationship with the validator you mention. 

@quartje or @thijskh or @pmeulen is the error message (which is an excerpt) below a common occurrence in the EB logs?

`Received an assertion with a session that has expired.`

library/EngineBlock/Corto/Module/Bindings.php
pablothedude: :100:

languages/messages.pt.php
```diff
@@ -240,6 +240,8 @@
     'error_attribute_validator_availability'        => '\'%arg3%\' é um schacHomeOrganization reservado para outro Fornecedor de Identidade',
     'error_unknown_service'         => 'Erro - Serviço desconhecido',
     'error_unknown_service_desc'    => '<p>O serviço solicitado não foi encontrado.</p>',
+    'error_clock_issue_title' => 'Erro - A asserção ainda não é válida ou expirou',
```
MKodde: @domgon would you be so kind to verify if the Portuguese translation makes sense. We already merged this feature, but I will open a new PR with an updated translation if need be.

domgon: Hi,
it's all must perfect, I thinks it's better as bellow.
'error_clock_issue_title' => 'Erro - A asserção ainda não é válida ou pode ter expirado',

MKodde: Thanks!

PR #656, opened 2019-04-01
Prevent blurring the search bar

MKodde:
When the mouse moves away from the search bar, previously the focus was blurred from the input field.

This behavior made using the search bar very un-user-friendly. Removing this behavior seems to fix the issue at hand.

Fixes https://www.pivotaltracker.com/story/show/165020904

### DIFF
--- a/theme/material/javascripts/application.js
+++ b/theme/material/javascripts/application.js
@@ -69,7 +69,7 @@ function initialize() {
         configuration.requestAccessUrl
     );
     const keyboardListener         = new KeyboardListener(idpPicker, $searchBar, requestAccessModalHelper);
-    const mouseListener         = new MouseListener(idpPicker, $searchBar, requestAccessModalHelper);
+    const mouseListener         = new MouseListener(idpPicker, requestAccessModalHelper);
 
     // Keyup, click and input are registered events for cross-browser compatibility with HTML5 'search' input
     $searchBar.addEventListener('keyup', throttle(event => idpPicker.searchBy(event.target.value), throttleAmountInMs));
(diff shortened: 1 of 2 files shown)


MKodde:
@pablothedude I rolled back one of the changes you explicitly made to fix a previous problem. I was not able to find the fine details on why that change was required. I hope you can help in giving more insight into this.

pablothedude:
@MKodde this was a feature and is needed in order to fulfill the requirement:
`mouse overrides keyboard and vice versa`
https://github.com/OpenConext/OpenConext-engineblock/pull/575#issuecomment-409190565

MKodde:
Thanks for pointing that out. Will verify that this change does not interfere with that requirement.

MKodde:
:warning: Rebased this branch on top of #658, but should be rooted back on master after that PR is merged.

MKodde:
The proposed fix does not seem to break/interfere with the feedback found in https://github.com/OpenConext/OpenConext-engineblock/pull/575#issuecomment-409190565. I tested in IE11 on Windows 8.1 and on Firefox on Ubuntu 18.10.

MKodde:
:warning: Backport this to the 5.9 release branch!

MKodde:
:information_source: Rebased onto master

pablothedude:
I saw that earlier on a :focus style was removed so the logic to blur the focus of the searchbar wasn't needed anymore. I've tested it and it seems to work as expected.

MKodde:
Thanks for checking that out @pablothedude !

Merging

PR #658, opened 2019-04-02
Repair build task and JS test collision

MKodde:
Grunt and Jest share the Babel library and both use it slightly different (version wise for example). The recent upgrade of Babel caused Grunt to fail its Browserfy step (which uses Babel).

By upgrading Babelify to a recent version, Grunt and Jest now both use the latest version of Babel. Effectively fixing the issue at hand.

In the process, I also installed two upgrades relating to a security vulnerability in the Grunt task runner. (ed13ce3)


MKodde:
:warning: Backport this to the 5.9 release branch!

theme/package.json
```diff
@@ -10,14 +10,15 @@
   },
   "author": "",
   "devDependencies": {
+    "@babel/core": "^7.4.0",
     "@babel/preset-env": "^7.4.2",
     "autoprefixer-core": "^5.2.0",
     "babel-plugin-transform-es3-member-expression-literals": "^6.8.0",
     "babel-plugin-transform-es3-property-literals": "^6.8.0",
     "babel-preset-es2015": "^6.14.0",
```
pablothedude: We could drop this line, I forgot to remove this after using preset-env.

theme/Gruntfile.js
```diff
@@ -19,7 +19,7 @@ module.exports = function(grunt) {
                         [
                             "babelify", {
                                 presets: [
-                                    ['es2015', {'loose': true}]
+                                    ["@babel/preset-env"]
```
pablothedude: Should we also specify a target because default preset-env will default to es2015?
https://babeljs.io/docs/en/babel-preset-env#targets

theme/package.json
MKodde: I tried that, but that breaks the Babelify task as well. As the `babel-plugin-transform-es3-*` plugins depend on it.

theme/Gruntfile.js
MKodde: That would make more verbose which browsers we support. We would need input to create this list, but I'm not certain how to configure this at this point as it is not perfectly clear which browsers we actually support.

My suggestion is to keep it this way, and when we know more about the actual targets, we configure them.

pablothedude: A good proposal!

PR #661, opened 2019-04-03
Make sure the mouse responds after removing previous IdP selection

MKodde:
The update focus was not called when the previous selection list was hidden from view, because all items where removed. This only was called when the done button was pressed. This update ensures that when the last delete button is pressed, we also call updateFocus on the IdP list. Effectively preparing the IdP list for interaction with mouse and keyboard.

See: https://www.pivotaltracker.com/story/show/165021022


MKodde:
:warning: Also backport this fix to EB 5.9

theme/material/javascripts/modules/IdpPicker.js
```diff
@@ -154,6 +154,12 @@ export class IdpPicker extends AbstractIdpPicker {
             $deletionButtons[i].removeEventListener('click', this.deleteIdpFromPreviousSelectionClickHandler());
             $deletionButtons[i].addEventListener('click', this.deleteIdpFromPreviousSelectionClickHandler());
         }
+
+        // When the last previous selection is removed, the previous chosen block is hidden, but we need an
+        // explicit call to updateFocus, in order to re-register the mouse listeners on the connected IdPs.
+        if ($deletionButtons.length === 0) {
```
pablothedude: I think it would be neater to move this check to the `deleteIdpFromPreviousSelection` method? And check the size of  `this.previousSelectionList` instead. So the `updateDeleteIdpFromPreviousSelectionClickListeners` method will only update the click listeners as the method would suggest.

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/WayfController.php
```diff
@@ -36,7 +36,7 @@ public function wayfAction(Request $request)
         return new Response($this->twig->render(
             '@theme/Authentication/View/Proxy/wayf.html.twig',
             [
-                'action' => 'https://continue-to-idp.example.com',
+                'action' => 'https://engine.vm.openconext.org/',
```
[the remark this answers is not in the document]
MKodde: I like that idea, can you elaborate your idea in a Pivotal ticket?

theme/material/javascripts/modules/IdpPicker.js
MKodde: Yes, I agree. Let's try moving it.

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/WayfController.php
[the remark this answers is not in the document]
pablothedude: https://www.pivotaltracker.com/story/show/165085055

theme/material/javascripts/modules/IdpPicker.js
```diff
@@ -168,10 +168,17 @@ export class IdpPicker extends AbstractIdpPicker {
         const idp = this.previousSelectionList.getFilteredIdpByIndex(index);
         this.previousSelectionList.removeIdpByIndex(index);
         this.idpList.addIdp(idp);
-
-        this.updateDeleteIdpFromPreviousSelectionClickListeners();
-        this.indexElements();
-        this.focusOnFirstIdp();
+        const numberOfdeletionButtons = this.targetElement.querySelectorAll('.result .deleteable.action').length;
```
pablothedude: Why not use `this.previousSelectionList.getLengthOfFilteredList()` ? I don't mind however if you keep it this way.

MKodde: Great suggestion, fixed it!

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/WayfController.php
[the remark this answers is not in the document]
MKodde: :heavy_check_mark:

PR #662, opened 2019-04-03
Color unconnected IdPs on mouse hover

MKodde:
Update focus after adding the disconnected IdPs

Doing so ensures the index is updated for the entire IdP list (connected
and unconnected) resulting in correct mouse behavior


MKodde:
Also see: https://www.pivotaltracker.com/story/show/165056451

PR #663, opened 2019-04-04
Replace Compass with node-sass

MKodde:
These tasks have been applied:
1. Node SASS was installed
2. Grunt-sass was installed
3. Compass task removed from Gruntfile
4. Compass stuff was stripped from the EB sass files
5. Compass reset is missing now, resulting in messed up styling
6. Update readme

Todo:
1. ~~Remove compass and ruby from the project and OC-deploy~~ Not required as these have already been stripped. An update of the readme should be sufficient.


README.md
```diff
@@ -27,14 +27,9 @@ In short, themes require compilation which can be done by running the following
 To setup the required tooling on the VM, the following steps might be useful:
 
     cd /opt/openconext/OpenConext-engineblock/theme
-    sudo yum install gcc
```
pablothedude: Don't we need this perhaps for npm?

README.md
```diff
@@ -27,14 +27,9 @@ In short, themes require compilation which can be done by running the following
...
     sudo curl --silent --location https://rpm.nodesource.com/setup_8.x | sudo bash -
     sudo yum install nodejs
     sudo npm install -g grunt grunt-cli
-    
-    sudo yum install gem ruby-devel
```
pablothedude: :100:

pablothedude:
There was an error while installing all dependencies through npm. I've fixed that.

README.md
MKodde: Judging from your comment below:

> There was an error while installing all dependencies through npm. I've fixed that.

We might need to add `sudo yum install gcc` back in the README.md?

PR #664, opened 2019-04-04
Run the smoke tests sequentially

MKodde:
This prevents unpredictable test results. Some tweaks to the timing have
been updated to make tests faster


MKodde:
:crossed_fingers: @pablothedude can you verify if this now works on Travis as I need to go home. If one of the tests still fail, its probably due to a timeout issue. These can be tweaked by upping the timeout time in the test.

pablothedude:
@MKodde there seemed to be also a problem with a race condition using `page.click` and `page.waitForNavigation`. I've added a promise which wait on both methods to finish and this seem to work as expected.

I'll merge this PR directly to let the build pass on master.

MKodde:
Excellent! Thanks for taking over this PR.

PR #665, opened 2019-04-04
Add Saml binding validator to acs endpoint

pablothedude:
This is done in order to make sure that the 'Current authentication
procedure exceptions' are split up. So the information returned
should be much more informative.

### DIFF
--- a/src/OpenConext/EngineBlock/Validator/SamlBindingValidator.php
+++ b/src/OpenConext/EngineBlock/Validator/SamlBindingValidator.php
@@ -20,14 +20,13 @@ namespace OpenConext\EngineBlock\Validator;
 
 use Exception;
 use OpenConext\EngineBlock\Exception\InvalidBindingException;
-use OpenConext\EngineBlock\Exception\InvalidRequestMethodException;
 use SAML2\Binding;
 use SAML2\HTTPPost;
 use SAML2\HTTPRedirect;
 use Symfony\Component\HttpFoundation\Request;
 
 /**
- * The SsoRequestValidator verifies valid requests on the SSO endpoint
+ * The SamlBindingValidator verifies valid saml binding
  *
  * Valid requests method / binding combinations for SSO are:
  *
@@ -38,23 +37,13 @@ use Symfony\Component\HttpFoundation\Request;
  */
 class SamlBindingValidator implements RequestValidator
 {
-    private $supportedRequestMethods = [Request::METHOD_GET, Request::METHOD_POST];
-
     public function isValid(Request $request)
     {
-        $requestMethod = $request->getMethod();
-        // Defense in depth; anything other than POST and GET are probably already rejected at routing time.
-        if (!in_array($requestMethod, $this->supportedRequestMethods)) {
-            // Only Redirect binding is supported for Single Sign On
-            throw new InvalidRequestMethodException(
-                sprintf('The HTTP request method "%s" is not supported on this SAML SSO endpoint', $requestMethod)
-            );
-        }
         try {
             $binding = Binding::getCurrentBinding();
         } catch (Exception $e) {
             throw new InvalidBindingException(
-                sprintf('No SAMLRequest or SAMLResponse parameter was found in the HTTP "%s" request parameters', $requestMethod),
+                sprintf('No SAMLRequest or SAMLResponse parameter was found in the HTTP "%s" request parameters', $request->getMethod()),
                 0,
                 $e
             );
--- a/src/OpenConext/EngineBlock/Validator/SamlBindingValidator.php
+++ b/src/OpenConext/EngineBlock/Validator/SamlBindingValidator.php
@@ -31,12 +31,12 @@ use Symfony\Component\HttpFoundation\Request;
  *
  * Valid requests method / binding combinations for SSO are:
  *
- *  | SAML Binding     | Request method | Parameter name |
- *  | ---------------- | -------------- | -------------- |
- *  | HttpRedirect     | GET            | SAMLRequest    |
- *  | HTTPPost         | POST           | SAMLRequest    |
+ *  | SAML Binding     | Request method | Parameter name              |
+ *  | ---------------- | -------------- | --------------------------- |
+ *  | HttpRedirect     | GET            | SAMLRequest,SAMLResponse    |
+ *  | HTTPPost         | POST           | SAMLRequest,SAMLResponse    |
  */
-class SsoRequestValidator implements RequestValidator
+class SamlBindingValidator implements RequestValidator
 {
     private $supportedRequestMethods = [Request::METHOD_GET, Request::METHOD_POST];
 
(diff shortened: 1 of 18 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/164238335

src/OpenConext/EngineBlock/Validator/SamlBindingValidator.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
MKodde: I like the fact that you seperated the binding and request validations into separate validators.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -91,8 +91,11 @@ public function authenticatedAt($requestId, Entity $identityProvider)
 
         if ($currentRequest === null) {
             throw new LogicException(
-                'Current authentication procedure cannot be authenticated:'
-                 . ' authentication procedure has not been started'
+                sprintf(
```
MKodde: The exception messages do not sit right with me, for example: 

> Requested authentication procedure cannot be found:  authentication procedure could not be found for given requestId (%s)

Can be shortened as both sentences now almost convey the same message (redundancy). I also suggest adding more information, to make the exception more easily understandable.

For example:

> The requested authentication procedure with requestId (%s) couldn't be found in the session storage.

MKodde: The same applies for the two other exception texts in this file.

src/OpenConext/EngineBlockBundle/Controller/IdentityProviderController.php
```diff
@@ -115,13 +119,16 @@ public function singleSignOnAction(Request $request, $keyId = null, $idpHash = n
...
+     * @param Request $request
      * @param null|string $keyId
      * @param null|string $idpHash
      * @return \Symfony\Component\HttpFoundation\RedirectResponse|\Symfony\Component\HttpFoundation\Response
-     * @throws Exception
      */
-    public function unsolicitedSingleSignOnAction($keyId = null, $idpHash = null)
+    public function unsolicitedSingleSignOnAction(Request $request, $keyId = null, $idpHash = null)
     {
+        $this->requestValidator->isValid($request);
```
MKodde: Thanks for fixing this!

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
```diff
@@ -278,16 +278,16 @@ Feature:
...
+     And I should see ART code "35954"
 
   Scenario: The SP uses the wrong request parameter while using HTTP Post binding
    Given the SP "Dummy SP" sends a malformed AuthNRequest
      And the SP "Dummy SP" uses the HTTP POST Binding
     When I log in at "Dummy SP"
      And I pass through the SP
-    Then I should see "No SAMLRequest parameter was found in the HTTP \"POST\" request parameters"
-     And I should see ART code "18993"
+    Then I should see "The parameter \"SAMLRequest\" is missing on this SAML SSO endpoint"
```
MKodde: Where you not able to add Behat scenarios for SAMLResponse validation violations. I really value the addition of those tests!

tests/unit/OpenConext/EngineBlock/Validator/AcsRequestValidatorTest.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+namespace OpenConext\EngineBlock\Validator;
+
+use OpenConext\EngineBlock\Exception\InvalidRequestMethodException;
+use PHPUnit_Framework_TestCase as TestCase;
+use Symfony\Component\HttpFoundation\Request;
+
+class AcsRequestValidatorTest extends TestCase
```
MKodde: Throwing of the `MissingParameterException` is not explicitly tested, also goes for the SSO request validator.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
pablothedude: I'll wait on the outcome of the discussion in Pivotal.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
pablothedude: I agree, I shall add this.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
pablothedude: Changed the texts.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -110,15 +112,19 @@ public function completeCurrentProcedure($requestId)
         $currentRequest = $this->authenticationProcedures->find($requestId);
         if ($currentRequest === null) {
             throw new LogicException(
-                'Current authentication procedure cannot be completed:'
-                . ' authentication procedure has not been started'
+                sprintf(
+                    'The requested authentication procedure with requestId (%s) couldn\'t be found in the session storage.',
```
MKodde: This is not what actually happened right? The authn procedure could not be started, this needs fixing. Ditto for the exception message on line 124

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
MKodde: Thanks for the update, see my new feedback below.

src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
```diff
@@ -50,21 +46,14 @@ public function isValid(Request $request)
...
-        }
-        if (!($binding instanceof HTTPRedirect || $binding instanceof HTTPPost)) {
-            // We only support HTTP Redirect binding
-            throw new InvalidBindingException(
-                sprintf('The binding type "%s" is not supported on this SAML SSO endpoint', get_class($binding))
+
+        if (($requestMethod ===  Request::METHOD_POST && !$request->request->has('SAMLRequest')) ||
+            ($requestMethod ===  Request::METHOD_GET && !$request->query->has('SAMLRequest'))) {
+            throw new MissingParameterException(
+                sprintf('The parameter "SAMLRequest" is missing on this SAML SSO endpoint')
```
MKodde: This new exception message is not correct in my opinion. An endpoint can not have a missing request parameter right? The request that was fired at the endpoint is missing the parameter. My suggestion is to reinstate the previous message. Also goes for the Acs validator counterpart.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
pablothedude: This is the mesasge as suggested in:
https://github.com/OpenConext/OpenConext-engineblock/pull/665#discussion_r273446033

MKodde: Yes, the proposal was only for the first exception message. The other two need a similar revision in my opinion, but the copy-paste is not correct and make the error messages incorrect.

pablothedude: Done

src/OpenConext/EngineBlock/Validator/SsoRequestValidator.php
pablothedude: Changed the message

PR #666, opened 2019-04-05
Update release script and  remove minified assets from VCS

pablothedude:
A .gitattributes is added to prevent merge conflicts when merging
the javascript.min.js. Also updated the makeRelease script to
generate the assets before releasing to have the latest version without
a merged minified js file which could be damaged.


pablothedude:

.gitattributes
```diff
@@ -0,0 +1 @@
+web/javascripts/application.min.js -diff merge=ours
```
MKodde: This is quite useful, but wouldn't the more pragmatic approach be:

1. Remove the minified content from VCS altogether
2. Add .gitignore for these resources.
3. Rely on the stylesheets being generated from the grunt task

bin/makeRelease.sh
```diff
@@ -40,17 +40,22 @@ echo "Preparing environment" &&
...
 cd ${PROJECT_DIR} &&
 git checkout ${TAG} &&
 
-echo "Running Composer Install";
+echo "Running Composer Install" &&
 php ./bin/composer.phar install -n --no-dev --prefer-dist -o &&
 
+echo "Build assets" &&
+cd ${PROJECT_DIR}/theme &&
+npm install &&
```
MKodde: Adding the production flag might be a good idea here.

pablothedude: It seems that we need the dev dependencies in order to build the assets themselves. Grunt and all grunts dependencies are dev dependencies. So I cannot remove them. I also like the idea that in travis and in the release script the options are the same. If you would like this changed I suggest to create an issue in Pivotal so we could discuss the pros and cons.

.gitattributes
pablothedude: Great suggestion, we then are even more sure that all assets are tested and the same or not pesent.

.gitignore
```diff
@@ -4,6 +4,7 @@
 bin/ignore_me.php
 /web/bundles
 /web/app_dev.php
+/web/javascripts/application.min.js
```
MKodde: Now that I'm looking at it for a second time, the following question arises; Why did you not apply the same solution for the other generated resources?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The 5.10 release focuses on making EngineBlock more robust. The biggest aim is t
  * Allow re-posting of SAML Responses, allowing page refresh on Consent screen #648
  * Allow for more than one in-flight authentication session per user session #649
  * Added further support for Portuguese language #650 (thanks @domgon!)
+ * Update release script and  remove minified assets from VCS #666
```
MKodde: Either change the solution so that it applies for all asset__s__, or remove the trailing 's' in the CHANGELOG :)

.gitignore
pablothedude: https://github.com/OpenConext/OpenConext-engineblock/pull/667

PR #667, opened 2019-04-09
Remove all assets from vcs

pablothedude:
Earlier on the minified js assets were already moved. To be uniform all
assets are now removed.


pablothedude:

PR #670, opened 2019-04-09
Error page styling

MKodde:
__Error screen redisign__
This is the first stage of styling the error page. The BEM methodology was applied to style the error screen. Some elements of the existing styling have been reused. This prevents duplicating the base styling.

This PR does not fully cover the design implementation that was proposed in https://www.pivotaltracker.com/story/show/164121204 but is a step in the right direction. The screenshot shows the current state of work.

Screenshot from 2019-04-09 11-48-30

Screenshot from 2019-04-09 12-06-23

__Test tips__
1. To view an error screen without feedback information, you could update the routing, for example removing the GET request method from the SSO endpoint. This triggers the 405 error page which has no feedback info
2. This errror page has a long text: https://engine.vm.openconext.org/authentication/feedback/unknown-preselected-idp

__Things to do:__
* ~~Make the design responsive~~
* ~~Apply the correct font~~
* Do something about the feedback information displaying
* ~~Finetune the footer button behavior~~
* ~~Add the surf-board to the background~~
* Browser testing


MKodde:
:warning: Note that this branch is targeted at release/5.11! We will not include the restyle in release 5.10

thijskh:
@arnoutt

pablothedude:
Great initial work!

Some initial thoughts:
- Should a footer partial be created?
- Is a mailto link in the footer the way to go?
- Footer links need a target blank
- Do we need to support ie8?
- Should we now drop some feedback-error boxes?
- The integrity sha is only changed from sha1 to sha512 in package-lock.json, this could be dropped. Maybe you could try running: https://github.com/npm/npm/issues/16938#issuecomment-339863980 to prevent this in the future.

MKodde:
>     * Should a footer partial be created?

Done! great suggestion!

>     * Is a mailto link in the footer the way to go?

I'm not sure, will ask about this in Pivotal

>     * Footer links need a target blank

Updated

>     * Do we need to support ie8?

I don't think so, but adding a background color fallback for these older browsers is something I believe to be good practice. 

>     * Should we now drop some feedback-error boxes?

Asked @arnoutt about this on the Pivotal ticket

>     * The integrity sha is only changed from sha1 to sha512 in package-lock.json, this could be dropped. Maybe you could try running: [npm/npm#16938 (comment)](https://github.com/npm/npm/issues/16938#issuecomment-339863980) to prevent this in the future.

Lets see if we stop having this issue, our Node/Npm version mis match might also have something to do with this.

MKodde:
Current state of affairs:

Screenshot from 2019-04-10 11-22-04

MKodde:
Rebased the work onto master@2d991ef4

theme/material/stylesheets/application/base/_layout.css.sass
```diff
@@ -3,14 +3,12 @@
   padding: 1rem
   .lt-ie9 &
     max-width: $container-width-em
-    padding: 1em
+    margin: 0 auto 3em
```
MKodde: Is this move correct? It was moved from a min-width media query to the 'main' `'lt-ie9` class.

theme/material/stylesheets/application/modules/_content.css.sass
```diff
@@ -60,10 +60,10 @@
         max-height: 1.7em
         margin-right: 1em
 
+        img[src$=".svg"]
```
MKodde: This is not correct right? The previous 'IE9 hide hack' was applied on the images with svg extention on the `header`. They are now selected on `img` elements in the `img` element.

theme/material/stylesheets/application/base/_layout.css.sass
pablothedude: But I don't know in the first place what this should do. Because the `lt-ie*` classes are on the html and I couldn't find `lt-ie*`-elements in the `l-container`. Also media queries will not work in ie8. Do you have any idea?

theme/material/stylesheets/application/modules/_content.css.sass
pablothedude: Good catch, should be moved a bit higher indeed.

PR #671, opened 2019-04-10
Add a surf board to the error page

pablothedude:
This commit will add the new background for the new error page.

I've omitted the responsiveness so this could be addressed later on at once.


tvdijen:
SURFism?

pablothedude:
Hi @tvdijen, the surfism is known, we will take theming into account but that's still work in progress.

PR #673, opened 2019-04-10
Make the error page responsive for small viewports

MKodde:
The different commits should provide you with context about the changes undertaken.

~~In addition to making the page responsive, I've added a watch styles grunt task. Which will be removed with the introduction of #672 I gues~~ I dropped that commit from this PR

This PR covers task 3 of this [Pivotal story](https://www.pivotaltracker.com/story/show/164121204)

### DIFF
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -159,7 +159,7 @@ TXT
     'slidein_read_more' => 'Read more',
 
     // Error screens
-    'error_feedback_info_intro' => '<span class="first-line">Does this error message recur?</span> Then use the error feedback codes listed below when contacting the help desk or e-mail. Please state the codes below:',
+    'error_feedback_info_intro' => '<span class="heading@small">Does this error message recur?</span> Then use the error feedback codes listed below when contacting the help desk or e-mail. Please state the codes below:',
     'error_wiki-href' => 'https://nl.wikipedia.org/wiki/SURFnet',
     'error_wiki-link-text' => '%suiteName% Wiki',
     'error_wiki-link-text-short' => 'Wiki',
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -162,10 +162,13 @@ TXT
     'error_feedback_info_intro' => 'Does this error message recur? Then use the error feedback codes listed below when contacting the help desk or e-mail. Please state the codes below:',
     'error_wiki-href' => 'https://nl.wikipedia.org/wiki/SURFnet',
     'error_wiki-link-text' => '%suiteName% Wiki',
+    'error_wiki-link-text-short' => 'Wiki',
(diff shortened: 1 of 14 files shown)


MKodde:
Screenshot from 2019-04-10 14-32-29

languages/messages.en.php
```diff
@@ -159,13 +159,16 @@
     'slidein_read_more' => 'Read more',
 
     // Error screens
-    'error_feedback_info_intro' => 'Does this error message recur? Then use the error feedback codes listed below when contacting the help desk or e-mail. Please state the codes below:',
+    'error_feedback_info_intro' => '<span class="first-line">Does this error message recur?</span> Then use the error feedback codes listed below when contacting the help desk or e-mail. Please state the codes below:',
```
pablothedude: Could the classname be  more descriptive in order to know what `first-line` actually does? Or remove it completely because it looks currently not used at all.

MKodde: It is used to show the first line in block mode on smaller view ports, this is one of the design alterations for smaller devices. 

Granted, the name is not that descriptive. I'll change it to `heading@small`

MKodde:
> Furthermore I think the background could be optimized a bit more

The background color is the same that is used throughout the rest of the OC suite. You might be onto something here, feel free to discuss it in the Pivotal story.

> As a side note, I personally would like a bit bigger commits for the style adjustments, so the commits would be more useful over time.

Yes, all the small commits might clutter the history slightly, but keep related changes closely together. Making them more useful over time in my opinion. Personally, when I'm inspecting the code, it's from a code perspective, I later check the commit and PR that corresponds to that change. 

Please elaborate on how you'd like to see this kind of changes in version control.

languages/messages.en.php
pablothedude: :+1:

pablothedude:
> > Furthermore I think the background could be optimized a bit more
> 
> The background color is the same that is used throughout the rest of the OC suite. You might be onto something here, feel free to discuss it in the Pivotal story.
> 
It's not the background color, but the alignment of the surfboard, I think this could be a bit better, escpecially jiust before it goes into small, but I think it could be better done in the end. After the important data is responsive.

> > As a side note, I personally would like a bit bigger commits for the style adjustments, so the commits would be more useful over time.
> 
> Yes, all the small commits might clutter the history slightly, but keep related changes closely together. Making them more useful over time in my opinion. Personally, when I'm inspecting the code, it's from a code perspective, I later check the commit and PR that corresponds to that change.
> 
> Please elaborate on how you'd like to see this kind of changes in version control.

I don't mind if it's a single commit for responsiveness, and in the commit message the things done in order to make it happen. But if it don't work for you, then don't.

PR #674, opened 2019-04-12
Add optimized theming support

pablothedude:
This PR will add:
- theming support for both js and css
- a release task

To test it you should do the following:
1. Run: `cp -R material/stylesheets/theme/surfconext material/stylesheets/theme/minbzk`
2. Run  `cp -R material/javascripts/theme/surfconext material/javascripts/theme/minbzk`
3. Edit the application.js and the _theme.sass to validate the ability to override things
4. Run `EB_THEME=minbzk npm run release` to build and test

The `EB_THEME` env var will work for all npm script tasks.

@MKodde if you have reviewed this I'll write the documentation

https://www.pivotaltracker.com/story/show/165240377


pablothedude:
> **Code review**
> 
> 1 > While viewing the changes in package.json to the build tasks I started wondering.
> 
>     1. I'd like to include the additional lint related tasks to the main QA tool chain. How do you feel about adding the two new lint tools + `npm audit` to the build.xml?
> 
Personally I would like to drop ant, and use the scripts section in composer.

>     2. Running the tests before starting a build might be a redundancy as we should already have a green build from the main QA tools. But as long as these test do not take too long, this is a very valuable additional safety guard
> 
I did that indeed for safety. I think we could decide later on if we should drop this. Feel free to do so if it proves not valuable.

> 
> 2 > Second, a suggestion to remove a newly introduced SURFism. The surfboard themeing is now put into a `surfconext` folder. This should either be omitted entirily from VSC and added to the SURF ansible build scripts. Or be kept in VSC under the openconext moniker. I personally opt for the latter option. As the surfboard styling is probably suitable for most users of the OpenConext software suite. Please also discuss this with @thijskh.
>
I'll do that.

> 3 > The directory structure is getting out of hand in my opinion. Some things that I strongly feel about that should be fixed are:
> 
>     1. The original material theme folder should be removed from the project entirily. Resulting in: `/theme/javascripts/theme/surfconext/application.js` versus `/theme/material/javascripts/theme/surfconext/application.js`
> 
I wanted to do that as well, but I didn't want to do structural changes at the moment if I didn't need to in order to prevent nasty merge conflicts.

>     2. The `theme` folder in the `javascripts` and `stylesheets` folders are more theme-overrides than a theme in themselves. I suggest renaming them to `overrides` or some other, more descriptive name.
>
I think generally a theme will be an override and I found the name `theme` more descriptive because currently this word is also used for theming.

>     3. The stylesheets overrides have an interesting folder override structure. The second `theme` folder in the `error-page` folder, is that realy required? (The second theme folder should be emphasized with bold text) `/theme/material/stylesheets/theme/surfconext/error-page/**theme**/_background.sass`
> 
Good catch, that was indeed a leftover of a test in development. Refactored that.

> 
> 4 > After the weekend somebody added a high risk vulnerability in one of the packages we use. Consider fixing this in this PR.
> 
> ```
> $ npm i
> added 343 packages from 319 contributors, removed 1 package, updated 1 package and audited 863704 packages in 94.212s
> found 1 high severity vulnerability
>   run `npm audit fix` to fix them, or `npm audit` for details
> ```
> 
> This concludes my code review, I will now test these changes functionally.

pablothedude:
@MKodde here the separate PR to fix the tar issue:
https://github.com/OpenConext/OpenConext-engineblock/pull/675

MKodde:
I've tested this PR functionally real quick. Your instructions seem to get the job done nicely. For improved user experience, it would be nice to be able to configure your desired theme somewhere other than in the package.json. This prevents users from overriding their custom theming if ever they forget to prepend the `EB_THEME` env var. Other than that, this already works better than the previous theming solution ever did :+1:

pablothedude:
I'll merge this back and ask during the standup what the default behavior should be for the `EB_THEME`. Then I'll create a new PR with these changes if necessary.

PR #675, opened 2019-04-15
Fix tar vulnerability

pablothedude:
In order to fix a vulnerability the tar package is locked at version
4.4.2. If this is fixed in node-gyp we could drop this line. The
node-gyp dependency is used in node-sass.


MKodde:
Lets add a reminder to see if we can undo the resolution in the near future.

pablothedude:
Done: https://www.pivotaltracker.com/story/show/165408610

PR #677, opened 2019-04-16
Add test endpoint for exceptions

pablothedude:
Added a test endpoint for exceptions.

https://www.pivotaltracker.com/story/show/165593518

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
@@ -55,6 +55,7 @@ class FeedbackController extends Controller
     {
         $key = $this->getTemplate($request);
         $feedbackInfo = $this->getFeedbackInfo($request);
+        $parameters = $this->getTemplateParameters($request);
 
         $session = $request->getSession();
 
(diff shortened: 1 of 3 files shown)


MKodde:
After my review, I've played around with https://github.com/americanexpress/jest-image-snapshot. A very simple to use image snapshotting utility that could be used in our test strategy. The catch is that we need to verify and possibly update the base snapshot images religiously with every PR that touches the UI. Are we ready for this? And is it worth the effort? Maybe not on the error pages, but on the WAYF/Consent screens, it might be.

pablothedude:
Personally I don't like snapshot testing because it only tests the markup. Not the markup, styling and scripting all together, what we are already doing with puppeteer/selenium. It is also prone to false positives and false negatives. The UI itself isn't static so dynamic content couldn't be really well tested. And you probably have to resolve a lot of merge conflicts when working on the same code because of the snapshots.

FYI: Jest has also snapshot testing integrated into Jest itself:
https://jestjs.io/docs/en/snapshot-testing.html

MKodde:
The beauty of `jest-image-snapshot` by AM is that  (and I quote)

> jest-image-snapshot is a Jest matcher that performs image comparisons using pixelmatch and behaves just like Jest snapshots do!

So it compares actual pixels and not just the markup, making it, in my opinion, more valuable.

pablothedude:
Feel free to add it, but please don't add the images themselves to VCS to prevent merge conflicts etc. during heavy development. I think if we're almost done it would be very valuable to detect regressions.

PR #679, opened 2019-04-17
Make the metadata push action more efficient

pablothedude:
The metadat push needed to be more effiecient because the memory
got exhausted when pushing metedata. Therefore the ORM layer was
dropped and only DBAL was used.

Please merge https://github.com/OpenConext/OpenConext-engineblock/pull/680 back first


pablothedude:
https://www.pivotaltracker.com/story/show/164329713

thijskh:
LGTM. Probably good to test with some production data

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
```diff
@@ -0,0 +1,269 @@
...
+ */
+class DoctrineMetadataPushRepository
+{
+    /**
+     * @var Connection
+     */
+    private $connection;
+
+    /**
+     * @var \Doctrine\ORM\Mapping\ClassMetadata
```
MKodde: This class is imported, please remove the full namespace (please do so for all occurrences)

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
```diff
@@ -0,0 +1,269 @@
+<?php
```
MKodde: Please:
1. Add the Apache copyright block
2. Remove the PHPstorm generated consturctor doc block

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
```diff
@@ -0,0 +1,269 @@
...
+     *
+     * This method is optimised in order to prevent runnning into memory limits.
+     * Therefore it doesn't use the ORM layer and only the DBAL layer is used.
+     * The entitymanager is only used to get the metadata of the entities.
+     *
+     * @param AbstractRole[] $roles
+     * @return SynchronizationResult
+     * @throws \Exception
+     */
+    public function synchronize(array $roles)
```
MKodde: The logic you moved from the `DoctrineMetadataRepository` can be removed right? Or is there an intentional fallback in place that I do not know about?

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
```diff
@@ -0,0 +1,269 @@
...
+
+    /**
+     * Synchronize the database with the provided roles.
+     *
+     * Any roles (idp or sp) already existing the database are updated. New
+     * roles are created. All identity- or service providers in the database
+     * which are NOT in the provided roles are deleted at the end of the
+     * synchronization process.
+     *
+     * This method is optimised in order to prevent runnning into memory limits.
```
MKodde: Typos in the second paragraph of the doc block:

~~runnning~~ running
~~entitymanager~~ entity manager

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
```diff
@@ -0,0 +1,269 @@
...
+    private $idpMetadata;
+
+    const FIELD_VALUE = 0;
+    const FIELD_TYPE = 1;
+
+    public function __construct(
+        EntityManager $entityManager,
+        Connection $connection
+    ) {
+        $this->connection = $connection;
```
MKodde: There is no direct need to inject the connection as it can be retrieved from the EntityManager. But only if this is the same connection, which I assume it is.

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/ConnectionsControllerTest.php
```diff
@@ -123,6 +125,65 @@ public function cannot_push_invalid_content_to_the_metadata_push_api($invalidJso
...
     }
 
+    /**
+     * @test
+     * @group Api
+     * @group Connections
+     * @group MetadataPush
+     *
+     */
+    public function pushing_data_to_engineblock_should_succeed()
```
MKodde: Adding this test is a very valuable addition. Was adding unit tests for the `DoctrineMetadataPushRepository` not an option? That repo is very heavy on business logic, and would be a top candidate for unit testing! This would entail a big deal of refactoring in order to be able to unit test the synchronize method amply.

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    private function findAllIdentityProviderEntityIds()
+    {
+        $query = $this->connection->createQueryBuilder()
+            ->select('id, entity_id')
+            ->from('sso_provider_roles_eb5')
+            ->where('type="idp"');
+
+        $result = $query->execute();
```
MKodde: Why are you not applying the composite filters? The counterpart of this method does this explicitly. This might cause side effects

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
pablothedude: done!

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
pablothedude: done!

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
pablothedude: I did restore it to be able to test before and after, but indeed that could be removed. Done!

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
pablothedude: I thought I was going to need that in order to get the functional tests up-and-running. But that indeed did not seem the case.

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
pablothedude: The append and visitor filters seemed to be only used by Corto so they weren't used by the push logic at all.

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/ConnectionsControllerTest.php
pablothedude: I thought about that but I found a unit test not that valuable and an acceptance test proved more valuable in order to validate if the roles that you want to get synchronized are synchronized in the end and it was very helpful in order to refactor the code.

src/OpenConext/EngineBlock/Metadata/MetadataRepository/DoctrineMetadataPushRepository.php
MKodde: Thanks for verifying that!

MKodde:
Thanks for fixing the feedback! Can you elaborate on my question in: https://github.com/OpenConext/OpenConext-engineblock/pull/679#pullrequestreview-229435116?

pablothedude:
Oh sorry, missed that.

We discussed IRL the need of addressing the time limit but that should be out-of-scope for now.

PR #680, opened 2019-04-19
Make the memory limit for push configurable

pablothedude:
The metadata memory limit needs to be configurable. This is done so
engineblock could be quickly configured to allow more data being
pushed.

https://www.pivotaltracker.com/n/projects/1425900/stories/164329554
https://github.com/OpenConext/OpenConext-deploy/pull/199

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -182,6 +182,8 @@ attributeAggregation.password = "secret"
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;; MISCELLANEOUS SETTINGS ;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; The memory limit used for the metadata push this setting is overridden in the ConnectionsController
+engineblock.metadata_push_memory_limit = "256M"
 
 ; Minimum execution time in milliseconds when a received response is deemed invalid (default: 5000 ms)
 minimumExecutionTimeOnInvalidReceivedResponse = 5000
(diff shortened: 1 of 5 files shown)


application/configs/application.ini
```diff
@@ -182,6 +182,8 @@ attributeAggregation.password = "secret"
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;; MISCELLANEOUS SETTINGS ;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; The memory limit used for the metadata push this setting is overridden in the ConnectionsController
+metadata_push_memory_limit = "256M"
```
MKodde: As this is an EB setting, you could have added the `engineblock.` prefix. But I would not lose any sleep over you not adding it. The ini file is rather polluted with different conventions.

src/OpenConext/EngineBlockBundle/Controller/Api/ConnectionsController.php
```diff
@@ -74,7 +82,9 @@ public function pushConnectionsAction(Request $request)
             );
         }
 
-        ini_set('memory_limit', '256M');
+        if ($this->memoryLimit) {
```
MKodde: Works like a charm. My initial thought was, is it wise to reset the memory limit after running the memory intensive action. After IRL discussion, we concluded that both the assemble and the sync actions consume a lot of memory. So resetting the memory limit is not yielding any benefits.

application/configs/application.ini
pablothedude: Good suggestion, added the prefix

pablothedude:
The removal of the minified assets are currently on master so they will be in 5.10, Ive changed the upgrade.md accordingly.

MKodde:
> Did you consider updating the UPGRADING.md (or other doc)? It might be a nice place to advertise this new config feature. Other OC users might have similar problems. While your at it, please fix the `5.8 -> 5.9` title typo. This should be `5.8 -> 5.11`

The proposal to add this change in the UPGRADING.md still applies, the second part of this is fixed in #681 I guess :)

pablothedude:
I did changed it, discussed IRL and now ready to merge back.

PR #681, opened 2019-04-23
Add minified assets for release 5.10

pablothedude:
This is done so the assets could be removed in 5.11 which will be
a release with more related features.


pablothedude:

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ The 5.10 release focuses on making EngineBlock more robust. The biggest aim is t
  * Allow re-posting of SAML Responses, allowing page refresh on Consent screen #648
  * Allow for more than one in-flight authentication session per user session #649
  * Added further support for Portuguese language #650 (thanks @domgon!)
- * Update release script and  remove minified assets from VCS #666
+ * Update release script and  remove minified assets from VCS #666 (minified assets were added so they could be dropped in 5.11 #681)
```
MKodde: This sentence is very confusing. Something like this might be more concise:

> * The release script was updated and minified assets have been removed from VCS #666
> * Reverted changes from #666: Added minified assets to VCS. #681

pablothedude: But not all changes from #666 were reverted only the removal of minified assets. I changed the suggested text accordingly.

MKodde: Exactly, that's what I try to communicate in my proposal, but this could be improved upon: 

> * The release script was updated and minified assets have been removed from VCS #666
> * Reverted some changes from #666: Added minified assets to VCS. #681

thijskh: Is it necessary to have both in the changelog? An alternative is to scratch both of these points. Because for the user this is effectively a no-op.

MKodde: Well, we kept some changes from the 666 PR, but reverted the removal of the minified resources. But I agree with you that this information is not that important to put into the changelog.

pablothedude: Changed, so only the real change will be mentioned: `Update release script #666 #681`

PR #683, opened 2019-04-24
Cleanup rolling update 5.8 compatibility support

pablothedude:
Drop backwards compatibilty support for the following issues:


UPGRADING.md
```diff
@@ -5,6 +5,15 @@
 ### Metadata push memory configurable
 The memory used in php for the metadata push is now configurable with the `engineblock.metadata_push_memory_limit` configuration option.
 
+### Rolling update column cleanup
+The features from 5.8.3 to suport rolling updates could be dropped because they are no longer used.
```
MKodde: Grammatically this is not a very well running sentence. I propose:

> The rolling update support features for release 5.8.3 have been dropped as they are no longer used.

UPGRADING.md
```diff
@@ -5,6 +5,15 @@
...
+Running the migrations `Version20180910134145` and `Version20180910175453` will take care of this.
+Be aware that the logic used to update the columns are also dropped in this release.
```
MKodde: What exactly are you warning for? Should the reader take action? Is he losing important data?

UPGRADING.md
```diff
@@ -5,6 +5,15 @@
...
+
+The features removed:
```
MKodde: This, for me, was very confusing. We are not removing these features, we are merely removing the way these features where triggered. As in, we dropped the database columns that contained a flag/setting. Please make this more clear in this very valuable piece of information.

UPGRADING.md
```diff
@@ -5,6 +5,15 @@
 ### Metadata push memory configurable
 The memory used in php for the metadata push is now configurable with the `engineblock.metadata_push_memory_limit` configuration option.
 
+### Rolling update column cleanup
+The features from 5.8.3 to suport rolling updates could be dropped because they are no longer used.
+Running the migrations `Version20180910134145` and `Version20180910175453` will take care of this.
```
MKodde: I don't recall why this should be two migrations? Can you remind me? If there is no reason for this, please merge them into one.

UPGRADING.md
pablothedude: That depends on how the release is deployed and if the migrations are migrated first and after that the code is updated or vice versa. Maybe we should have an intermediate release to support rolling updates?

UPGRADING.md
pablothedude: Made it more clear

UPGRADING.md
pablothedude: Because it were originally two separate features. I would like to keep it split up by feature for future reference.

UPGRADING.md
pablothedude: I changed it to the sentence below to have the best of both worlds.
`The features from 5.8.3 to support rolling updates have been dropped as they are no longer used.`

UPGRADING.md
MKodde: Yes you did! 

I still think mentioning the PRs where the original feature was added is rather confusing. A casual reader can now interpret this as if those features are dropped. But they are not, only an implementation detail of those features have changed. My suggestion is to either remove the list, or to mention what specific part of that feature is dropped.

UPGRADING.md
MKodde: As you suggested IRL, let's discuss this tomorrow at the daily stand up.

UPGRADING.md
pablothedude: Removed the list.

PR #668, opened 2019-04-09
Add the session exception translations

pablothedude:
The translations were discussed so now we have the latest version to
show the enduser.

https://www.pivotaltracker.com/story/show/164076545

PR #684, opened 2019-04-24
Remove ul from session exception messages

pablothedude:
The ul needed to be removed from the session exception message.


MKodde:
As per: https://github.com/OpenConext/OpenConext-engineblock/pull/668#discussion_r277990960

PR #686, opened 2019-04-25
Update the browser support

pablothedude:
The current browsers we are supporting are updated.

https://www.pivotaltracker.com/story/show/165377239

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,20 @@ If you are using this pattern, an update can be done with the following:
 
 5. Change the symlink.
 
+## Browsers support
+
+The list of browsers that should be supported: 
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -202,6 +202,15 @@ If you are using this pattern, an update can be done with the following:
 
 5. Change the symlink.
 
+## Browsers support
+
+The list of browsers that should be supported:
```
MKodde: Why use the _should_ modal verb? Should is considered a suggestion and is the weakest of the available modal verbs.

Did you use this intentional, as in, we aim to provide the following browser support? 

I suggest changing this into something like:

> The following browsers are supported:

README.md
```diff
@@ -202,6 +202,15 @@ If you are using this pattern, an update can be done with the following:
...
...irefox_48x48.png" alt="Firefox" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Firefox | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/chrome/chrome_48x48.png" alt="Chrome" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Chrome | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/safari/safari_48x48.png" alt="Safari" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Safari | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/safari-ios/safari-ios_48x48.png" alt="iOS Safari" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>iOS Safari | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/samsung-internet/samsung-internet_48x48.png" alt="Samsung" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Samsung | [<img src="https://raw.githubusercontent.com/alrra/browser-logos/master/src/opera/opera_48x48.png" alt="Opera" width="24px" height="24px" />](http://godban.github.io/browsers-support-badges/)</br>Opera |
```
MKodde: Do we actually support, and more importantly, test Safari, Samsung, and Opera? And a nice addition might be to add the platform we test the specified browsers on.

README.md
pablothedude: It was intentional because it's on a best effort basis and I didn't want to make a commitment to test them all. I thought this could also be an incentive for users to do a contribution themselves.

README.md
pablothedude: See my answer above. Mentioning the platform of the browser is in my opinion not that valuable.

README.md
MKodde: @thijskh what's your opinion on this?

README.md
MKodde: Again @thijskh can you give your opinion on this?

pablothedude:
I did revert the package-lock.json

README.md
thijskh: Best in my opinion is to just make explicit what we do. I think our 'browser support' is split into two aspects. So we can both mention for what browser support we configure our toolchain (as Bas did) and additionally document with which browsers we perform manual tests (that is: IE11, latest FF, latest Chrome, latest Edge, latest Safari, latest edge, latest iPhone, latest Android)

pablothedude: @thijskh thanks, I'll update the documentation accordingly.

MKodde: One last suggestion: Maybe instead of the sentence: 

> The list of browsers that are currently being tested:

Use:

> The list of browsers that are tested before every release:

In my opinion, using: 'that are currently being tested', implies that we are testing them right now.

MKodde: Oh, and maybe add Android -> Chrome to the list of browsers we 'should' support. As we explicitly verify that combination, and you also mentioned IoS Safari.

thijskh: I've asked our testers and it seems that with "Android" they use "a" browser on these platforms and it's not necessarily the same every time

pablothedude: @thijskh do you want me to remove the `Chrome Android` and `iOS Safari` entries from the list of tested browsers?

thijskh: No, we’ll change the testing procedures to match

pablothedude: :+1:

PR #687, opened 2019-04-25
Implement visual regression tests for error pages

MKodde:
Different error pages are visited and compared to previous snapshots
when a screenshot differs from a previous one, the test will fail. This
test should be run manually for now.

See task 3 of https://www.pivotaltracker.com/story/show/165593518

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -168,6 +168,20 @@
         </exec>
     </target>
 
+    <target name="js-visual-regression-tests">
+        <exec dir="${basedir}/theme" executable="node_modules/.bin/jest" failonerror="true">
+            <arg line="--runInBand" />
+            <arg line="material/javascripts/tests/visual-regression" />
+        </exec>
+    </target>
+
+    <target name="js-visual-regression-tests-update-snapshots">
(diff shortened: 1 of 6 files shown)


pablothedude:
Great work, this is going to be very valuable to test for regressions later on, maybe consider to enable this after we have finished the 5.11 release?

1. In the initial run I got a timeout... maybe the timeout could be set less strict.
1. In the second run the diff doesn't seem to work because the loop over the viewports in the same test and a race condition in the viewports diffs the image with the wrong size so it fails. I would suggest a test per viewport and error page  and not only per error page or a custom identifier to keep the names a bit shorter. This in order to quickly see which dimension is broken even if we support more dimensions.
1. Also I would like to have a fullage snapshot to also compare the hidden areas for small viewports.

MKodde:
Thanks, 

1. Upped the timeouts, hopefully, this resolves your issues
2. Fixed
3. Fixed, thanks for the fullpage tip.

As a bonus I looked into also adding locale support (rendering all pages in NL and EN) but could not manipulate the lang cookie. Might be a rookie mistake on my end, but postponing that for now.

PR #688, opened 2019-04-25
Improve error feedback blocks

MKodde:
In this PR, the error feedback blocks are optimized and visual quirks are taken care of.

Please see the individual commits for details on the (many) changes.


pablothedude:
I've tested it funcitonally and it seems to work as expected in FF and Chrome. IN IE10 and 11  the boxes are always below each other and seem to miss padding/spacing. Was that intentional?

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
```diff
@@ -89,18 +89,29 @@ private function getTemplate(Request $request)
...
-        $default = '{
-            "timestamp":"2019-04-15T19:21:06+02:00",
-            "requestId":"5cb4bd3879b49",
-            "userAgent":"Mozilla\/5.0 (X11; Ubuntu; Linux x86_64; rv:66.0) Gecko\/20100101 Firefox\/66.0",
-            "ipAddress":"192.168.66.98",
-            "artCode":"31914"
-        }';
-
-        $feedbackInfo = $request->get('feedback-info', $default);
+        $feedbackInfo = [
```
pablothedude: These changes will probably give you some merge conflicts.

src/OpenConext/EngineBlockBundle/Value/FeedbackInformation.php
```diff
@@ -0,0 +1,54 @@
...
+        $this->key = $key;
+        $this->value = $value;
+    }
+
+    public function getKey()
+    {
+        return $this->key;
+    }
+
+    public function getAttrSafeKey()
```
pablothedude: Couldn't the keys themselves not be made safe, so there is no additional logic needed to make this possible?

src/OpenConext/EngineBlockBundle/Value/FeedbackInformationMap.php
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlockBundle\Value;
+
+use OpenConext\EngineBlockBundle\Exception\RuntimeException;
+
+class FeedbackInformationMap
+{
+    private $data = [];
```
pablothedude: Please typehint

src/OpenConext/EngineBlockBundle/Value/FeedbackInformation.php
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlockBundle\Value;
+
+use OpenConext\EngineBlock\Assert\Assertion;
+
+class FeedbackInformation
+{
+    private $key;
```
pablothedude: Please typehint

src/OpenConext/EngineBlockBundle/Value/FeedbackInformation.php
```diff
@@ -0,0 +1,54 @@
...
+
+    private $value;
```
pablothedude: Please typehint

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
MKodde: Thanks for the heads-up!

src/OpenConext/EngineBlockBundle/Value/FeedbackInformation.php
MKodde: This is just a defense in depth thingy, granted this could be fixed at another place, for instance in the twig template or at the place where we configure the keys. This seemed the most convenient place. I've added a doc block with information.

MKodde:
> I've tested it funcitonally and it seems to work as expected in FF and Chrome. IN IE10 and 11 the boxes are always below each other and seem to miss padding/spacing. Was that intentional?

To be honest, I have not yet tested the changes in IE. I'm going to do this now. Hopefully, I won't run into too many IE quirks here.. :crossed_fingers:

MKodde:
Getting the blocks to render perfectly in IE10/11 was not realistic as the 'dense' `grid-auto-flow` is not supported in IE10/11. The autoprefixer did not do the trick sadly..

So as a solution I propose to use good old block layout in those browsers. Resulting in the following (see screenshot)

Screenshot from 2019-05-01 14-01-06

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
MKodde: I've taken your concern into account by removing the parameters and dropping the commit where the feedback controller was updated. Alternatively, I added the sp/idp parameters to the feedback-info json.

arnoutt:
> Getting the blocks to render perfectly in IE10/11 was not realistic as the 'dense' `grid-auto-flow` is not supported in IE10/11. The autoprefixer did not do the trick sadly..
> 
> So as a solution I propose to use good old block layout in those browsers. Resulting in the following (see screenshot)
> 
> Screenshot from 2019-05-01 14-01-06
> 
> @arnoutt is this acceptable? Or should we invest more effort into getting things the same between modern and older (IE) browsers?

Yes this is acceptable.

MKodde:
Merging :cat2:

PR #689, opened 2019-04-26
Add configurable theming

pablothedude:
Set 'openconext' as the default theming and remove Surfisms from the feedback page background images. These images could be changed with help of Ansible. See : https://github.com/OpenConext/OpenConext-deploy/pull/203

https://www.pivotaltracker.com/story/show/165240377


theme/material/templates/modules/Authentication/View/Proxy/form.html.twig
```diff
@@ -10,8 +10,6 @@
         <meta content="initial-scale=1.0,user-scalable=no,maximum-scale=1,width=device-width" name="viewport">
         <meta content="translucent-black" name="apple-mobile-web-app-status-bar-style">
         <link href="/favicon.ico" rel="shortcut icon">
-        <!--[if lte IE 8]>
```
MKodde: Thanks for removing this. I suppose `html5shiv.min.js` was missing for quite some time?

theme/package.json
```diff
@@ -11,17 +11,19 @@
...
     "lint:js": "jshint material/javascripts/",
     "lint:css": "stylelint 'material/stylesheets/**/*.sass' --syntax sass",
-    "build": "npm run build:clean && npm run build:js && npm run build:css && npm run build:copy",
-    "build:clean": "rimraf ../web/images/* ../web/javascripts/* ../web/stylesheets/*",
-    "build:js": "mkdir -p ../web/javascripts && npm run build:js:babelify",
-    "build:js:babelify": "browserify material/javascripts/application.js material/javascripts/theme/${EB_THEME:=surfconext}/application.js --plugin [ tinyify ] -t [ babelify ] -o ../web/javascripts/application.min.js",
-    "build:css": "mkdir -p ../web/stylesheets && npm run build:css:sass && npm run build:css:postcss",
-    "build:css:sass": "node-sass material/stylesheets/ --include-path=material/stylesheets/theme/${EB_THEME:=surfconext} --error-bell -o ../web/stylesheets/",
-    "build:css:postcss": "BROWSERSLIST_CONFIG=.browserslistrc postcss ../web/stylesheets/*.css --verbose --use autoprefixer --use cssnano --no-map --dir ../web/stylesheets/",
+    "clean": "rimraf ../web/images/* ../web/javascripts/* ../web/stylesheets/*",
```
MKodde: The addition of the `theme:` tasks is very confusing to me. For example when do I use `npm run clean` versus `npm build:theme:clean`?

build.xml
```diff
@@ -6,10 +6,11 @@
     <target name="functional-tests" depends="api-functional-tests,behat-travis" />
     <target name="functional-tests-wip" depends="behat-travis-wip" />
     <target name="functional-tests-selenium" depends="behat-travis-selenium"/>
-    <target name="code-quality-ci" depends="test-suites,code-quality"/>
+    <target name="code-quality-ci" depends="test-suites,code-quality,npm-lint"/>
     <target name="code-quality" depends="php-lint,phpmd,phpcs,phpcs-legacy"/>
-    <target name="security-tests" depends="security-checker" />
+    <target name="security-tests" depends="security-checker,npm-audit" />
     <target name="test-suites" depends="eb4-tests,unit-tests,integration-tests,js-unit-tests,js-smoke-tests"/>
+    <target name="validate" depends="composer-validate,npm-audit"/>
```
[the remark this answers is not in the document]
pablothedude: The `npm-audit` indeed didn't supposed to be there, removed it.

theme/material/templates/modules/Authentication/View/Proxy/form.html.twig
pablothedude: It looked like it was copied because I couldn't find any trace it was there before and it seemed to be only on the form template and for example not on the default template.

theme/package.json
pablothedude: To give you some context. It was previously possible to build a `openconext` and a `surfconext` theme so we could easily test them both.  With `build` all themes were build and with `build:theme` a specified theme was build so with `build:theme:clean` only the asests for a specific theme was cleaned and `clean` cleaned all assets from all themes. I've removed it now because it isn't possible anymore.

theme/material/templates/modules/Authentication/View/Proxy/form.html.twig
pablothedude: > Nice work! See my feedback below.
> 
> **Questions:**
> 
>     1. What's the point of adding the empty background images? They are dropped into the installation during deployment, right?
> 
As default they are kept during deployment. But only if they are set they are replaced. This because a requirement was to only replace images and not build multiple themes for `openconext` and `surfconext`. So as default there is a theme without Surfisms and those Surfisms are added only by replacing images. I've added them to prevent extensive 404's in the logging.

>     2. The openconext theme seems to be the key to installing the SURFconext specific background images. You removed the surfismns, as in, you renamed custom SURF themeing to OpenConext. Would it not make more sense to configure OC-deploy to install the custom theme into EB? Or at least to make the default configuration to not use the `openconext` theme, which actually is the SURFconext theme. I'm confused here. Can you explain? Possibly by updating the theme portion of the readme.
> 
The `openconext` theme is the default and we are using that theme for Surfconext. The requirement was not to build a own theme. My personal preference was also to build the two themes and only set the theme in OC-deploy, mainly because we could test them both. But only replacing the images to add the Surfisms was a hard requirement.

> 
> **Remark:**
> 
>     1. I still think that the material folder should be removed or at the very least be renamed. Material used to be the name of our previous theme. It overcomplicates the folder structure.

It's still open and would like to do it on the end to prevent merge conflicts, (As a side note personally I wouldn't mind keep using the folder `material` because it was maybe intended as theme name but it functionally  wasn't and the name `material` seems to describe what the contents are):
https://www.pivotaltracker.com/story/show/165240377

theme/package.json
MKodde: Perfect

PR #690, opened 2019-04-26
Add the possibility to sign a SAML response

pablothedude:
An option is needed to sign the saml response if the option
`metadata:coin:sign_response` is set. This because some sp's require
the saml response to be signed and not only the assertions.

### DIFF
--- a/database/DoctrineMigrations/Version20190425205743.php
+++ b/database/DoctrineMigrations/Version20190425205743.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace OpenConext\EngineBlock\Doctrine\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20190425205743 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 ADD sign_response TINYINT(1) DEFAULT NULL');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 DROP sign_response');
+    }
+}
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -583,8 +583,16 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
                     $assertion->setCertificates($sspMessage->getCertificates());
                     $assertion->setSignatureKey($sspMessage->getSignatureKey());
                 }
-                // BWC dictates that we don't sign responses.
-                $messageElement = $sspMessage->toUnsignedXML();
+
+                // According to saml2int 0.2, the assertion element must be directly signed, which we do.
+                // This suffices for most SP's. However, some SP's require the outer Response element to be signed directly.
+                // If configured to do so for that SP, Engineblock will sign the outer response element in addition to the signed Assertion element.
+                // It might make sense in the future to always sign both, if it has been shown that this causes no problems.
+                if ($remoteEntity instanceof ServiceProvider && $remoteEntity->signResponse) {
+                    $messageElement = $sspMessage->toSignedXML();
+                } else {
+                    $messageElement = $sspMessage->toUnsignedXML();
+                }
             }
             else {
                 $messageElement = $sspMessage->toSignedXML();
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
@@ -546,6 +546,32 @@ class MockSpContext extends AbstractSubContext
     }
 
     /**
+     * @Given /^SP "([^"]*)" requires a signed response$/
+     */
+    public function spRequiresASignedResponse($spName)
+    {
+        /** @var MockServiceProvider $mockSp */
+        $mockSp = $this->mockSpRegistry->get($spName);
+
+        $this->serviceRegistryFixture
+            ->setSpSignRepsones($mockSp->entityId())
+            ->save();
+    }
+
+    /**
+     * @Given /^SP "([^"]*)" does not require a signed response$/
+     */
+    public function spDoesNotRequireASignedResponse($spName)
+    {
+        /** @var MockServiceProvider $mockSp */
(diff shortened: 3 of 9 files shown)


pablothedude:
https://www.pivotaltracker.com/story/show/165311614

database/DoctrineMigrations/Version20190425205743.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace OpenConext\EngineBlock\Doctrine\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
```
MKodde: Please remove the auto-generated text and replace it with a small explanation. Or remove the comment altogether.

database/DoctrineMigrations/Version20190425205743.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 ADD sign_repsonse TINYINT(1) DEFAULT NULL');
```
MKodde: The `sign_repsonse` field could be made to be more self-explaining, and cohere better to the naming of the other feature flags, by adding `_required` to the name. Or add a prefix like `must_sign_response` Or choose a better title that makes clear we are dealing with a feature flag here.

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -583,8 +583,13 @@ public function send(
                     $assertion->setCertificates($sspMessage->getCertificates());
                     $assertion->setSignatureKey($sspMessage->getSignatureKey());
                 }
-                // BWC dictates that we don't sign responses.
-                $messageElement = $sspMessage->toUnsignedXML();
+
+                // Sign response if required
+                if ($remoteEntity instanceof ServiceProvider && $remoteEntity->signResponse) {
```
MKodde: Did you run into situations where the remote entity was not an SP?

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -583,8 +583,13 @@ public function send(
...
+                    $messageElement = $sspMessage->toSignedXML();
+                } else {
+                    $messageElement = $sspMessage->toUnsignedXML();
```
MKodde: You removed the `BWC dictates that we don't sign responses` comment. Maybe it's a good idea to still add some comments here explaining why certain SP's do, and why by default, we do not sing the responses.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
```diff
@@ -545,6 +534,32 @@ public function spRequiresAttributeAggregation($spName)
             ->save();
     }
 
+    /**
+     * @Given /^SP "([^"]*)" will require a signed response$/
```
MKodde: I'd like to suggest a change in this step definition:

> SP "Foobar" requires a signed response

Will indicates that the requirement will be stated in the future. And all step definitions are in the present tense.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
```diff
@@ -545,6 +534,32 @@ public function spRequiresAttributeAggregation($spName)
...
+        /** @var MockServiceProvider $mockSp */
+        $mockSp = $this->mockSpRegistry->get($spName);
+
+        $this->serviceRegistryFixture
+            ->setSpSignRepsones($mockSp->entityId())
+            ->save();
+    }
+
+    /**
+     * @Given /^SP "([^"]*)" will not require a signed response$/
```
MKodde: Like my previous comment. ~~will~~ _does_

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Encryption.feature
```diff
@@ -83,3 +83,25 @@ Feature:
       And I give my consent
       And I pass through EngineBlock
      Then the response should contain "urn:mace:terena.org:attribute-def:schacHomeOrganization"
+
+  Scenario: EngineBlock should support not signed responses
```
MKodde: Proposal to make the scenario text more in line with the others:

> EngineBlock supports unsigned responses

And for the scenario below:

> EngineBlock supports signed responses

library/EngineBlock/Corto/Module/Bindings.php
thijskh: not sing :x::woman_singer:

thijskh: Something like "According to saml2int 0.2, the assertion element must be directly signed, which we do. This suffices for most SP's. However, some SP's require the outer Response element to be signed directly. If configured to do so for that SP, Engineblock will sign the outer response element in addition to the signed Assertion element. It might make sense in the future to always sign both, if it has been shown that this causes no problems."

database/DoctrineMigrations/Version20190425205743.php
thijskh: In any case I would use `response` instead of `repsonse` ;-)

pablothedude:
I removed the `theSpUsesHttpPostBinding` method to test if it still was used and ran the tests to see if the method was no longer being used. The tests succeeded so I assumed the method could be safely removed. But it seemed that the tests succeeded even when they were broken. I therefore have added the `--strict` flag to the ant behat script to let the tests fail if the behat tests themselves are not valid.

database/DoctrineMigrations/Version20190425205743.php
pablothedude: Removed, it's not in the way so it doesn't bother me

database/DoctrineMigrations/Version20190425205743.php
pablothedude: I've fixed the `repsonse` typo. I left the name of the property as-is. I'm sharing the concern that the name could be more descriptive, but it seems the standard to convert the coin property names from manage is to lower camel case the names and use them that way. I will however add your concern to the issue in pivotal were we could address the serialization of coin properties.

https://www.pivotaltracker.com/story/show/165670296/comments/202145201

library/EngineBlock/Corto/Module/Bindings.php
pablothedude: I personally did not but the `$remoteEntity` is an `AbstractRole` and I saw for example that `\EngineBlock_Corto_ProxyServer::sendAuthenticationRequest` seems to use this method with a IDP.

library/EngineBlock/Corto/Module/Bindings.php
pablothedude: @thijskh thanks for the supplied documentation, I'm currently singing responses which I normally won't do as default :notes: :smile:

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
pablothedude: Changed

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/MockSpContext.php
pablothedude: Changed

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Encryption.feature
pablothedude: Done

PR #691, opened 2019-04-29
Make footer buttons configurable

pablothedude:
The visibility of the mail support and wiki needs to be configurable.
So the styling needs to be changed accordingly. This will also
allow the options to be tested with the functional testing feedback
controller.

In order to test this you could use the following parameters:
`supportEmail` `string`
`showWikiButton` `bool`

Also the background offset didn't work as expected when the viewport size
was bigger then 100%. This was boyscouted.

[edit]
Also the .gitignore from the themes folder was removed. This was also boyscouted.


pablothedude:
https://www.pivotaltracker.com/story/show/164121204

theme/material/templates/modules/Default/View/Error/partial/footer.html.twig
```diff
@@ -1,27 +1,31 @@
...
                 </a>
             </div>
-            <div class="footer-buttons__link">
-                <a href="mailto:{{ 'error_help-desk-email'|trans }}" target="_blank" class="footer-button">
+            {% if supportEmail is defined %}
+            <div class="footer-button">
+                <a href="mailto:{{ supportEmail }}" target="_blank" class="footer-button__link">
                     <i class="footer-button__icon fa-paper-plane"></i>
-                    <p class="footer-button__text">{{ 'error_help-desk-email-text'|trans }}</p>
+                    <p class="footer-button__text">{{ supportEmail }}</p>
```
MKodde: Possible bug: `supportEmail` is not passed to the template or is not declared within the template.

MKodde:
This might also be a nice place to add usage information about the feedback test page in `docs/js_testing.md`

theme/material/templates/modules/Default/View/Error/partial/footer.html.twig
pablothedude: What do you mean?

pablothedude:
I've created a PR with the tests for the footer:
https://github.com/OpenConext/OpenConext-engineblock/pull/692

I don't want to spend a lot of time on this at this moment. I will write the documentation when we get the results back from @arnoutt regarding the configuration of the different exceptions and the contact address in the footer.

theme/material/templates/modules/Default/View/Error/partial/footer.html.twig
MKodde: I assumed this solution was also applicable for the real feedback controller. But for now, the support mail button can only be rendered from the test endpoint.

theme/material/stylesheets/error-page/modules/_footer.sass
```diff
@@ -38,16 +46,9 @@
       display: none
       +max-width($small)
         display: block
+
   &__icon
     font-family: FontAwesome
     text-align: center
     font-size: $f-large
     font-weight: bold
-  .fa-wikipedia-w
```
MKodde: Why did the wiki styling change? This was in accordance to the design that can be found in the pivotal story.

pablothedude: I removed the footer styling and rebuild it from the ground up to allow the ability to have 1 or multiple buttons on the same row while centering them all and keeping the borders in place. I then forgot to re-add it again, because it looked already beautiful to me :wink:

pablothedude:
Reverted the wiki logo

pablothedude:
I did revert the package.lock it changed because of the tar resolution. Hopefully that gets fixed soon.
Added a note to the description of the PR about the .gitignore.

PR #692, opened 2019-04-30
Add visual tests for the feedback page footer

pablothedude:
In order to detect regressions in the footer of the feedback page
some tests are added.

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
@@ -57,8 +57,6 @@ class FeedbackController extends Controller
         $feedbackInfo = $this->getFeedbackInfo($request);
         $parameters = $this->getTemplateParameters($request);
 
-        $this->validateFooterTemplateParameters($parameters);
-
         $session = $request->getSession();
 
         $template = sprintf(
@@ -106,7 +104,6 @@ class FeedbackController extends Controller
         return $feedbackInfo;
     }
 
-
     /**
      * @param Request $request
      * @return mixed|string
@@ -122,21 +119,4 @@ class FeedbackController extends Controller
         return $parameters;
     }
 
-
-    /**
-     * @param array $parameters
-     * @return null|string
-     */
-    private function validateFooterTemplateParameters(array &$parameters)
-    {
-        $key = 'supportEmail';
-        if (!array_key_exists($key, $parameters)) {
-            $parameters[$key] = null;
-        }
-
-        $key = 'showWikiButton';
-        if (!array_key_exists($key, $parameters)) {
-            $parameters[$key] = false;
-        }
-    }
 }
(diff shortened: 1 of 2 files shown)


theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
```diff
@@ -0,0 +1,58 @@
...
+        for (const footer of footerDifferences) {
+            for (const viewport of viewports) {
+                it(`${footer.name}-${viewport.width}x${viewport.height}`, async () => {
+
+                    await page.goto(footer.url);
+                    await page.setViewport(viewport);
+                    await page.waitFor(".error-container");
+                    const screenshot = await page.screenshot({
+                        fullPage: true,
+                        path: `./material/javascripts/tests/visual-regression/error-page/screenshots/footer/${footer.name}-${viewport.width}x${viewport.height}.png`
```
MKodde: can you add the `footer` folder by adding a .gitkeep file in that dir?

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
```diff
@@ -57,8 +57,6 @@ public function feedbackAction(Request $request)
         $feedbackInfo = $this->getFeedbackInfo($request);
         $parameters = $this->getTemplateParameters($request);
 
-        $this->validateFooterTemplateParameters($parameters);
```
MKodde: Why did you remove the validate method?

theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
pablothedude: Discussed IRL and decided to remove the .gitignore from the theme folder and used the one in the root of the project.

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/FeedbackController.php
pablothedude: Because it wasn't needed after all because of the `is defined` logic in the footer template. So the parameters are not required.

theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
MKodde: Moving the gitignore will not fix this problem, I suspect. The test fails if the folder is not present, my proposal is to add the empty folder with a .gitkeep file to prevent end users from having to create directories to get tests to pass.

MKodde:
Like in #691, is the change to the Package.json a requirement? If so, please elaborate.Otherwise please drop it from the PR.

theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
pablothedude: I lacked to push the change, done.

pablothedude:
Reverted the package-lock.json change

theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
pablothedude: I lacked to push the change, done.

PR #693, opened 2019-05-01
Fix npm installation consistency between builds

pablothedude:
Npm was not consistent between builds and changed the package-lock.json.
Therefore we are now using the 'ci' commmand.

https://docs.npmjs.com/cli/ci.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ env:
 cache:
   directories:
     - ~/.composer/cache/files
+    - "$HOME/.npm"
 
 before_install:
   - mysql -e "create database IF NOT EXISTS eb_test;" -uroot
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -65,7 +65,7 @@ before_script:
   - composer config secure-http false
   - composer install --prefer-dist -n -o
   - app/console cache:clear --env=test
-  - npm install --prefix=theme
+  - npm ci --prefix=theme
```
MKodde: To speed up Travis even more you could add a NPM cache location to the chache directive. As suggested by:

https://docs.npmjs.com/cli/ci.html#example

pablothedude: A great suggestion, done!

PR #696, opened 2019-05-02
Make uncaught error page reloadable

MKodde:
That was previously also possible (but might have leaded to a new error condition), and resulted in a new unique request id, timestamp combination upon reload. Making it very hard for service desk to match the request id with the actual error message in logs.

This is resolved by redirecting to the feedback controller. On this controller a new action handles the 'unknown error'.

This action is intentionally not accessible on the authentication/feedback base route. This because the unknown error might not be authentication related.

See: https://www.pivotaltracker.com/story/show/164076480

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Resources/config/routing.yml
+++ b/src/OpenConext/EngineBlockBundle/Resources/config/routing.yml
@@ -28,7 +28,12 @@ certificate:
 
 feedback:
     resource: routing/feedback.yml
-    prefix:   '/authentication/feedback'
+    prefix: '/authentication/feedback'
+
+feedback_unknown_error:
+    path: /feedback/unknown-error
+    methods: [GET]
+    defaults: { _controller: engineblock.controller.authentication.feedback:unknownErrorAction }
 
 wayf:
     resource: routing/wayf.yml
--- a/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
@@ -58,6 +58,14 @@ class FeedbackController
         );
     }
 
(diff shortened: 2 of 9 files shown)


MKodde:
Removed a `var_dump` debug leftover.

Can you elaborate on: 

>What should happen if the session expires? It doesn't bother if the message changes?

Do you see problems with this? Should this route be added to the 'whitelisted' routes that can be served without session?

src/OpenConext/EngineBlockBundle/Resources/config/routing.yml
```diff
@@ -28,7 +28,12 @@ certificate:
 
 feedback:
     resource: routing/feedback.yml
-    prefix:   '/authentication/feedback'
+    prefix: '/authentication/feedback'
+
+feedback_unknown_error:
```
pablothedude: Should the endpoint itself also be tested?
theme/material/javascripts/tests/visual-regression/error-page/ErrorPage.test.js

pablothedude:
If the session expires and the page gets reloaded after that the requirement below will not be met:
`Pressing reload on an error page should not change the error page, nor create a new error.`

MKodde:
To be clear: @pablothedude is stating the possibility of losing your session (for example due to timeout) while on the generic error page. In this case a page reload would not show you the exact same page. You are still going to be on the generic error page, but the feedback information will no longer be on the page. As that info was stored in the session.

We discussed some possible solutions: 
1. Backup the feedback info in a cookie. And use that as a fallback when the session is lost
2. Use the local storage (if available)
3. Accept this edge case.

@thijskh what is your opinion on this?

thijskh:
Accept it for now. Solve it when it turns out to be a real world problem.

PR #697, opened 2019-05-03
Whitelist debugSingleSignOnService for session

pablothedude:
The debugSingleSignOnService was only added to the configuration
array in the corto proxyserver on master but not in the 5.10 release
branch while this was needed. This was done in order to prevent it from
checking if the session did already start.


pablothedude:
https://www.pivotaltracker.com/story/show/164076545/comments/202263473

pablothedude:
Also cherrypicked this into master.

PR #705, opened 2019-06-04
Move and merge two 5.11 migrations

MKodde:
These two migrations (Version20180910175453.php and Version20180910134145.php), where migrations to be run in 5.10. They have been dropped from that release and moved into the 5.11 release. This required them to be moved to a new migration with a new timestamp in order to allow running the migration without a problem.

Task 3 from https://www.pivotaltracker.com/story/show/166435147


database/DoctrineMigrations/Version20190504110100.php
```diff
@@ -6,19 +6,27 @@
 use Doctrine\DBAL\Schema\Schema;
 
 /**
- * Auto-generated Migration: Please modify to your needs!
+ * This is a combined migration of the previous migration versions:
+ * - Version20180910134145
```
pablothedude: As a suggestion, maybe you could add the initial PR's?

database/DoctrineMigrations/Version20190504110100.php
```diff
@@ -27,8 +35,9 @@ public function up(Schema $schema)
     public function down(Schema $schema)
     {
         // this down() migration is auto-generated, please modify it to your needs
```
pablothedude: How valuable is this? Because you effectively can't go back if the column data is lost.

thijskh: You can: should you downgrade to an older version: then just re push

database/DoctrineMigrations/Version20190504110100.php
MKodde: Yes, good suggestion, added it to the docblock

PR #707, opened 2019-06-05
Move the SURFboard to the left on smaller than large viewports

MKodde:
When the viewport is smaller than large (1024px), then the surfboard is moved to a position of '-8% left' of the viewport boundary. This makes sure the image is well readable when the viewport size is smaller.

The SVG itself (not tracked in this repo) is updated, the sign is moved to the left of the surfboard.

See point 1 of https://www.pivotaltracker.com/story/show/166437858

### DIFF
--- a/theme/material/stylesheets/theme/openconext/error-page/_background.sass
+++ b/theme/material/stylesheets/theme/openconext/error-page/_background.sass
@@ -17,6 +17,9 @@
     position: absolute
     height: 100%
     width: 100%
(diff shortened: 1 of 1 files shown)


MKodde:
The updated SVG is added to this message as a zip archive (GitHub does not allow uploading of SVG images).

[background-front.zip](https://github.com/OpenConext/OpenConext-engineblock/files/3256111/background-front.zip)

PR #717, opened 2019-07-11
Decrease the surfboard offset on the error page

pablothedude:
The surfboard could be better aligned on wide screens. The surfboard
scales better now with the use of css.

https://www.pivotaltracker.com/story/show/167202154

### DIFF
--- a/theme/material/stylesheets/theme/openconext/error-page/_background.sass
+++ b/theme/material/stylesheets/theme/openconext/error-page/_background.sass
@@ -12,15 +12,14 @@
     display: none
 
   &__front
-    background: url(/images/background-front.svg) left bottom no-repeat
+    background: url(/images/background-front.svg) no-repeat left bottom
+    background-size: contain
     bottom: 0
-    left: 10%
+    left: 2%
     position: absolute
-    height: 100%
-    width: 100%
+    height: 40%
+    width: 20%
     z-index: 1
-    +max-width($large)
-      left: -8%
 
   &__back
     background: url(/images/background-back.svg) left bottom repeat-x


theme/material/stylesheets/theme/openconext/error-page/_background.sass
```diff
@@ -12,15 +12,14 @@
...
     bottom: 0
-    left: 10%
+    left: 2%
     position: absolute
-    height: 100%
-    width: 100%
+    height: 40%
+    width: 20%
     z-index: 1
-    +max-width($large)
```
MKodde: This feature was explicitly added in #707 why was it removed in this commit?

theme/material/stylesheets/theme/openconext/error-page/_background.sass
```diff
@@ -12,15 +12,14 @@
...
-    background: url(/images/background-front.svg) left bottom no-repeat
+    background: url(/images/background-front.svg) no-repeat left bottom
+    background-size: contain
     bottom: 0
-    left: 10%
+    left: 2%
     position: absolute
-    height: 100%
-    width: 100%
+    height: 40%
```
MKodde: An interesting solution. By shrinking the surfboard on smaller viewports, more of the actual surfboard might be visible. Is this solution shown to @thijskh and @arnoutt via a screenshot? 

**Update** I've added some screenshots to the Pivotal story

theme/material/stylesheets/theme/openconext/error-page/_background.sass
pablothedude: Because that didn't do the trick on very wide displays.

theme/material/stylesheets/theme/openconext/error-page/_background.sass
pablothedude: I didn;t supply them with screenshots because in order yo be able to see the actual scaling I found that a browser would  give you a better feel about the scaling.

theme/material/stylesheets/theme/openconext/error-page/_background.sass
MKodde: That fix was for smaller displays. But granted, it still sometimes resulted in the surfboard rendering behind the content area. Which by the way is still possible with your solution, but more of the actual surfboard remains in view as it scales down in size.

theme/material/stylesheets/theme/openconext/error-page/_background.sass
MKodde: Yes, that's true, but it also allows for a smaller feedback loop, where the biggest issues can be addressed before rolling the change onto a test/staging env.

PR #719, opened 2019-07-15
Add AuthnContextClassRef attribute blacklist filter

pablothedude:
Add a configurable regex filter to be able to blacklist certain AuthnContextClassRef attribute values. This will be used to prevent the use of internally used attributes by IdP's other than EngineBlock. And to allow second factor only (SFO) flow in EB.

Also created a [PR](https://github.com/OpenConext/OpenConext-deploy/pull/237) to support the configuration parameter in deploy.

See: https://www.pivotaltracker.com/story/show/166702153

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -162,6 +162,12 @@ pdp.password = "secret"
 pdp.client_id = "EngineBlock"
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;; SFO SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; This PCRE regex is used to blacklist incoming AuthnContextClassRef attributes on
+sfo.authn_context_class_ref_blacklist_regex = ""
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;; PROFILE SETTINGS ;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
--- a/bin/composer/dump-required-ini-params.sh
+++ b/bin/composer/dump-required-ini-params.sh
@@ -236,7 +236,10 @@ $ymlContent = array(
                 'attributeDefinitionFile', $projectRoot . 'application/configs/attributes-v2.2.0.json'
             )
         ),
-        'monitor_database_health_check_query'                     => $config->get('openconext.monitor_bundle_health_query', '')
+        'monitor_database_health_check_query'                     => $config->get('openconext.monitor_bundle_health_query', ''),
+
+        // SFO
+        'sfo.authn_context_class_ref_blacklist_regex'             => escapeYamlValue($config->get('sfo.authn_context_class_ref_blacklist_regex'))
     )
 );
 
--- a/languages/messages.nl.php
+++ b/languages/messages.nl.php
@@ -244,6 +244,8 @@ Je %organisationNoun% heeft je de toegang geweigerd tot deze dienst. Je zult dus
     'error_stuck_in_authentication_loop_desc' => '<p>
         Je bent succesvol ingelogd bij je Identity Provider maar de dienst waar je naartoe wilt stuurt je weer terug naar %suiteName%. Omdat je succesvol bent ingelogd, stuurt %suiteName% je weer naar de dienst, wat resulteert in een oneindig zwart gat. Dit komt waarschijnlijk door een foutje aan de kant van de dienst.
     </p>',
+    'error_authn_context_class_ref_blacklisted'                     => 'Error - Waarde van AuthnContextClassRef is niet toegestaan',
+    'error_authn_context_class_ref_blacklisted_desc'                => '<p>Je kunt niet inloggen omdat je %organisationNoun% een waarde stuurde voor AuthnContextClassRef die niet is toegestaan.</p>',
     'error_no_authentication_request_received' => 'Geen authenticatie-aanvraag ontvangen.',
     /**
      * %1 AttributeName
--- a/library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
+++ b/library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
@@ -1,0 +1,65 @@
+<?php
+
+use OpenConext\EngineBlock\Assert\Assertion;
+use Psr\Log\LoggerInterface;
+
+/**
+ * The validator will block any incoming IdP assertion with a AuthnContextClassRef value in the configured namespace
+ * with an error. As it MUST be impossible for an IdP to 'impersonate' the blacklisted values used for AuthnContextClassRef.
+ **/
+class EngineBlock_Corto_Filter_Command_ValidateAuthnContextClassRef extends EngineBlock_Corto_Filter_Command_Abstract
+{
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
+     * @var string
+     */
+    private $authnContextClassRefBlacklistPattern;
+
+    public function __construct(LoggerInterface $logger, $authnContextClassRefBlacklistPattern)
+    {
+        Assertion::nullOrString($authnContextClassRefBlacklistPattern, 'The authnContextClassRefBlacklistPattern must be a string or null');
+
+        $this->logger = $logger;
+        $this->authnContextClassRefBlacklistPattern = $authnContextClassRefBlacklistPattern;
+    }
+
+    /**
+     * @throws EngineBlock_Corto_Exception_AuthnContextClassRefBlacklisted
+     */
+    public function execute()
+    {
+        if (empty($this->authnContextClassRefBlacklistPattern)) {
+            $this->logger->notice('No authn_context_class_ref_blacklist_regex found in the configuration, not validating AuthnContextClassRef');
+            return;
+        }
+
+        $authnContextClassRef = $this->_response->getAssertion()->getAuthnContextClassRef();
+
+        if (!$this->validateAuthnContextClassRef($authnContextClassRef, $this->authnContextClassRefBlacklistPattern)) {
+            throw new EngineBlock_Corto_Exception_AuthnContextClassRefBlacklisted(
+                sprintf(
+                    'Assertion from IdP contains a blacklisted AuthnContextClassRef "%s"',
+                    $authnContextClassRef
+                )
+            );
+        }
+    }
+
+    /**
+     * @param string $value
+     * @param string $regex
+     * @return bool
+     */
+    private function validateAuthnContextClassRef($value, $regex)
+    {
+        if (preg_match($regex, $value)) {
+            return false;
+        }
+
+        return true;
+    }
+}
--- a/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
@@ -14,6 +14,7 @@ use Twig_Environment;
 /**
  * @SuppressWarnings(PHPMD.TooManyPublicMethods) Mimics the previous methodology, will be refactored
  *  see https://www.pivotaltracker.com/story/show/107565968
+ * @SuppressWarnings(PMD.TooManyMethods)
  */
 class FeedbackController
 {
@@ -191,6 +192,17 @@ class FeedbackController
         return new Response(
             $this->twig->render('@theme/Authentication/View/Feedback/missing-required-fields.html.twig'),
             400
+        );
+    }
+
+    /**
+     * @return Response
+     */
+    public function authnContextClassRefBlacklistedAction()
+    {
+        return new Response(
+            $this->twig->render('@theme/Authentication/View/Feedback/authn-context-class-ref-blacklisted.html.twig'),
+            403
         );
     }
 
--- a/src/OpenConext/EngineBlockBundle/Resources/config/routing/feedback.yml
+++ b/src/OpenConext/EngineBlockBundle/Resources/config/routing/feedback.yml
@@ -48,6 +48,11 @@ authentication_feedback_missing_required_fields:
     methods:    [GET]
     defaults:   { _controller: engineblock.controller.authentication.feedback:missingRequiredFieldsAction }
 
+authentication_authn_context_class_ref_blacklisted:
+    path:       '/authn_context_class_ref_blacklisted'
+    methods:    [GET]
+    defaults:   { _controller: engineblock.controller.authentication.feedback:authnContextClassRefBlacklistedAction }
+
 authentication_feedback_invalid_attribute_value:
     path:       '/invalid-attribute-value'
     methods:    [GET]
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
@@ -320,6 +320,15 @@ Feature:
       # The session is lost, so the feedback information (containing the request id) can no longer be rendered on page
     Then I should not see the same request id on the error page
 
+  Scenario: I log in at my Identity Provider, that has an invalid AuthnContextClassRef attribute.
+    Given the IdP "Dummy Idp" sends AuthnContextClassRef with value "invalid-authn-context-class-ref"
+    When I log in at "Dummy SP"
+      And I pass through EngineBlock
+      And I pass through the IdP
+      And I give my consent
+    Then I should see "Error - AuthnContextClassRef value is disallowed"
+
+
 
 #  Scenario: I try an unsolicited login (at EB) but mess up by not specifying a location
 #  Scenario: I try an unsolicited login (at EB) but mess up by not specifying a binding
--- a/tests/library/EngineBlock/Test/Corto/Filter/Command/ValidateAuthnContextClassRefTest.php
+++ b/tests/library/EngineBlock/Test/Corto/Filter/Command/ValidateAuthnContextClassRefTest.php
@@ -1,0 +1,77 @@
+<?php
+
+use Monolog\Handler\TestHandler;
+use Monolog\Logger;
+use OpenConext\EngineBlock\Metadata\Entity\IdentityProvider;
+use OpenConext\EngineBlock\Metadata\ShibMdScope;
+use PHPUnit_Framework_TestCase as UnitTest;
+use SAML2\Assertion;
+use SAML2\Response;
+
+class EngineBlock_Test_Corto_Filter_Command_ValidateAuthnContextClassRefTest extends UnitTest
+{
+    /**
+     * @var TestHandler
+     */
+    private $handler;
+
+    /**
+     * @var Logger
+     */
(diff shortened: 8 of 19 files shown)


pablothedude:
@domgon would you be so kind to validate the Portuguese translations in this PR?
https://github.com/OpenConext/OpenConext-engineblock/pull/719/files#diff-a9d09e45c9163e92df703f169a1c2309

domgon:
Hi,

the translations are bellow.

```
    'error_authn_context_class_ref_blacklisted'                     => 'Erro - O valor para AuthnContextClassRef não é permitido',
    'error_authn_context_class_ref_blacklisted_desc'                => '<p>Não pode autenticar-se porque a sua %organisationNoun% enviou um valor para AuthnContextClassRef que não é permitido.</p>',
```

Best regards,
Domingos

languages/messages.nl.php
```diff
@@ -244,6 +244,8 @@
     'error_stuck_in_authentication_loop_desc' => '<p>
         Je bent succesvol ingelogd bij je Identity Provider maar de dienst waar je naartoe wilt stuurt je weer terug naar %suiteName%. Omdat je succesvol bent ingelogd, stuurt %suiteName% je weer naar de dienst, wat resulteert in een oneindig zwart gat. Dit komt waarschijnlijk door een foutje aan de kant van de dienst.
     </p>',
+    'error_authn_context_class_ref_blacklisted'                     => 'Error - Waarde van AuthnContextClassRef is niet toegestaan',
+    'error_authn_context_class_ref_blacklisted_desc'                => '<p>Je kunt niet inloggen omdat je %organisationNoun% een waarde stuurde voor AuthnContextClassRef die niet is toegestaan</p>',
```
MKodde: Missing '.' ?

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+use Psr\Log\LoggerInterface;
+
+/**
+ * Validate if the AuthnContextClassRef is not blacklisted.
```
MKodde: This would be an excellent place to state/explain the business rules that resulted in the creation of this validator.

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
```diff
@@ -0,0 +1,63 @@
...
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
+     * @var string
+     */
+    private $authnContextClassRefBlacklistPattern;
+
+    public function __construct(LoggerInterface $logger, $authnContextClassRefBlacklistPattern)
```
MKodde: Have you considered creating a modern day Symfony service that is injected into this legacy command? 

This way we venture away from creating new Corto BL. And increase testability. I would greatly applaud that being changed here!

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->authnContextClassRefBlacklistPattern = $authnContextClassRefBlacklistPattern;
+    }
+
+    /**
+     * @throws EngineBlock_Corto_Exception_AuthnContextClassRefBlacklisted
+     */
+    public function execute()
+    {
+        if (empty($this->authnContextClassRefBlacklistPattern)) {
+            $this->logger->notice('No authn_context_class_ref_blacklist_regex found in the configuration, not verifying AuthnContextClassRef');
```
MKodde: For the sake of consistency: ~~verifying~~ validating

MKodde: Ask yourself, when reading this log message, is it directly clear why this validation is skipped? Maybe you could add something about not testing for SFO impersonation by remote IdP? I'm not fully sold on the current log message. @thijskh @pmeulen @quartje what are your thoughts on this? As you guys will be actually reading this message in the future.

tests/library/EngineBlock/Test/Corto/Filter/Command/ValidateAuthnContextClassRefTest.php
```diff
@@ -0,0 +1,78 @@
...
+    public function testNotMatchedBlacklistedRegexpPasses()
+    {
+        $verifier = new EngineBlock_Corto_Filter_Command_ValidateAuthnContextClassRef($this->logger, '/test/');
+        $verifier->setResponse($this->response);
+        $verifier->setIdentityProvider(new IdentityProvider('OpenConext'));
+
+        $verifier->execute();
+    }
+
+    public function testMatchedBlacklistedRegexpThrowsException()
```
MKodde: It would be nice to make this test match more closely with the actual use case we are going to use this for. This test also advocates the move of the logic to modern EB! as per my previous comment :top: 

Idea: create a data provider that tries different AuthnContextClassRefs with different of our configured loa's.

bin/composer/dump-required-ini-params.sh
```diff
@@ -236,7 +236,10 @@ $ymlContent = array(
                 'attributeDefinitionFile', $projectRoot . 'application/configs/attributes-v2.2.0.json'
             )
         ),
-        'monitor_database_health_check_query'                     => $config->get('openconext.monitor_bundle_health_query', '')
+        'monitor_database_health_check_query'                     => $config->get('openconext.monitor_bundle_health_query', ''),
```
MKodde: Do we want/need additional input validation? If not for our own safety, it could also help users with setting a correct regex.

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
```diff
@@ -0,0 +1,63 @@
...
+            return;
+        }
+
+        $authnContextClassRef = $this->_response->getAssertion()->getAuthnContextClassRef();
+
+        if (!$this->validateAuthnContextClassRef($authnContextClassRef, $this->authnContextClassRefBlacklistPattern)) {
+            throw new EngineBlock_Corto_Exception_AuthnContextClassRefBlacklisted(
+                sprintf(
+                    'Disallowed AuthnContextClassRef value "%s" in response from IdP "%s" to SP "%s"',
```
MKodde: I think we could improve this error / log message. 

Suggestion, follow the story provided message: 

>"Assertion from IdP <IdPEntityID> contains a blacklisted AuthnContextClassRef: '<the blacklisted AuthnContextClassRef>'"

Or something along the lines of:

> The IdP: "%s" provided a blacklisted AuthnContextClassRef: "%s" while sending an assertion to SP: "%s"

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
```diff
@@ -194,6 +195,17 @@ public function missingRequiredFieldsAction()
         );
     }
 
+    /**
+     * @return Response
+     */
+    public function authnContextClassRefBlacklistedAction()
```
MKodde: Didn't the story explicitly state that no additional error screen was required? 

>The error when a assertion in blocked due to a blacklisted AuthnContextClassRef does not need its own screen, but should be clearly logged. E.g. "Assertion from IdP <IdPEntityID> contains a blacklisted AuthnContextClassRef: '<the blacklisted AuthnContextClassRef>'"

MKodde: Is 403 the correct error code? 400 - Forbidden also seems appropriate. Or as per the original intent, a 500 error?

src/OpenConext/EngineBlockBundle/Resources/config/routing/feedback.yml
```diff
@@ -48,6 +48,11 @@ authentication_feedback_missing_required_fields:
     methods:    [GET]
     defaults:   { _controller: engineblock.controller.authentication.feedback:missingRequiredFieldsAction }
 
+authentication_authn_context_class_ref_blacklisted:
```
MKodde: If we are keeping this error page, please also add it to the visual regression test

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
```diff
@@ -320,6 +320,15 @@ Feature:
       # The session is lost, so the feedback information (containing the request id) can no longer be rendered on page
     Then I should not see the same request id on the error page
 
+  Scenario: I log in at my Identity Provider, that has an invalid AuthnContextClassRef attribute.
+    Given the IdP "Dummy Idp" sends AuthnContextClassRef "invalid-authn-context-class-ref"
```
MKodde: Consider setting this with more a more in context value. Like `https://engine.example.org/loa1` or whatnot. This [sh|c]ould make the test more recognizable in the future.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
```diff
@@ -6,6 +6,7 @@
 use Behat\Mink\Exception\ExpectationException;
 use DOMDocument;
 use DOMXPath;
+use EngineBlock_ApplicationSingleton;
```
MKodde: The imports in this class can be removed I suppose?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
MKodde: For consistency with the other step definitions I suggest::
> Given the IdP "Dummy Idp" sends AuthnContextClassRef **with value** "invalid-authn-context-class-ref"

application/configs/application.ini
```diff
@@ -161,6 +161,11 @@ pdp.username = "pdp_admin"
 pdp.password = "secret"
 pdp.client_id = "EngineBlock"
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;; SFO SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+sfo.authn_context_class_ref_blacklist_regex = ""
```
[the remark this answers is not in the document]
pablothedude: @MKodde I didn't want to focus on any documentation before a functional review was done. Your suggestions however are very valuable. I shall create a story to do this afterwards and will add the raised concerns to the story if you agree that will be a workable solution.

I did also think about about using a custom feature flag to enable/disable SFO completely so you don't need a regex to implicitly enable/disable SFO.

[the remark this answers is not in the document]
pablothedude: Created story: https://www.pivotaltracker.com/story/show/167320950

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
pablothedude: If you look into git you could see why it is added.

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
pablothedude: I've considered it but I didn't found it an improvement. I think it will be better to bring Corto up to date at once by fixing the issues below so it won't be considered legacy anymore:
* add namespaces
* prevent the use of superglobals
* eliminate the magic loading of services
* prevent magic method calls `$object->$parameterName()`
* introduce DI if not used

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
pablothedude: @MKodde  changed the message as suggested

tests/library/EngineBlock/Test/Corto/Filter/Command/ValidateAuthnContextClassRefTest.php
pablothedude: The only thing I can think of why the code isn't that modern, is the lack of using namespaces so I don't agree on this one.

The responsibility of the class is to match against a regex and that gets tested, the regex itself is language specific and I don't think that needs to be tested. Could you elaborate what the added benefit will be to test that logic?

bin/composer/dump-required-ini-params.sh
pablothedude: Input validation for a regex? How do you picture that?

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
pablothedude: I did reuse the the battle tested message from `library/EngineBlock/Corto/Filter/Command/ValidateAllowedConnection.php`. I however did change it to your first suggestion in terms of efficiency.

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
pablothedude: I see it, I think I misread it. Status code 400 is bad request and 403 is forbidden, what do you mean by `400 forbidden`?

src/OpenConext/EngineBlockBundle/Resources/config/routing/feedback.yml
pablothedude: Good suggestion, done

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
pablothedude: I deliberately wanted to make it more unique because the regex was injected with the DiContainer and therefore it could be found much easier and prevent a certain level of WTF.
`\EngineBlock_Application_FunctionalTestDiContainer::getAuthnContextClassRefBlacklistRegex`

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/EngineBlockContext.php
pablothedude: :+1:

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/ClearErrorMessages.feature
pablothedude: Done

application/configs/application.ini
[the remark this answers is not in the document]
MKodde: Okay, thanks for the story! I still think this is the time to add in-line documentation to some parts of the code. For example to the application.ini. The additional wiki page is still a great idea, but I feel we should also be helpfull in a sense of in-line documentation when it comes to writing user-facing code (like config files).

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
MKodde: I do agree with you, you would introduce some coupling to the legacy code. Which in my opinion is not that big of a deal. Delving new BL into a legacy system in my opinion is far worse. Having the code in the new place prevents us from having to refactor it when the big bang happens. For now the logic is quite simple and the problem is not that evident. But in upcoming PR's we will be adding additional rules to this Command.

tests/library/EngineBlock/Test/Corto/Filter/Command/ValidateAuthnContextClassRefTest.php
MKodde: I think a perfect unit test covers all of the different possible uses of a unit of code regardless of implementation. If a requirement states that a validator should accept loa-1, loa-2, loa-3 but NOT loa-4, then it makes sense to add a unit test for those rules. How the actual piece of code is written is beside the point. It's all about creating a set of tests that validate our business rules.

And for my previous comment:

> It would be nice to make this test match more closely with the actual use case we are going to use this for

What I meant by this is that instead of setting the blacklisted regex to something not to be used in real life: `/urn:oasis:names:tc:SAML:2\.0:ac:classes:Password/`, you could write the test to more closely match reality.

bin/composer/dump-required-ini-params.sh
MKodde: You could test for it being either an empty string, or it being an actual string value. There are ways to validate if the provided string would function as a regular expression. But validating that seems less useful.

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
MKodde: My bad, 400 indeed is a bad request. Which could also apply here dont you think?

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
MKodde: What's wrong with adding some details about the validator to the class' doc block? Reading git log will probably get you there as well, but is far less user-friendly.

Suggestion: 

> The validator will block any incoming IdP assertion with a AuthnContextClassRef value in our namespace with an error. As it MUST be impossible for an IdP to 'impersonate' our values used for AuthnContextClassRef.

application/configs/application.ini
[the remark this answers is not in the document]
pablothedude: Added additional documentation to the `applicatio.ini`

library/EngineBlock/Corto/Filter/Command/ValidateAuthnContextClassRef.php
pablothedude: I don't see the introduced code as legacy as the only thing which should be addressed is the lack of namespaces. If I did have encountered other smells I would have directly boyscouted it so it could be refactored very easy later on.

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
pablothedude: In the ticket I've asked Thijs. For now I'll keep it a 403 and will hotfix it if another status code is required.

thijskh: 403 is fine I think

tests/library/EngineBlock/Test/Corto/Filter/Command/ValidateAuthnContextClassRefTest.php
pablothedude: I think a unittest should only test the possibilities of a certain unit. The class should filter the `AuthnContextClassRef` attribute and that gets tested.

bin/composer/dump-required-ini-params.sh
pablothedude: A good suggestion, done

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
pablothedude: :+1:

pablothedude:
> Nice work! I've found some points of discussion during my code review, I've added them below.
> 
> Other concerns:
> 
>     1. Consider updating the readme with a section on SFO configuration. Or link to a OC wiki page where more information about this topic can be read. And improve in-line documentation (see some suggestions in code review)
> 
Created story

>     2. Are full PCRE regexes allowed? And should the users provide delimiters, or are they pre-configured? Again information that should become clear when opening the configuration.
> 
Added a bit of documentation to `application.ini` and a task to the documentation story.

>     3. Please document the additional requirement of the custom error page in Pivotal, or remove it from the PR. I personally opt for the first.
I've updated the task to reflect this.

PR #722, opened 2019-07-16
Hide language switch on 400 and 405 page

pablothedude:
The language switch is hidden because it will not work out of the box
because the lang parameter is read from the request but the parameters
are empty when on a 404 page.

See: https://www.pivotaltracker.com/story/show/167202103/comments/204653786


MKodde:
:shipit:

PR #725, opened 2019-07-19
Create gateway SFO mock

pablothedude:
In order to test the SFO functionality in EB a gateway mock is needed.
Therefore a new endpoint is created in the functional testing bundle.
The results could be mocked in order to fake the possible SFO results.

`functional-testing/gateway/second-factor-only/single-sign-on`

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/SfoContext.php
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/SfoContext.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace OpenConext\EngineBlockFunctionalTestingBundle\Features\Context;
+
+use OpenConext\EngineBlockFunctionalTestingBundle\Fixtures\FunctionalTestingSfoGatewayMockConfiguration;
+use OpenConext\EngineBlockFunctionalTestingBundle\Mock\EntityRegistry;
+use OpenConext\EngineBlockFunctionalTestingBundle\Mock\MockIdentityProviderFactory;
+use OpenConext\EngineBlockFunctionalTestingBundle\Mock\MockServiceProviderFactory;
+use SAML2\Constants;
+
+class SfoContext extends AbstractSubContext
+{
+    /**
+     * @var EntityRegistry
+     */
+    private $mockSpRegistry;
+    /**
+     * @var EntityRegistry
+     */
+    private $mockIdpRegistry;
+    /**
+     * @var FunctionalTestingSfoGatewayMockConfiguration
+     */
+    private $gatewayMockConfiguration;
+    /**
+     * @var MockIdentityProviderFactory
+     */
+    private $idpFactory;
+    /**
+     * @var MockServiceProviderFactory
+     */
+    private $spFactory;
+
+    /**
+     * @param EntityRegistry $mockSpRegistry
+     * @param EntityRegistry $mockIdpRegistry
+     * @param MockIdentityProviderFactory $idpFactory
+     * @param MockServiceProviderFactory $spFactory
+     * @param FunctionalTestingSfoGatewayMockConfiguration $gatewayMockConfiguration
+     */
+    public function __construct(
+        EntityRegistry $mockSpRegistry,
+        EntityRegistry $mockIdpRegistry,
+        MockIdentityProviderFactory $idpFactory,
+        MockServiceProviderFactory $spFactory,
+        FunctionalTestingSfoGatewayMockConfiguration $gatewayMockConfiguration
+    ) {
+        $this->mockSpRegistry = $mockSpRegistry;
+        $this->mockIdpRegistry = $mockIdpRegistry;
+        $this->gatewayMockConfiguration = $gatewayMockConfiguration;
+        $this->idpFactory = $idpFactory;
+        $this->spFactory = $spFactory;
+    }
+
+    /**
+     * @Given /^SFO is used$/
+     */
+    public function sfoIsUsed()
+    {
+        $mockIdp = $this->idpFactory->createNew('sfoHostedIdp');
+        $this->gatewayMockConfiguration->setMockIdentityProvider($mockIdp);
+
+        $mockSp = $this->spFactory->createNew('ebSfoSp');
+        $this->gatewayMockConfiguration->setMockServiceProvider($mockSp);
+    }
+
+    /**
+     * @Given /^SFO will successfully verify a user$/
+     */
+    public function sfoWillsSuccessfullyVerifyAUser()
+    {
+        $this->gatewayMockConfiguration->unsetMessage();
+    }
+
+    /**
+     * @Given /^SFO will fail as the user cancelled$/
+     */
+    public function sfoWillFailAsTheUserCancelled()
+    {
+        $this->gatewayMockConfiguration->setMessage(Constants::STATUS_RESPONDER, Constants::STATUS_AUTHN_FAILED, 'Authentication cancelled by user');
+    }
+
+    /**
+     * @Given /^SFO will fail as the loa can not be given$/
+     */
+    public function sfoWillFailAsTheLoaCanNotBeGiven()
+    {
+        $this->gatewayMockConfiguration->setMessage(Constants::STATUS_RESPONDER, Constants::STATUS_NO_AUTHN_CONTEXT);
+    }
+}
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
@@ -1,0 +1,21 @@
+Feature:
+  In order to support 2FA we need to support gateway interactions
+  As EngineBlock
+  I want to support 2FA by utilizing the Stepup gateway SFO capabilities
+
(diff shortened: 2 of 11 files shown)


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/SfoContext.php
```diff
@@ -0,0 +1,90 @@
...
+    ) {
+        $this->mockSpRegistry = $mockSpRegistry;
+        $this->mockIdpRegistry = $mockIdpRegistry;
+        $this->gatewayMockConfiguration = $gatewayMockConfiguration;
+        $this->idpFactory = $idpFactory;
+        $this->spFactory = $spFactory;
+    }
+
+    /**
+     * @Given /^SFO is used$/
```
MKodde: 1. In the light of an authentication SFO is not used, we are using StepUp's SFO capability to facilitate a 2FA. By not using the SFO term in the Behat scenarios & step definitions. The test will become easier to understand for new users.
2. And personally, I'd use 2FA *is enabled* instead of _is used_. Other EB features are also named like that in the other step definitions.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
```diff
@@ -0,0 +1,21 @@
+Feature:
+  In order to support 2FA we need to support gateway interactions
```
MKodde: It will probably be unclear for a casual reader what gateway interactions are. Please rephrase this sentence to be more clear upon first glance.

Suggestion:

_In order to support second factor authentications_
_As EngineBlock_
_I want to utilize Stepup Gatway's SFO capabilities_

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
```diff
@@ -0,0 +1,21 @@
...
+  As EngineBlock
+  I want to support 2FA by utilizing the Stepup gateway SFO capabilities
+
+  Background:
+    Given an EngineBlock instance on "vm.openconext.org"
```
MKodde: Perhaps configure only one SP / IdP pair for now? This eliminates the WAYF step in the Behat tests. Also, consider enabling 2FA in the `Background` tasks? Or maybe this is not doable for a reason unknown to me?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
```diff
@@ -0,0 +1,21 @@
...
+      And no registered SPs
+      And no registered Idps
+      And an Identity Provider named "SSO-IdP"
+      And an Identity Provider named "SSO-Foobar"
+      And a Service Provider named "SSO-SP"
+      And a Service Provider named "SSO-Foobar"
+
+  Scenario: Sfo should be supported
```
MKodde: Suggestion for improved step definition phrasing:

_Scenario: 2FA can be used _
_Given 2FA is enabled_
_When I log in at "SSO-SP"_ **<-- do we not need to specify that this SP uses 2FO?**
_And I select "SSO-IdP" on the WAYF_ **<-- can maybe be omitted as per my comment above**
_And I pass through EngineBlock_
_And I successfully verify the second factor of the user_
_Then the AuthnRequest to submit should match xpath'_ **<-- can we test a specific xpath related to the 2FO authentication?**

src/OpenConext/EngineBlockFunctionalTestingBundle/Mock/MockSfoGateway.php
```diff
@@ -0,0 +1,408 @@
...
+        $response->setStatus($status);
+
+        return $response;
+    }
+
+    /**
+     * @param Response $response
+     * @return string
+     */
+    public function getResponseAsXML(Response $response)
```
MKodde: Is this method in use?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/SfoContext.php
pablothedude: # If you read the issues and documentation you'll find that everywhere SFO is used and I did want to confine myself to the ubiquitous language and not create new terminology.
# Regarding 2fa see my answer above, I deliberately didn't want to use `enabled` beacause this is no feature toggle while all other uses are. I however would like to add a toggle and did create a separate issue in pivotal to discuss this. If we add a feature toggle I think it would be a good sugegstion to change it.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
pablothedude: Why should a casual reader understand `Stepup Gatway's SFO capabilities` and not `gateway interactions`? Do you have an example of what kind of casual user you mean because personally I can't come up with any?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
pablothedude: I did want the to select the IdP in the Wayf step so it would become clear which IdP is selected
without an IdP being implicitly selected.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
pablothedude: This was changed later on.

src/OpenConext/EngineBlockFunctionalTestingBundle/Mock/MockSfoGateway.php
pablothedude: Good catch, removed

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
MKodde: Isn't the whole point of Gherkin that a test is clear and understandable? Preferably written in a manner that a 'customer' can understand (taken from Gherkin/Cucumber docs). In my opinion, using a somewhat vague reference to software that is interacted with is not very clear.

> Why should a casual reader understand Stepup Gatway's SFO capabilities

A casual reader is now informed the Stepup Gateway is used, and specifically, the SFO feature of the gateway is used.

> Do you have an example of what kind of casual user you mean because personally, I can't come up with any?

A new development team, or a contributor from the community.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
MKodde: But what's the added value? This test does not test the WAYF, but SFO right? It's no impediment, but the more focussed a test is, the easier it is to read/understand/maintain.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
pablothedude: Won;t fix, this could be addressed when renaming SFO -> 2FA

pablothedude: Personally I don't think that your suggestion do make much of a difference especially beacause it's mentioned a single line lower:

`I want to support 2FA by utilizing the Stepup gateway SFO capabilities`

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
pablothedude: I think there is an added value to not implicitly select an IDP on the wayf. So it would make things much more clear. And in order to handle SFO an IDP is needed.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
pablothedude: But feel free to address it in https://www.pivotaltracker.com/story/show/167864360

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Sfo.feature
MKodde: I still don't see the added benefit. It might become viable when different SP/IdP combinations are configured to use step-up auth under different conditions. I'm fine with leaving it like this, just wondering what your train of thought was. You did not react to this portion of my feedback yet:

> Also, consider enabling 2FA in the Background tasks? Or maybe this is not doable for a reason unknown to me?

pablothedude: I left the `enable 2FA` in there to toggle the feature flag. But a feature toggle is not needed. I've added this to the sfo to 2fa cleanup task issue to be removed then.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/SfoContext.php
pablothedude: This will be addressed later on:
https://www.pivotaltracker.com/story/show/167864360

PR #727, opened 2019-07-23
Improved error texts

arnoutt:
As preparation for the new error page designs.

### DIFF
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -179,75 +179,54 @@ If you want to use this application you have to:<br />
 <ul><li>restart your browser</li>
 <li>login again</li>
 <li>share your information</li></ul>',
-    'error_no_idps'                 => 'Error - No Identity Providers found',
-    'error_no_idps_desc'            => '<p>
-The service you\'re trying to reach (&lsquo;Service Provider&rsquo;) is not accessible through %suiteName%.<br /><br />
-    </p>',
+    'error_no_idps'                 => 'Error - No %organisationNounPlural% found',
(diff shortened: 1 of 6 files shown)


languages/messages.en.php
```diff
@@ -179,75 +179,54 @@
 <ul><li>restart your browser</li>
 <li>login again</li>
 <li>share your information</li></ul>',
-    'error_no_idps'                 => 'Error - No Identity Providers found',
-    'error_no_idps_desc'            => '<p>
-The service you\'re trying to reach (&lsquo;Service Provider&rsquo;) is not accessible through %suiteName%.<br /><br />
-    </p>',
+    'error_no_idps'                 => 'Error - No %organisationNounPlural% found',
+    'error_no_idps_desc'            => 'Logging into this service is not possible via %suiteName%. The service is not connected to any institutions.',
```
thijskh: Also use `%organisationNounPlural%` here instead of institutions

thijskh:
@pablothedude Arnout's change drops a number of '<p>' tags. Is that a problem you think? It does look better to me now not having that HTML in there...

pablothedude:
> @pablothedude Arnout's change drops a number of `<p>` tags. Is that a problem you think? It does look better to me now not having that HTML in there...

@thijskh it shouldn't be an issue because it would only be used for styling and if it looks good to you I'm fine with that. I don;t now if you've seen the changes yourself else it would be possible to generate some screen shots in order to validate the changes.

thijskh:
@pablothedude This can be merged but also needs to be merged to 5.11, can you do that?

pablothedude:
Done

PR #730, opened 2019-08-06
Add theme variable overrides

pablothedude:
The vars couldn't be overridden due to the loading order of sass files.
This should be fixed now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,9 @@ before_script:
   - sudo a2dissite 000-default
   - sudo a2ensite default-ssl
   - sudo service apache2 restart
+  - echo "Chrome version `google-chrome --version`"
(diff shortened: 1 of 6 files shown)


tvdijen:
Works like a charm!

pablothedude:
I've merged the missed variable from #732 back into this branch so it could be fixed all at once.

MKodde:
We should target this release at 5.11 and onward. The current target is master (5.13). Be sure to cherry-pick these changes to the appropriate branches when finishing this!

pablothedude:
On `master` is currently 5.11. The upcoming 5.11 and 5.12 releases are currently only in `feature/sfo` and are not yet merged back. So I don't think any cherry picking will be needed.

PR #731, opened 2019-08-07
Fix color

tvdijen:
This one stays azure if you change the base colors


pablothedude:
Thanks @tvdijen, I will merge this back into the theme overrides branch so it will be fixed at once.

PR #734, opened 2019-08-14
Fix the keys

pablothedude:
The engineblock keys used for SFO seemed to be coming from a lot of
different places on different environments. When implementing sfo only
the ci keys were used. So now this is refactored to use the same keys
for ~~dev and test~~ test and functionaltest to make things much more transparent.

The behat tests are currently ran as root but this is not necessary
anymore as it wil be chmodded to 440 and chowned as user:root to reflect
the production situation much better.

https://www.pivotaltracker.com/story/show/167869916


MKodde:
This change cleans up things nicely!

Should updating https://github.com/OpenConext/OpenConext-engineblock/blob/feature/sfo-fix-keys-env/application/configs/application.ini#L174 part of this change?

pablothedude:
> This change cleans up things nicely!
> 
> Should updating https://github.com/OpenConext/OpenConext-engineblock/blob/feature/sfo-fix-keys-env/application/configs/application.ini#L174 part of this change?

I have left it this way so no one could be accidentally use the test keys in prod as defence in depth. I see I did mention dev and test in the description but I meant test and functional test. I'll update the documentation accordingly when you've finished your review.

PR #736, opened 2019-08-15
Make Wiki links configurable

MKodde:
The wiki links can now be configured on a 'per error template' basis. In addition to that, the wiki links are also internationalized. So each language can have its own page.

There is some controversy in this PR. My aim was to limit the number of code changes to achieve this change. And to prevent additional complexity when creating a new custom error page. The controversy lies in the way the decision is made whether or not to show the wiki link. This is done in the Twig template. Other solutions included:

1. Make the wiki link decision in the controller actions of each and every error page.
2. Create an 'After filter' event that is executed after every controller action.
3. Attempt to normalize the feedback controller actions and include the wiki link in that solution.

For now, I stuck with the currently proposed solution. Feel free to think along, and suggest an improvement.

https://www.pivotaltracker.com/story/show/165742913

Disclaimer; there is some small cleanup work left to do, but I'm planning to do this after a first review round.


pablothedude:
@MKodde as requested some initial thoughts regarding the error pages.

* I wonder if it wouldn't be relatively easy to use a single twig template because it seems the same content type (an error page) is being used. This will speed up development in the future and will make sure that any additional info will be handled the same instead of any possible way you could think of. Also testing will become easier and the styling will remain constant across templates and it will become easier to maintain. Then the only exception would be the `authorization-policy-violation` template, but that's something we could address later on.
* We could relatively easily create a helper to set translation parameters into the session (feedbackInfo, feedback_custom) and to prevent getters from being used.
* Maybe we could set a key and a status code in the routing.yml and handle it with the same controller action. That key could be used in twig to determine the pagetitle, message and optional wiki link.

PR #741, opened 2019-08-20
Add GW status to feedback info

pablothedude:
The response from GW needed to be shown as additional info on
the error pages.

https://www.pivotaltracker.com/story/show/167834563


src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
```diff
@@ -453,8 +453,10 @@ public function clockIssueAction()
      */
     public function stepupCalloutUserCancelledAction(Request $request)
     {
+        $feedbackInfo = $request->getSession()->get('feedbackInfo');
```
MKodde: No need to change the feedback controller. The Feedback info is not passed as a twig param, but fished out of the session by the Feedback twig extension.

thijskh:
Re 2: you could consider to always strip off any `urn:oasis:names:tc:SAML:2.0:status:` prefix of the statuscode and substatus code before displaying. I would be fine with that.

src/OpenConext/EngineBlockBundle/Controller/StepupController.php
```diff
@@ -65,6 +67,21 @@ public function __construct(
      */
     public function consumeAssertionAction(Request $request)
     {
+
+
+        $statusCodeDescription = 'Responder/AuthnFailed';
```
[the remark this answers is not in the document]
pablothedude: Removed

library/EngineBlock/Corto/Module/Service/StepupAssertionConsumer.php
```diff
@@ -167,13 +168,15 @@ private function handleInvalidGatewayResponse(EngineBlock_Corto_Exception_Receiv
                 }
 
                 throw new EngineBlock_Corto_Exception_InvalidStepupLoaLevel(
+                    $e,
```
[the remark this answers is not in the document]
pablothedude: Fixed

src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php
pablothedude: A nice that we have a TD ticket already then!

src/OpenConext/EngineBlockBundle/Controller/StepupController.php
```diff
@@ -21,6 +21,8 @@
 
 use EngineBlock_ApplicationSingleton;
 use EngineBlock_Corto_Adapter;
+use EngineBlock_Corto_Exception_ReceivedErrorStatusCode;
```
MKodde: These use statements are now unused and should be dropped from this commit

pablothedude: done

pablothedude:
> Almost there! See some code related feedback below.
> 
> In addition to that, some additional remarks:
> 
>     1. The status code and status message feedback information labels might need an abbreviation.
> 
The fields were already implemented without an abbreviation. I have consulted Thijs and for now it was fine to keep it as-is to prevent becoming to cryptic. This could easiliy be changed later on.

>     2. The status code field should be wider. The value does not fit in the cell.
> 
Done, seems to be forgotten when implemented the first time. Uses the full width now.

>     3. I miss the addition of the stepup link as per requirement: `Provide link to page (https://support.surfconext.nl/stepup-noauthncontext) that can be customized with advice e.g. the right link to the self-service portal and add translation.`
> 
Added the link trough translations

>     4. Consider updating the behat test, covering the error pages with a verification that the status code & message is rendered on the stepup pages.
Added the status code to the tests.

PR #743, opened 2019-08-21
Use an SP for stepup authentication entity metadata

pablothedude:
In order to show valid metadata to use when configuring GW a SP
needs to be used. This is fixed now.

https://www.pivotaltracker.com/story/show/166737276

### DIFF
--- a/library/EngineBlock/Application/DiContainer.php
+++ b/library/EngineBlock/Application/DiContainer.php
@@ -20,7 +20,7 @@ use Doctrine\ORM\EntityManager;
 use OpenConext\EngineBlock\Metadata\MetadataRepository\MetadataRepositoryInterface;
 use OpenConext\EngineBlock\Validator\AllowedSchemeValidator;
 use OpenConext\EngineBlockBundle\Stepup\StepupGatewayCallOutHelper;
-use OpenConext\EngineBlockBundle\Stepup\StepupIdentityProvider;
+use OpenConext\EngineBlockBundle\Stepup\StepupEntityFactory;
 use Symfony\Component\DependencyInjection\ContainerInterface as SymfonyContainerInterface;
 
 class EngineBlock_Application_DiContainer extends Pimple
@@ -325,7 +325,19 @@ class EngineBlock_Application_DiContainer extends Pimple
      */
     public function getStepupIdentityProvider(EngineBlock_Corto_ProxyServer $server)
     {
-        return StepupIdentityProvider::fromStepupEndpoint(
+        return StepupEntityFactory::idpFrom(
+            $this->getStepupEndpoint(),
+            $server->getUrl('stepupAssertionConsumerService')
+        );
+    }
+
+    /**
+     * @param EngineBlock_Corto_ProxyServer $server
+     * @return \OpenConext\EngineBlock\Metadata\Entity\ServiceProvider
+     */
+    public function getStepupServiceProvider(EngineBlock_Corto_ProxyServer $server)
+    {
+        return StepupEntityFactory::spFrom(
             $this->getStepupEndpoint(),
             $server->getUrl('stepupAssertionConsumerService')
         );
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -21,7 +21,7 @@ use OpenConext\EngineBlock\Metadata\Entity\IdentityProvider;
 use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
 use OpenConext\EngineBlock\Validator\AllowedSchemeValidator;
 use OpenConext\EngineBlockBundle\Exception\ResponseProcessingFailedException;
-use OpenConext\EngineBlockBundle\Stepup\StepupIdentityProvider;
+use OpenConext\EngineBlockBundle\Stepup\StepupEntityFactory;
 use SAML2\AuthnRequest;
 use SAML2\Binding;
 use SAML2\Certificate\KeyLoader;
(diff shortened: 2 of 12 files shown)


src/OpenConext/EngineBlockBundle/Stepup/StepupEntity.php
```diff
@@ -33,7 +34,7 @@ class StepupIdentityProvider extends ServiceProvider
      * @return IdentityProvider
      * @throws \EngineBlock_Exception
      */
-    public static function fromStepupEndpoint(StepupEndpoint $stepupEndpoint, $acsLocation)
+    public static function idpFromStepupEndpoint(StepupEndpoint $stepupEndpoint, $acsLocation)
```
MKodde: Nit: Consider renaming these factory methods to: idpFrom and spFrom. The first parameter will indicate you are building it from a stepup endpoint. This results in slightly more readable code in my opinion.

src/OpenConext/EngineBlockBundle/Stepup/StepupEntity.php
```diff
@@ -20,11 +20,12 @@
 use EngineBlock_X509_CertificateFactory;
 use OpenConext\EngineBlock\Metadata\Entity\IdentityProvider;
 use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
+use OpenConext\EngineBlock\Metadata\IndexedService;
 use OpenConext\EngineBlock\Metadata\Service;
 use RobRichards\XMLSecLibs\XMLSecurityKey;
 use SAML2\Constants;
 
-class StepupIdentityProvider extends ServiceProvider
+class StepupEntity
```
MKodde: Suggestion: Rename to StepupEntityFactory as this more clearly identifies the purpose of this class.

library/EngineBlock/Application/DiContainer.php
```diff
@@ -325,7 +325,19 @@ public function getTranslator()
      */
     public function getStepupIdentityProvider(EngineBlock_Corto_ProxyServer $server)
     {
-        return StepupIdentityProvider::fromStepupEndpoint(
+        return StepupEntity::idpFromStepupEndpoint(
+            $this->getStepupEndpoint(),
+            $server->getUrl('stepupAssertionConsumerService')
```
MKodde: This looks strange. An IdP is not configured with ACS location, right? Can you explain to me why this configuration is required, and why you use it to set the response processing services?

library/EngineBlock/Corto/Module/Service/Metadata.php
```diff
@@ -18,7 +18,7 @@
 
 use EngineBlock_Corto_Module_Service_Metadata_ServiceReplacer as ServiceReplacer;
 use OpenConext\EngineBlock\Metadata\MetadataRepository\EntityNotFoundException;
-use OpenConext\EngineBlockBundle\Stepup\StepupIdentityProvider;
+use OpenConext\EngineBlockBundle\Stepup\StepupEntity;
```
MKodde: Please remove unused `use` statement

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -105,7 +105,7 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
     private $acsLocationSchemeValidator;
 
     /**
-     * @var StepupIdentityProvider
+     * @var StepupEntity
```
MKodde: It's not really a StepupEntity, but an IdP entity. The StepupEntity is merely the factory that churned out the IdP instance.

src/OpenConext/EngineBlockBundle/Stepup/StepupEntity.php
pablothedude: Good suggestion

src/OpenConext/EngineBlockBundle/Stepup/StepupEntity.php
pablothedude: Good suggestion

library/EngineBlock/Application/DiContainer.php
pablothedude: It's needed to be able to validate the entities when they are posted back to the ConsumeAssertion endpoint where EB is expecting an IDP.

library/EngineBlock/Corto/Module/Service/Metadata.php
pablothedude: Done

library/EngineBlock/Corto/Module/Bindings.php
pablothedude: Fixed

library/EngineBlock/Application/DiContainer.php
MKodde: I don't understand. Which entitties are validated? And where does this happen? The StepupController is the entrypoint for Stepup Gatweway right, and acts as an SP. Once ready it sends the response back into Engine. 

1. Could you further explain this to me, can also be done IRL
2. This might be a good place to add some inline documentation.

By the way: the consume assertion endpoint should only accept POST requests. I'm opening a Pivotal ticket for this.

pablothedude: I misinterpreted your question. I didn't use the ResponseProcessingService because I thought that I could use the ACS location for that. Changed it to internally use the ResponseProcessingService.

MKodde: Quick summary of our IRL talk regarding this. In order to set the ACL location on the AuthNRequest we probably need to inject the SP into the method next to the IdP.

pablothedude: Changed the code so only a IDP is used and for the acs location I can just use the EB stepup acs endpoint so therefore we don't need to have a SP. I hope this will resolve your tension. Feel free to open a new story to discus this further.

PR #746, opened 2019-08-26
Fix the EB private key for both dev and Travis

pablothedude:
When implementing SFO initially the ci test keys were used. This however seemed to work out in Travis but not for local development.
The Behat tests were executed as the local Vagrant user while the private key could only be read as the engine_fpm_user (engine) due to how Openconext-deploy is configured.

https://www.pivotaltracker.com/story/show/167869916


.travis.yml
```diff
@@ -43,12 +43,12 @@ before_script:
...
   # Set rights to private key to reflect prod
-  - sudo chown travis:root ~/build/OpenConext/OpenConext-engineblock/tests/resources/key/engineblock.pem
-  - sudo chmod 0440 ~/build/OpenConext/OpenConext-engineblock/tests/resources/key/engineblock.pem
+  - sudo chown travis:root /etc/openconext/engineblock.pem
+  - sudo chmod 0440 /etc/openconext/engineblock.pem
   ###
   - mkdir -p /tmp/engineblock/cache/test
   - mkdir -p /tmp/engineblock/logs/test
-  - sudo apt-get update
+  #- sudo apt-get update
```
MKodde: Did you disable the apt update intentionally? IMHO either remove or re-enable it.

library/EngineBlock/Application/FunctionalTestDiContainer.php
```diff
@@ -64,16 +64,19 @@ public function getAuthnContextClassRefBlacklistRegex()
     }
 
     /**
+     * This method is used to mock the private key and will return the same private key as in /etc/openconext/engineblock.pem
+     * This is used to let the allow the default user to run behat tests with to open the private key
+     * The private key is needed to validate stepup authentication responses which will use the same key for the gateway
+     *
```
MKodde: Good addition! I think you need to update some of the parts of the comment. For example you state:

> and will return the same private key as in /etc/openconext/engineblock.pem

Where it actually returns the `$basePath . '/ci/travis/files/engineblock.pem'`.

And: 

> ... which will use the same key for the gateway

Using the term gateway is fairly unclear. It's not evident you are referring to the Stepup Gateway. Maybe, using the 'stepup authentication remote gateway' term, or something along those lines, is more clear here.

application/configs/application.ini
```diff
@@ -69,11 +69,11 @@ engineblock.maximumAuthenticationProceduresAllowed  = 5
 ; When EngineBlock signs responses (when it acts as an Idp)
 ; or requests (when it acts as an SP) it uses these X.509 certs.
 encryption.keys.default.publicFile = "/etc/openconext/engineblock.crt"
-encryption.keys.default.privateFile = "/etc/openconext/engineblock.key"
+encryption.keys.default.privateFile = "/etc/openconext/engineblock.pem"
```
MKodde: Using the `.pem` extension is not considered best practice for private keys right? I would expect a `.key` extension. If you have good reason to change this convention. Please also advertise this in the upgrading release notes as this would probably need attention when upgrading to this EB version.

.travis.yml
pablothedude: Will fix this later in a separate PR, This was broken in Travis itself.

library/EngineBlock/Application/FunctionalTestDiContainer.php
pablothedude: Updated the comment to reflect that it's the same key only a different files to be able to read the key when not the php-fpm user in tests.

application/configs/application.ini
pablothedude: As far as I know there isn't a best practice to use, because the keys should be validated based on the PEM envelope being used. I've changed it to reflect the situation as currently used in `etc/openconext/application.ini` (OC-Deploy) and to keep it uniform across the project. This to keep them at least the same throughout the project.

@thijskh do you know why `.pem` is being used in OC-Deploy for the private key and not `.key`?

PR #747, opened 2019-08-27
Set copyright year to 2010 in all docheaders

pablothedude:
The docheader is fixed to 2010 to show the initial year of publication.
In order the view the year of change one could easiliy look into the git
history.


pablothedude:
This is fixed in the SFO branch, so I will close this

PR #748, opened 2019-08-27
Fix the use of SFO on homepage

pablothedude:
On the homepage the SFO term was still being used.

https://www.pivotaltracker.com/story/show/167864360/comments/205827824


pablothedude:
@thijskh thanks, I've changed this

PR #752, opened 2019-09-04
Store coin properties in a single column

pablothedude:
The coin serialization should be available for the next (5.12) release.

The new column is added, but the dedicated fields are kept in place for version roll-back compatibility. We intend to drop these columns in release 5.14 also know as version 6.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ We will continue to post relevant release notes on the GitHub release page. More
 
 More information about our release strategy can be found in the [Development Guidelines](https://github.com/OpenConext/OpenConext-engineblock/wiki/Development-Guidelines#release-notes) on the EngineBlock wiki.
 
+## 5.12.0
+We've changed the way how coin entity properties are stored in the database. As of now, they will be stored in a serialized manner, in a single column. This greatly simplifies adding or removing coins in the future, as this no longer requires database schema changes. 
+
+ *  Store coin properties in a single column #752  
+
 ## 5.11.3
 The footer links have been made configurable to a greater extent in this release.
 
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,21 @@
 # UPGRADE NOTES
 
+## 5.11 > 5.12.0
+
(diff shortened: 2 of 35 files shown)


MKodde:
Aside from the unpolished commit history, this looks good to me. In addition to all the changes, you might want to add the updated release notes in this PR. Also, an entry in the UPGRADING.md might be warranted? Even tho we are not introducing BC breaking changes. The requirement of running the migrations might be a good warning.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ We will continue to post relevant release notes on the GitHub release page. More
 
 More information about our release strategy can be found in the [Development Guidelines](https://github.com/OpenConext/OpenConext-engineblock/wiki/Development-Guidelines#release-notes) on the EngineBlock wiki.
 
+## 5.12.0
+How coins from the metadata are stored in the database is changed. In order to prevent database migrations when introducing or removing features which rely extensively on those coins the coin data is serialized.
```
MKodde: Proposal for change, grammatically these sentences can be more clear/readable:

> We changed the way how coin entity properties are stored in the database. As of now, they will be stored in a serialized manner, in a single column. This greatly simplifies adding or removing coins in the future, as this no longer requires database schema changes.

pablothedude:
I've squashed the commits and updated the changelog and upgrading docs.

UPGRADING.md
```diff
@@ -1,5 +1,21 @@
...
 
+## 5.11 > 5.12.0
+
+### Serialised column cleanup
+In version 5.12.0 multiple `coin` columns are serialised into a single column with serialised data.
+This change will allow easier maintenance because then no migrations are needed when adding or removing a `coin`.
+There is no migration to populate the old `coin` columns to the new serialised column while this is needed to let EB function properly with the new changes. 
+Therefore you should push the data from Manage after you have updated the codebase and ran the `Version20190703235333`.
+
+Be aware that you need to be logged in into manage to push the data after updating the codebase and database schema.
```
MKodde: :100:

PR #755, opened 2019-09-10
Cleanup old coin columns

pablothedude:
The old coin columns could be cleaned now and are no longer needed.

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/AbstractRole.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/AbstractRole.php
@@ -228,30 +228,9 @@ abstract class AbstractRole
     /**
      * @var bool
      *
-     * @ORM\Column(name="disable_scoping", type="boolean")
-     */
-    public $disableScoping;
-
-    /**
-     * @var bool
-     *
-     * @ORM\Column(name="additional_logging", type="boolean")
-     */
-    public $additionalLogging = false;
-
-    /**
-     * @var bool
-     *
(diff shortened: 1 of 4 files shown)


src/OpenConext/EngineBlock/Metadata/Entity/AbstractRole.php
```diff
@@ -225,34 +225,13 @@ abstract class AbstractRole
      */
     public $publishInEduGainDate;
```
MKodde: Is there a reason why the `coin:publish_in_edugain_date` field is not cleaned up? Idem for: `publish_in_edugain`

pablothedude: The `publish_in_edugain` fields are still needed by the MetadataRepositories because the DoctrineMetadataRepository is querying on that column in the roles table. Because I wanted to confine myself to the coin serialization only to keep the release as atomic as possible and there was already an issue to drop Edugain altogether this wasn't done earlier.

We could decide to move the removal of Edugain and the columns into `6.0.0`. I think this would be possible if Edugain is already phased out and we don't need a rollback so those fields could also be cleaned up in this release. This will have the benefit that all column removal currently known of is done at once.

But I think that's for @thijskh to decide. @thijskh what are your thoughts on this?

thijskh: I would clean them up when we work on reimagining the metadata = 6.1

MKodde:
I found a left-over in the Twig templates, fixed that in #756

PR #756, opened 2019-09-18
Get termsOfServiceUrl from coins, not SP metadata entity

MKodde:
After moving the COIN properties from the SP/AbstractRole metadata entity, the termsOfServiceUrl property was not migrated in the twig template.

I've looked through the templates for other (mis)usages of coins in the twig templates. To no avail.

https://www.pivotaltracker.com/story/show/168565818


pablothedude:
It's very strange that the tests weren't failing. I've tested this and the exception is muted if twig isn't in dev mode. I've created a PR for this to reflect this behavior so this could be fixed and be prevented in the future.

`https://github.com/OpenConext/OpenConext-engineblock/pull/758`

pablothedude:
Be aware that this maybe have to be backported to older releases `5.12` `513`.

PR #757, opened 2019-09-18
Enable the visual regression tests for error pages

MKodde:
For now we only ran these tests in the development environment. Now that the error page styling reached a certain level of maturity, enabling these automated tests on Travis builds, should yield good feedback. As these tests are rather fragile, and would have caused many false positives during the development of the error pages styling.

https://www.pivotaltracker.com/story/show/165743005


MKodde:
Hi, @epinxteren I would like your input on our (now broken) visual regression tests. Our setup consists of Jest with Puppeteer test (using Chrome & Selenium). And we use the americanexpress jest-puppeteer snapshot tool in order to compare actual pixels.

The tests aim to create snapshots of different EB error pages. They are found on top of the https://github.com/OpenConext/OpenConext-engineblock/blob/feature/enable-visual-regression-tests/theme/material/javascripts/tests/visual-regression/error-page/ErrorPage.test.js test. We then further set up the test, creating a page for each error page and putting other stuff into the test sets.

The problem we currently run into is that the `page` instance that is set on each 'set'. Is not an actual Puppeteer page instance, but remains undefined. Possibly because the timeout time is reached or some other async related issue...

Do you see what's wrong with our current setup? And more importantly, do you see a way we could improve this test setup. Hopefully making it less fragile.

**Additional information**
Original PR: https://github.com/OpenConext/OpenConext-engineblock/pull/687

The vis-reg tests can be run using the `ant ant js-visual-regression-tests` and `ant ant js-visual-regression-tests-update-snapshots` build targets.

MKodde:
The visual regression tests are now failing on differences between the snapshots generated locally, versus the ones created on Travis.

@pablothedude do you have ideas on how we can fix this without too much effort? I'm starting to feel more like keeping these tests as a dev machine only test tool. As this might significantly decrease the robustness of the tests.

pablothedude:
@MKodde I suggest we wait, if we could use docker for the tests this shouldn't be a problem at all.

PR #758, opened 2019-09-18
Test twig enable debug

pablothedude:
This is a test to check how failures in Twig would result into failing tests. This was done after failing tests in:

https://github.com/OpenConext/OpenConext-engineblock/pull/756

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,7 @@ script:
 branches:
   only:
     - master
+    - feature/serialize-coins-cleanup
 
 after_failure:
   - sudo tail -500 /var/log/apache2/error.log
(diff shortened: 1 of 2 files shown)


MKodde:
Is your conclusion to enable Twig debug in the test environment? That seems like a good idea to me. But this does not fully explain why in your PR, where the coin fields/getters where removed from the SP entity, the build passed. As a missing property or getter on an object would trigger a 500 from the Twig template, it being in debug mode or not.

Or could this be some obscure cache related issue? Where the opcache would still contain the 'old' templates in cache?

MKodde:
Closing this test PR

PR #759, opened 2019-09-23
Use twig to build metadata

pablothedude:
A proof of concept to show we can generate valid SAML metadata for EngineBlock, using Twig


src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        switch (true) {
+            case ($role instanceof ServiceProvider):
+                $template = '@theme/Authentication/View/Metadata/sp.html.twig';
+                break;
+            case ($role instanceof IdentityProvider):
+                $template = '@theme/Authentication/View/Metadata/idp.html.twig';
+                break;
+            default:
+                throw new RuntimeException('invalid role requested');
```
MKodde: I know this is more of a proof of concept. But for when you will solidifying this:

1. Use one of our own exceptions for this, or create one specifically for this scenario.
2. Make the exception message more verbose, and possbily add the requested, invalid role to the message.

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
```diff
@@ -0,0 +1,81 @@
...
+            'wantSignature' => false,
+            'publicKey' => $signingKeyPair->getCertificate()->toCertData(),
+        ];
+        $baseXml = $this->twig->render($template, $params);
+
+        // Sign source
+        $params = $this->addSignature($baseXml, $params, $signingKeyPair->getPrivateKey());
+
+        // Generate signed metadata
+        $params['wantSignature'] = true;
```
MKodde: wantSignature doesn't sound good to me. Maybe using 'signatureRequired' or 'signatureRequested' is better?

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
pablothedude: I wanted to create a method per Role type (SP, IdP) so we don't need any validation or custom exceptions. I think it will be a good practice to use custom exceptions for this so I'll keep your suggestion in mind.

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
pablothedude: I want to create a DTO for this in the end and get rid of the array, but it made my life a bit easier during the POC.

MKodde: Not aiming at the array here but at the naming of the `wantSignature` parameter. When moving this to a DTO please ensure you use a clear method name then!

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
MKodde: At the very least use the [Bundles own RuntimeException](https://github.com/OpenConext/OpenConext-engineblock/blob/b23d71d8df5e8c6c5974273a3f4e5864cd0b4e38/src/OpenConext/EngineBlockBundle/Exception/RuntimeException.php)

MKodde:
@pablothedude I've created a 6.1 release branch. And targetted this PR onto that. The 6.1 branch is based on master WITH the 7.2 compat changes. But other 6.0 changes should still be merged to this branch when they start landing on master.

pablothedude:
But I don't want to merge features into a release branch so I changed the target back to master.

MKodde:
Allright, but do not merge this to master before release 6.0 is completed! My intention was to use the 6.1 release branch as a temporary 6.1 development branch. It might not be fully GitFlow compatible, but with our challenging version landscape, it might not be a bad idea.

pablothedude:
No problem, I will leave it as is for now but maybe we could make a temporary branch if we're not sure to merge it back to master and use that as work branch to prevent the pollution of release branches.

pablothedude:
I've created an `feature/metadata` intermediate branch for this purpose.

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
pablothedude: I did creat two methods, one to create a SP and one to create an IdP so no Exception is needed.

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
pablothedude: I agree this looks like a smell and removed this from twig. The signing is now completely handled by `XMLSecurityDSig` which should be less prone to errors.

MKodde:
Merging this POC as a starting point for our other `reimagine metadata` work. Thanks @pablothedude

PR #760, opened 2019-09-24
Prevent WAYF c-button from floating left

MKodde:
The button, displayed when the edit previously chosen selection editor is activated. Would float to the left when the viewport size decreased to <= medium. Resulting in a not so elegantly rendered list of previously chosen IdP's.

As a solution. The button now remains floated to the right, but still retains his own row. This to ensure the IdP labels are able to use the maximum width of the row.

See: https://www.pivotaltracker.com/story/show/166694766

### DIFF
--- a/theme/material/stylesheets/application/modules/_results.css.sass
+++ b/theme/material/stylesheets/application/modules/_results.css.sass
@@ -163,10 +163,6 @@
     .action
       display: inline-block
 
-      +max-width($medium)
-        margin-right: auto
-        margin-left: 0
-
   // remove last line of last link
   .list:last-child
     a:last-child,


MKodde:
Screenshot_2019-09-24 OpenConext - Select an organisation to login to the service

theme/material/stylesheets/application/modules/_results.css.sass
```diff
@@ -162,10 +162,10 @@
   .show-buttons
     .action
       display: inline-block
-
-      +max-width($medium)
-        margin-right: auto
-        margin-left: 0
+      &:not(.deleteable)
```
pablothedude: Please don't use `deletable` for the styling because this class is used as selector for js logic.
Why not remove the `not` and use a `float: right` always?

MKodde: Thanks for pointing that out! It turned out the entire max-width mixin could be removed. As it was introduced for the request access buttons who later received different behavior altogether. Leaving this block to only bug the close buttons.

pablothedude: :+1:

PR #763, opened 2019-09-25
Update Composer for improved PHP 7.2 support

MKodde:
The Composer PHP requirements where not set correctly yet. They still forced PHP 5. And that is not particularly useful when working with PHP 7.2

https://www.pivotaltracker.com/story/show/168775915

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "source": "https://github.com/OpenConext/OpenConext-engineblock"
     },
     "require": {
-        "php": ">=5.6,<7",
+        "php": ">=7.2",
         "ext-json": "*",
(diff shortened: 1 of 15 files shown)


composer.json
```diff
@@ -11,7 +11,7 @@
         "source": "https://github.com/OpenConext/OpenConext-engineblock"
     },
     "require": {
-        "php": ">=5.6,<7",
+        "php": ">=7.2",
```
pablothedude: Consider a constraint to allow only 7.2.

MKodde: Yes, good point. I pinned the constraint on 7.2 for now.

PR #765, opened 2019-09-26
Integrate metadata factory

MKodde:
The MetadataFactory that was created in the #759 is now used in the MetadataController. 

So this PR can be seen as yet another POC that now proves we can generate the XML metadata without existing Corto services.

Some remarks:

1. I only implemented the SP metadata generation. And hardcoded the EB entityID. This should be fixed, and the entity ID should be set some other way? Maybe from config. (TODO)
2. The MetadataFactory was refactored slightly, as the PublicKey was not set in the sp template. 
3. The MetadataFactory now uses a KeyPairFactory to set its keys, opening up the possibility to more easily set a different key in the case of key rollover. (TODO)

The two TODO items should be addressed at some other point and have been left out ouf this POC

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
+++ b/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
@@ -18,10 +18,13 @@
 
 namespace OpenConext\EngineBlock\Metadata\X509;
 
+use OpenConext\EngineBlock\Assert\Assertion;
 use OpenConext\EngineBlock\Exception\RuntimeException;
 
 class KeyPairFactory
 {
+    const DEFAULT_KEY_PAIR_IDENTIFIER = 'default';
+
     private $keyPairConfiguration = [];
 
     /**
@@ -38,8 +41,10 @@ class KeyPairFactory
      *
      * @throws RuntimeException
      */
-    public function buildFromIdentifier(string $identifier = 'default') : X509KeyPair
+    public function buildFromIdentifier(string $identifier) : X509KeyPair
     {
+        Assertion::nonEmptyString($identifier, 'identifier');
+
         if (array_key_exists($identifier, $this->keyPairConfiguration)) {
             $keys = $this->keyPairConfiguration[$identifier];
             $privateKey = new X509PrivateKey($keys['privateFile']);
--- a/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
+++ b/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
@@ -47,7 +47,7 @@ class KeyPairFactory
             return new X509KeyPair($publicKey, $privateKey);
         }
         throw new RuntimeException(
-            sprintf('Unable to find the encryption keypair identified by "%s"', $identifier)
+            sprintf('Unable to find the encryption key pair identified by "%s"', $identifier)
         );
     }
 }
--- a/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
+++ b/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
@@ -1,0 +1,53 @@
+<?php declare(strict_types=1);
+
+/**
+ * Copyright 2010 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlock\Metadata\X509;
+
+use OpenConext\EngineBlock\Exception\RuntimeException;
+
+class KeyPairFactory
+{
+    private $keyPairConfiguration = [];
+
+    /**
+     * @param array $keyPairConfiguration
+     */
+    public function __construct(array $keyPairConfiguration)
+    {
+        $this->keyPairConfiguration = $keyPairConfiguration;
+    }
+
+    /**
+     * @param string $identifier
+     * @return X509KeyPair
+     *
+     * @throws RuntimeException
+     */
+    public function buildFromIdentifier(string $identifier = 'default') : X509KeyPair
+    {
+        if (array_key_exists($identifier, $this->keyPairConfiguration)) {
+            $keys = $this->keyPairConfiguration[$identifier];
+            $privateKey = new X509PrivateKey($keys['privateFile']);
+            $publicKey = new X509Certificate(openssl_x509_read(file_get_contents($keys['publicFile'])));
+            return new X509KeyPair($publicKey, $privateKey);
+        }
+        throw new RuntimeException(
+            sprintf('Unable to find the encryption keypair identified by "%s"', $identifier)
+        );
+    }
+}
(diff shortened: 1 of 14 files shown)


src/OpenConext/EngineBlockBundle/Metadata/Service/MetadataService.php
```diff
@@ -0,0 +1,46 @@
...
+
+use OpenConext\EngineBlock\Metadata\Entity\AbstractRole;
+
+interface MetadataService
+{
+    /**
+     * Set the optional key (for key rollover)
+     * @param string $keyId
+     */
+    public function setKeyId(string $keyId): void;
```
pablothedude: I think this method should be omitted as it will set state.

src/OpenConext/EngineBlockBundle/Metadata/Service/MetadataService.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Generate XML metadata for a role (either IdP or SP)
+     *
+     * @param AbstractRole $role
+     * @return string
+     */
+    public function metadataFrom(AbstractRole $role): string;
```
pablothedude: I would merge the two methods together and use the entityId as parameter instead the role and also add the used keyId as parameter.
`public function metadataFrom(string $entityId, string $keyId): string;`

src/OpenConext/EngineBlockBundle/Metadata/Service/MetadataService.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Find the IdP or SP entity associated with the provided entityId
+     *
+     * @param string $entityId
+     * @return AbstractRole
+     */
+    public function getRoleByEntityId(string $entityId) : AbstractRole;
```
pablothedude: I think this method could be omitted

src/OpenConext/EngineBlockBundle/Metadata/Service/ServiceProviderMetadataService.php
```diff
@@ -0,0 +1,86 @@
...
+namespace OpenConext\EngineBlockBundle\Metadata\Service;
+
+use OpenConext\EngineBlock\Metadata\Entity\AbstractRole;
+use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
+use OpenConext\EngineBlock\Metadata\MetadataRepository\MetadataRepositoryInterface;
+use OpenConext\EngineBlockBundle\Exception\EntityCanNotBeFoundException;
+use OpenConext\EngineBlockBundle\Metadata\MetadataFactory;
+use Webmozart\Assert\Assert;
+
+class ServiceProviderMetadataService implements MetadataService
```
pablothedude: See related remarks in MetadataServiceInterface

src/OpenConext/EngineBlockBundle/Metadata/Service/MetadataService.php
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlockBundle\Metadata\Service;
+
+use OpenConext\EngineBlock\Metadata\Entity\AbstractRole;
+
+interface MetadataService
```
pablothedude: Why not use `MetadataServiceInterface` as name ?

theme/material/templates/modules/Authentication/View/Metadata/sp.html.twig
```diff
@@ -13,9 +12,7 @@
         <md:KeyDescriptor use="signing">
             <ds:KeyInfo xmlns:ds="http://www.w3.org/2000/09/xmldsig#">
                 <ds:X509Data>
-                    <ds:X509Certificate>
-                        {{ publicKey }}
-                    </ds:X509Certificate>
+                    <ds:X509Certificate>{{ publicKey }}</ds:X509Certificate>
```
pablothedude: This should be the certificates of the entity itself. We could use a `MetadataDto` for the conversion.

MKodde:
>     3. For now this is a very good solution. Maybe we should rename the `MetadataFactory` to `MetadataRenderer` as it will describe better what it does. Maybe we could do this in the end to prevent merge conflicts. What are your thoughts on this?

Your rename suggestion is fine with me! We can also move this logic to the EngineBlock namespace. It now resides in the EB Bundle, and that should be reserved for infrastructural EngineBlock code. The metadata code, in my opinion, is more at home in the application layer.

src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->keyPairConfiguration = $keyPairConfiguration;
+    }
+
+    /**
+     * @param string $identifier
+     * @return X509KeyPair
+     *
+     * @throws RuntimeException
+     */
+    public function buildFromIdentifier(string $identifier = 'default') : X509KeyPair
```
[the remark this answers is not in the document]
MKodde: Addressed this in commit: 24f9f1947

src/OpenConext/EngineBlockBridge/ResponseFactory.php
```diff
@@ -36,4 +36,11 @@ public static function fromEngineBlockResponse(EngineBlock_Http_Response $respon
 
         return new Response($response->getBody(), $response->getStatusCode(), $response->getHeaders());
     }
+
+    public static function fromXml(string $metadataFrom)
+    {
+        $response = new Response($metadataFrom);
```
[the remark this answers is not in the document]
MKodde: Addressed this in commit: 24f9f1947

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
```diff
@@ -77,6 +79,7 @@ private function renderMetadataXml(AbstractRole $role, $template)
         $params = [
             'id' => $this->samlIdGenerator->generate(self::ID_PREFIX, EngineBlock_Saml2_IdGenerator::ID_USAGE_SAML2_METADATA),
             'entity' => $role,
+            'publicKey' => $this->signingKeyPair->getCertificate()->toCertData()
```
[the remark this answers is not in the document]
MKodde: Yes, let's look into this in an upcoming PR

theme/material/templates/modules/Authentication/View/Metadata/sp.html.twig
MKodde: https://github.com/OpenConext/OpenConext-engineblock/pull/765#discussion_r329546880

src/OpenConext/EngineBlockBundle/Controller/MetadataController.php
```diff
@@ -55,25 +66,29 @@ public function idpMetadataAction($keyId = null)
 
     /**
      * @param null|string $keyId
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse|\Symfony\Component\HttpFoundation\Response
+     * @return Response
      */
     public function spMetadataAction($keyId = null)
```
pablothedude: Why not require always require a `keyId` and typehint it as `string`? We could then set a default value in the routing configuration. Feel free to leave it as-is.

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
```diff
@@ -72,11 +82,17 @@ public function fromIdentityProviderEntity(IdentityProvider $role)
         return $signedXml;
     }
 
+    public function setKey(string $keyId)
```
pablothedude: Could this not be better added as parameter to the `from*Entity` method because this method should always be called before calling that method.

src/OpenConext/EngineBlockBundle/Controller/MetadataController.php
MKodde: Addressed in 5b17234

src/OpenConext/EngineBlockBundle/Metadata/MetadataFactory.php
MKodde: Addressed in 5680937

PR #767, opened 2019-09-30
Remove eduGAIN support from project

MKodde:
EduGain metadata generation support is removed from the project. This includes the home made md-rpi support.

**Todo & Discussion**
* Include migration to drop the edugain columns. Or do we need to remove this in a more rolling update friendly manner? In that case we want to keep these columns and revert some of the changes to the metadata assembler.


MKodde:
@thijskh I suppose we should handle the removal of the edugain metadata fields from the roles table like we did for the previous upgrades? In short:

1. Remove the usages from code in release A (6.1), but keep provisioning them in the metadata push assembler.
2. Actually remove the column from the database in release B (6.2).

Or is the use case different for the Edugain metadata feature? As this was disabled from a UI point of view some releases ago (5.10 from the top of my head).

thijskh:
It is fine to drop them if nothing breaks in 5.13 outside of requesting the specific feed. If unsure, the safe thing is to drop later.

library/EngineBlock/Corto/Mapper/Metadata/EdugainDocument.php
```diff
@@ -18,24 +18,25 @@
 
 use OpenConext\EngineBlock\Metadata\Entity\AbstractRole;
 
+/**
+ * TODO: Not yet removed with the story: https://www.pivotaltracker.com/story/show/164925138
+ * this should have been made obsolete once all Metadata related logic is moved out of Corto
+ */
 class EngineBlock_Corto_Mapper_Metadata_EdugainDocument
```
pablothedude: Could we rename the class to prevent confusion?

src/OpenConext/EngineBlock/Metadata/Entity/AbstractRole.php
```diff
@@ -172,6 +172,8 @@ abstract class AbstractRole
     /**
      * @var bool
      *
+     * @deprecated: This coin is no longer used in EngineBlock and will be removed in release 6.2
+     *
      * @ORM\Column(name="publish_in_edugain", type="boolean")
      */
     public $publishInEdugain;
```
pablothedude: Do you have this in scope or would you be so kind to open a PR for this so we won't forget this?

src/OpenConext/EngineBlock/Metadata/MetadataRepository/CachedDoctrineMetadataRepository.php
```diff
@@ -208,12 +208,4 @@ public function findReservedSchacHomeOrganizations()
     {
         return $this->invoke(__FUNCTION__, func_get_args());
     }
-
-    /**
-     * @return AbstractRole[]
-     */
-    public function findEntitiesPublishableInEdugain()
-    {
-        return $this->invoke(__FUNCTION__, func_get_args());
```
pablothedude: What a treat to remove this :cake:

library/EngineBlock/Corto/Mapper/Metadata/EdugainDocument.php
MKodde: Yes we could. But we are removing it in the very near future (this sprint) So for now I'm opting to leave it be.

src/OpenConext/EngineBlock/Metadata/Entity/AbstractRole.php
MKodde: Yup, wilco!

PR #768, opened 2019-10-01
Remove idp metadata with sp-entity-id from home

MKodde:
The EB homepage advertised the possibility to open IdP metadata with the certificate of a specified SP. This feature was not longer supported in Engine. At least, I was not able to reproduce the behavior described for this feature.

There was not a lot to clean up, so only the homepage links have been removed.

See: https://www.pivotaltracker.com/story/show/168249086

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/FrontPage.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/FrontPage.feature
@@ -1,0 +1,44 @@
+Feature:
+  In order to inform users of EngineBlock's capabilities
+  As EngineBlock
+  I want to display a front page with links to my metadata endpoints and other features
+
+  Background:
+    Given an EngineBlock instance on "vm.openconext.org"
+      And no registered SPs
+      And no registered Idps
+
+  Scenario: The user can see all available metadata links
+    When I go to Engineblock URL "/"
+    Then I should see 15 links on the front page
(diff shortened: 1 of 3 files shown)


MKodde:
Nope, there is a clear distinction between IdP and IdPs metadata. The first is no longer used in context with the sp-entity-id, the latter is used for wayf construction.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/FrontPage.feature
```diff
@@ -0,0 +1,43 @@
...
+  In order to inform users of EngineBlock's capabilities
+  As EngineBlock
+  I want to display a front page with links to my metadata endpoints and other features
+
+  Background:
+    Given an EngineBlock instance on "vm.openconext.org"
+      And no registered SPs
+      And no registered Idps
+
+  Scenario: The user can see all available metadata links
```
pablothedude: Consider splitting the tests the way the frontpage is grouped.

Consider asserting the number of links to detect if entries are added. That way someone will need to fix this and is notified of the existence of the tests.

pablothedude:
A good call to add the test!

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/FrontPage.feature
MKodde: Good points!
1. Updated the display order of the step definitions to match that of the page
2. Added an expected number of links assertion

Issue #211, opened 2015-11-10
More friendly error message when fetching IdPs-metadata for non-existing SP


baszoetekouw:
When requesting IdPs-metadata for a non-existing SP, Engine currently returns "error an error occurred: It is not possible to sign in. Please try again."

```
curl -v 'https://engine.test.surfconext.nl/authentication/proxy/idps-metadata?sp-entity-id=urn:bestaat:niet'
[...]
> GET /authentication/proxy/idps-metadata?sp-entity-id=urn:bestaat:niet HTTP/1.1
> Host: engine.test.surfconext.nl
> User-Agent: curl/7.44.0
> Accept: */*
>
< HTTP/1.1 500 Internal Server Error
< Server: nginx/1.8.0
< Date: Tue, 10 Nov 2015 13:56:44 GMT
< Content-Type: text/html; charset=UTF-8
[...]
 <div class="mod-content">
    <h1>Error - An error occurred</h1>
    <p>
        It is not possible to sign in. Please try again.
    </p>
```

The logs show:

```
Nov 10 14:56:44 t03 EBLOG[31074]: [2015-11-10 14:56:44] engineblock.INFO: Handling incoming request: GET /authentication/proxy/idps-metadata {"request_id":"5641f79c53af1","session_id":null} []
Nov 10 14:56:44 t03 EBLOG[31074]: [2015-11-10 14:56:44] engineblock.INFO: Calling service 'idpsMetadataService' {"request_id":"5641f79c53af1","session_id":"KPrgY6VPRGotIZKCfNMO-W2,HFf"} []
[...]

PR #770, opened 2019-10-01
Show custom error when metadata not found

MKodde:
The metadata actions where results can be modified using an SP entity ID would result in a generic 500 error. This is not terribly user-friendly. And this PR aims to make the error more clear to the end user by
providing a custom error message. This message is partly built up out of the exception message (which states what entity id was referenced) and a custom error message (translatable).

See: https://www.pivotaltracker.com/story/show/168258776


MKodde:
Translations might need to be revised. As the PM's have been asked for input and the Portuguese translation is missing altogether. This will later be added to this PR, or addressed in a new one. Please do not refrain from reviewing this ticket.

MKodde:
I do not agree. It might not be required for the precise use case I implemented this in. But the IdP's metedata endpoint filtered by a specific entity Id (see original issue #211 for more background info). The exception I introduced should be thrown when that specific entity can not be found. And voila.

pablothedude:
I missed the additional context, thanks for pointing that out!

PR #773, opened 2019-10-07
Create metadata entity factory

pablothedude:
A uniform entity factory which will return immutable entities is
required to enforce immutable state in entities. This will also prevent
a proliferation of factories to initiate different entities.

https://www.pivotaltracker.com/story/show/168980646

### DIFF
--- a/src/OpenConext/EngineBlock/Entities/Adapter/IdentityProviderEntity.php
+++ b/src/OpenConext/EngineBlock/Entities/Adapter/IdentityProviderEntity.php
@@ -1,0 +1,259 @@
+<?php declare(strict_types=1);
+/**
+ * Copyright 2010 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlock\Entities\Adapter;
+
+use DateTime;
+use OpenConext\EngineBlock\Entities\IdentityProviderEntityInterface;
+use OpenConext\EngineBlock\Metadata\Coins;
+use OpenConext\EngineBlock\Metadata\ConsentSettings;
+use OpenConext\EngineBlock\Metadata\ContactPerson;
+use OpenConext\EngineBlock\Metadata\Entity\IdentityProvider;
+use OpenConext\EngineBlock\Metadata\Logo;
+use OpenConext\EngineBlock\Metadata\Organization;
+use OpenConext\EngineBlock\Metadata\Service;
+use OpenConext\EngineBlock\Metadata\ShibMdScope;
+use OpenConext\EngineBlock\Metadata\X509\X509Certificate;
+
+class IdentityProviderEntity implements IdentityProviderEntityInterface
+{
+    /**
+     * @var IdentityProvider
+     */
+    private $entity;
+
+    public function __construct(IdentityProvider $entity)
+    {
+        $this->entity = $entity;
+    }
+
+    /**
+     * @return int
+     */
+    public function getId(): int
+    {
+        return $this->entity->id;
+    }
+
+    /**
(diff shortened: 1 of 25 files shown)


src/OpenConext/EngineBlock/Entities/Adapter/IdentityProviderEntity.php
```diff
@@ -0,0 +1,260 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlock\Entities\Adapter;
+
+
+use DateTime;
```
MKodde: Chore for the OpenConext\EngineBlock\Entities\**\* files:
* Remove excessive whitespaces
* Remove unused imports
* Add meaningful doc blocks for the entities and factories

pablothedude:
>     1. The new Entity decorator setup looks great. It will help us to use the entities in the context they are intended for in each use case we have for them. Some thoughts:
>        
>        1. We should somehow force the use of the factory, or state clearly to NOT use the entities directly.
I agree but currently I didn't bother to document the classes to be able to speed up the POC. My intention is to add this after the architecture is finished.
>        2. The *Proxy entity classes are in need of extra explanation. It might not be clear to a user that these entities are used in the context of EngineBlock as a proxy. Most confusion will be caused by when is an entity considered a Proxy and when is it a regular entity.
I agree, we currently have the benefit that we can now use different enitities for the specific situations. I think it will be valuable to add this as documentation in the proxy classes too.
>        3. The Doctrine entities (`OpenConext/EngineBlock/Metadata/Entity/*`) and the Entities namespace (`OpenConext/EngineBlock/Entities`) might need to be merged into one? Both are not exclusivly used for displaying Metadata. I'd opt for moving them to the newly created namespace.
I agree it would be better to have them at one place but I think it would be clear that the new `Entities` should be used in favor of the mutable counterparts.
>        4. What exactly will the Adapter entities be used for?
The adapter entities are used to convert the old mutable entities to the new immutable entities. That way we have BC support but it would be very easy to get rid of the old entities by only confining to the `EntityInterface`.



>     2. Regarding the Factory
>        
>        1. The logic from: `src/OpenConext/EngineBlockBundle/Metadata/MetadataEntityFactory.php` (c|s)hould be merged into the `OpenConext\EngineBlock\Entities\Adapter` factories.
Personally I would opt to get rid of the `MetadataEntityFactory` in favor of the new factory. While you could now use `\OpenConext\EngineBlock\Entities\Factory\ServiceProviderFactory::createMinimalEntity` for this.
>        2. The names of the factory methods are not perfectly clear upon first glance. For example. Why would one want to create a minimal entity? Or what does: `createStepupFromEntity` mean?
For internal use we only need a entity with minimal values set  (entityId, sso/acs, certificate). This will be sufficient. If you have a better suggestion for the name of the methods I'll be happy if you'd share that with me.
>     3. Can you try to update the PR's initial comment to more closely show the intent of this PR?
Will do!
>     4. Please remove the commits from `feature/metadata-refactor-old` as soon as possible. Basing this branch on the `feature/metadata-refactor-old` might be much easier. As you could easily rebase any changes on the base branch onto your feature branch?
Personally I want the changes of the `feature/metadata-refactor-old` in the metadata branch.
>     5. Please remove [ae52e05](https://github.com/OpenConext/OpenConext-engineblock/commit/ae52e05366077c4e492091af483d207c836a85ec) from this branch
Will do if the tests are green.

MKodde:
RE 1.1& 1.2:  I'm fine with that, please add this as a task or open a new ticket
RE 1.3: Yes, in a perfect world we would phase out the mutability, but moving the entities to one namespace does still seem like a good idea to me.
RE 1.4: Ack! Please add this as documentation on the class, or mention this as a task
RE 2.1: Yes it seems that factory has been made obsolete. The way keys are used in that factory could be moved to the new ones.
RE 2.2: Personally I would leave the part after From away from the name. Leaving you with: `createEntityFrom(ServiceProvider $entity)` Which reads a little bit nicer. Applies for the other factory methods too. And again the use of the Proxy term needs clearification. This will be part of the documentation chore I presume?
RE 2.3: Ack!
RE 2.4: Ack! Personally I would have gone for the rebase route, but this works too.
RE 2.5: Ack!

pablothedude:
For the documentation I've create a task n the issue to create this factory solution.

2.2 I don't like `createEntityFrom` more because if you want to create an instance from something else you can't use `createEntityFrom` multiple times but with different parameters. I want to keep it for now as-is till we find something better. Added this as task to the refactor issue.

PR #775, opened 2019-10-08
Add missing documentation to EntityFactory

pablothedude:
The EntityFactory architecture was created during a POC but
was lacking documentation. This is added to improve the
inline documentation of the architecture.

This is to address:
https://github.com/OpenConext/OpenConext-engineblock/pull/773#issuecomment-539434966

https://www.pivotaltracker.com/story/show/168980646

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/IdentityProvider.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/IdentityProvider.php
@@ -31,14 +31,13 @@ use RobRichards\XMLSecLibs\XMLSecurityKey;
 use SAML2\Constants;
 
 /**
- * Class IdentityProvider
  * @package OpenConext\EngineBlock\Metadata\Entity
  * @ORM\Entity
  * @SuppressWarnings(PHPMD.CamelCasePropertyName)
  * @SuppressWarnings(PHPMD.ExcessiveParameterList)
  *
- *
- * PLEASE DON'T USE THIS ENTITY DIRECTLY BUT USE `\OpenConext\EngineBlock\Factory\Factory\IdentityProviderFactory`
+ * WARNING: Please don't use this entity directly but use the dedicated factory instead.
+ * @see \OpenConext\EngineBlock\Factory\Factory\IdentityProviderFactory
  */
 class IdentityProvider extends AbstractRole
 {
@@ -86,7 +85,8 @@ class IdentityProvider extends AbstractRole
     public $shibMdScopes = array();
 
     /**
-     * PLEASE DON'T USE THIS ENTITY DIRECTLY BUT USE `\OpenConext\EngineBlock\Factory\Factory\IdentityProviderFactory`
+     * WARNING: Please don't use this entity directly but use the dedicated factory instead.
+     * @see \OpenConext\EngineBlock\Factory\Factory\IdentityProviderFactory
      *
      * @param string $entityId
      * @param Organization $organizationEn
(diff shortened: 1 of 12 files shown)


src/OpenConext/EngineBlock/Metadata/Entity/IdentityProvider.php
```diff
@@ -36,6 +36,9 @@
  * @ORM\Entity
  * @SuppressWarnings(PHPMD.CamelCasePropertyName)
  * @SuppressWarnings(PHPMD.ExcessiveParameterList)
+ *
+ *
+ * PLEASE DON'T USE THIS ENTITY DIRECTLY BUT USE `\OpenConext\EngineBlock\Factory\Factory\IdentityProviderFactory`
```
MKodde: Thanks for adding this warning. Maybe we could diminish the tone of voice slightly? Proposal:

> WARNING: Please don't use this entity directly but use the dedicated factory instead.
> @see \OpenConext\EngineBlock\Factory\Factory\IdentityProviderFactory

src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
```diff
@@ -29,6 +29,10 @@
 use OpenConext\EngineBlock\Metadata\ShibMdScope;
 use OpenConext\EngineBlock\Metadata\X509\X509Certificate;
 
+/**
+ * This Entity is the immutable counterpart of the legacy Entity
+ * This adapter is used to support the old legacy entity so it would be easier to remove the legacy entity ultimately
+ */
```
MKodde: I'm not sure we should call the Doctrine entity a legacy entity. Maybe explicitly call it a Doctrine ORM entity and explain we use it throughout the application, but we aim to start using immutable dedicated entities instead.

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/AbstractIdentityProvider.php
```diff
@@ -27,6 +27,10 @@
 use OpenConext\EngineBlock\Metadata\ShibMdScope;
 use OpenConext\EngineBlock\Metadata\X509\X509Certificate;
 
+/**
+ * This Abstract class is used to circumvent the implementation of all methods of the EntityInterface
+ * so only the wanted methods could be implemented if this abstract class is extended
```
MKodde: Suggestion to change:

> This Abstract class is used to circumvent the implementation of all methods of the EntityInterface so only the wanted methods could be implemented if this abstract class is extended

Into:

> This abstract class is used to circumvent the implementation of all methods of the EntityInterface. So only the methods required for the specific implementation have to be created on the decorated Entity that is extended from this abstract IdP entity.

MKodde: Also applies for the SP counterpart

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/IdentityProviderProxy.php
```diff
@@ -21,6 +21,10 @@
 use OpenConext\EngineBlock\Metadata\X509\X509KeyPair;
 use SAML2\Constants;
 
+/**
+ * This decorator is used for using an entity when EB is used as authentication proxy
```
MKodde: Missing `.`

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ServiceProviderStepup.php
```diff
@@ -17,6 +17,10 @@
 
 namespace OpenConext\EngineBlock\Factory\Decorator;
 
+/**
+ * This decorator is used for using an entity when EB is doing a stepup callout during authentication
```
MKodde: ~decorator~ decoration

~for using an entity~ as an SP entity representation of the SFO SP entity

pablothedude:
@MKodde thanks for the commit!

PR #776, opened 2019-10-09
Display the actual EB SP/IdP metadata in the generated XML

MKodde:
This PR aims to display the actual SP and IdP metadata on the `authentication/sp/metadata` and `authentication/idp/metadta` endpoints.

Actual data is data retrieved from the EngineBlock ini config. The database entry for the EB SP and IdP are no longer used. We lose some details with this, for example the support contacts are a little less specific than they used to. But this is acceptable as approved by our PM's. For details see the Pivotal ticket.

Some additional chores have been addressed in this PR:

1. The XML document signing logic was extracted from the MetadataRenderer
2. TimeProvider was moved from the Corto `library` into the Symfony app

https://www.pivotaltracker.com/story/show/168258709


src/OpenConext/EngineBlockBundle/Metadata/ValueObjects/ServiceProviderMetadata.php
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-namespace OpenConext\EngineBlock\Xml\ValueObjects;
+namespace OpenConext\EngineBlockBundle\Metadata\ValueObjects;
```
pablothedude: Is this the desired namespace?

tests/unit/OpenConext/EngineBlock/Xml/test.pem.crt
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
pablothedude: Be aware that this certificate will expire on `May 11 13:22:16 2024 GMT`. If not using the default one consider adding this as a comment to the test.

theme/material/templates/modules/Authentication/View/Metadata/sp.xml.twig
```diff
@@ -0,0 +1,34 @@
+<md:EntityDescriptor xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" ID="{{ id }}" entityID="{{ metadata.entityId }}" validUntil="{{ metadata.validUntil }}">
```
pablothedude: Please rename the original file instead deleting the old one and creating anew one so version history will stay intact and your changes could be reviewed easier.

theme/material/templates/modules/Authentication/View/Metadata/idp.xml.twig
```diff
@@ -0,0 +1,27 @@
+{% spaceless %}
+<md:EntityDescriptor xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" ID="{{ id }}" entityID="{{ metadata.entityId }}" validUntil="{{ metadata.validUntil }}">
```
pablothedude: Please rename the original file instead deleting the old one and creating anew one so version history will stay intact and your changes could be reviewed easier.

src/OpenConext/EngineBlock/Metadata/Factory/Factory/ServiceProviderFactory.php
```diff
@@ -18,65 +18,157 @@
...
 
-    public function createProxyFromEntity(ServiceProvider $entity): ServiceProviderEntityInterface
-    {
-        return new ServiceProviderProxy($this->createEntityFromEntity($entity), $this->proxyKeyPair, $this->attributes, $this->consentService);
+    public function createEngineBlockEntityFrom(
+        string $entityId,
+        string $acsLocation,
+        string $keyId
+    ): ServiceProviderEntityInterface {
+        $entity = $this->buildServiceProviderEntity($entityId, $acsLocation, $keyId);
```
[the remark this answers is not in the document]
MKodde: Fixed in: 56db4ac1

tests/unit/OpenConext/EngineBlock/Xml/test.pem.crt
MKodde: What do you consider to be the 'default one'? The test suite holds different certificates.

Good point to add a warning about the expiration date.

theme/material/templates/modules/Authentication/View/Metadata/sp.xml.twig
MKodde: I do not think it is very valuable to put extra effort into this. The files have been added in this development effort (as in this sprint) and not much has happened in them that would benefit from a complete history.

I hope you agree with this

theme/material/templates/modules/Authentication/View/Metadata/idp.xml.twig
MKodde: https://github.com/OpenConext/OpenConext-engineblock/pull/776#discussion_r334964778

theme/material/templates/modules/Authentication/View/Metadata/sp.xml.twig
pablothedude: I'm OK with it if you don't fix it now but this would have cost no effort at all if this was done initially. Also this now will look like the initial version while it actually isn't. It would also have helped me a lot when reviewing your PR.

It's just something you could consider next time.

tests/unit/OpenConext/EngineBlock/Xml/test.pem.crt
pablothedude: 1. With the default one I meant the one in `/tests/resources/key/*`.
2. :+1:

src/OpenConext/EngineBlock/Metadata/Factory/Factory/ServiceProviderFactory.php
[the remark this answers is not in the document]
pablothedude: Nice to have the EngineBlockConfiguration VO!

I only don't understand why you didn't create a separate Decorator for the metadata EB configuration, for example a EBSupportDecorator, so the state of the entity itself stays immutable? Or else on the Metadata decorator itself if it's needed everywhere.

src/OpenConext/EngineBlockBundle/Metadata/ValueObjects/ServiceProviderMetadata.php
MKodde: Forgot to remove them, they have been replaced by `OpenConext\EngineBlock\Metadata\Factory\` entities

PR #778, opened 2019-10-10
Add missing trusted proxy to stepup callout

pablothedude:
When configuring stepup functionality the trusted proxy wasn't taken into account. So only the configuration of the trusted proxy was used and not the configuration of the proxied SP.

This need to be cherrypicked to release 5.13 and included in release 6.1

https://www.pivotaltracker.com/story/show/169033643

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
+++ b/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
@@ -136,6 +136,12 @@ class EngineBlock_Corto_Module_Service_AssertionConsumer implements EngineBlock_
             $receivedResponse
         );
 
+        // When dealing with an SP that acts as a trusted proxy, we should use the proxying SP and not the proxy itself.
+        if ($sp->getCoins()->isTrustedProxy()) {
+            // Overwrite the trusted proxy SP instance with that of the SP that uses the trusted proxy.
+            $sp = $this->_server->findOriginalServiceProvider($receivedRequest, $log);
+        }
+
         // Goto consent if no Stepup authentication is needed
         if (!$this->_stepupGatewayCallOutHelper->shouldUseStepup($idp, $sp)) {
             $this->_server->sendConsentAuthenticationRequest($receivedResponse, $receivedRequest, $currentProcessStep->getRole(), $this->getAuthenticationState());
--- a/library/EngineBlock/Corto/Module/Service/StepupAssertionConsumer.php
+++ b/library/EngineBlock/Corto/Module/Service/StepupAssertionConsumer.php
@@ -167,6 +167,12 @@ class EngineBlock_Corto_Module_Service_StepupAssertionConsumer implements Engine
                 $idp = $this->_server->getRepository()->fetchIdentityProviderByEntityId($originalReceivedResponse->getIssuer());
                 $sp = $this->_server->getRepository()->fetchServiceProviderByEntityId($originalReceivedRequest->getIssuer());
 
+                // When dealing with an SP that acts as a trusted proxy, we should use the proxying SP and not the proxy itself.
+                if ($sp->getCoins()->isTrustedProxy()) {
+                    // Overwrite the trusted proxy SP instance with that of the SP that uses the trusted proxy.
+                    $sp = $this->_server->findOriginalServiceProvider($receivedRequest, $log);
+                }
+
                 if ($this->_stepupGatewayCallOutHelper->allowNoToken($idp, $sp)) {
 
                     $log->warning('Allow no token allowed from sp/idp configuration, continuing', array('key_id' => $receivedRequest->getId(), 'result' => $e->getFeedbackInfo()));
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
@@ -11,6 +11,7 @@ Feature:
       And a Service Provider named "SSO-SP"
       And an Identity Provider named "Dummy-IdP"
       And a Service Provider named "Dummy-SP"
+      And a Service Provider named "Proxy-SP"
 
   Scenario: Stepup authentication should be supported if set through SP configuration
     Given the SP "SSO-SP" requires Stepup LoA "http://test.openconext.nl/assurance/loa2"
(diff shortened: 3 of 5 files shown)


library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -136,8 +136,16 @@ public function serve($serviceName, Request $httpRequest)
             $receivedResponse
         );
 
+        // When dealing with an SP that acts as a trusted proxy, we should perform the callout decision on the proxied SP and not the
+        // proxy itself.
+        $originalSp = $sp;
```
MKodde: I'm concerned about the `$sp` variable not being terminated after this line. It can still be used, but it should probably not! My preferred solution is to just override the `$sp` (like we do in the SSO service) and not introduce a new `$originalSp` var.

library/EngineBlock/Corto/Module/Service/StepupAssertionConsumer.php
```diff
@@ -167,7 +167,15 @@ private function handleInvalidGatewayResponse(EngineBlock_Corto_Exception_Receiv
                 $idp = $this->_server->getRepository()->fetchIdentityProviderByEntityId($originalReceivedResponse->getIssuer());
                 $sp = $this->_server->getRepository()->fetchServiceProviderByEntityId($originalReceivedRequest->getIssuer());
 
-                if ($this->_stepupGatewayCallOutHelper->allowNoToken($idp, $sp)) {
+                // When dealing with an SP that acts as a trusted proxy, we should perform the callout decision on the proxying SP and not the
+                // proxy itself.
+                $originalSp = $sp;
```
MKodde: See my previous concern. Either `unset($sp)` or overwrite it.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
```diff
@@ -97,3 +98,33 @@ Feature:
       Then I should see "Error - Unknown strong authentication failure"
         And the url should match "/feedback/stepup-callout-unknown"
         And the response status code should be 400
+
+    # Trusted proxy logic
+    Scenario: Stepup authentication should be supported for trusted proxy if set in proxied SP configuration
+        Given the SP "SSO-SP" requires Stepup LoA "http://test.openconext.nl/assurance/loa2"
+        And SP "Proxy-SP" is authenticating for SP "SSO-SP"
+        And SP "Proxy-SP" is a trusted proxy
+        And SP "Proxy-SP" signs its requests
```
MKodde: Why not make this part of the background task? It wouldn't harm the other tests I suppose.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
```diff
@@ -97,3 +98,33 @@ Feature:
       Then I should see "Error - Unknown strong authentication failure"
         And the url should match "/feedback/stepup-callout-unknown"
         And the response status code should be 400
+
+    # Trusted proxy logic
+    Scenario: Stepup authentication should be supported for trusted proxy if set in proxied SP configuration
```
MKodde: Please rephrase this sentence. It confuses me. Suggestion:

> Step-up authentication should be requested for the proxied SP when using a trusted proxy setup

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
```diff
@@ -97,3 +98,33 @@ Feature:
...
+        When I log in at "Proxy-SP"
+        And I select "SSO-IdP" on the WAYF
+        And I pass through EngineBlock
+        And I pass through the IdP
+        And Stepup will successfully verify a user
+        And I give my consent
+        And I pass through EngineBlock
+        Then the url should match "/functional-testing/Proxy-SP/acs"
+
+    Scenario: Stepup authentication should handle stepup for trusted proxy if LoA level is not met but no token is allowed in proxied SP configuration
```
MKodde: Please rephrase this scenario along the lines of my suggestion stated above.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
pablothedude: I don't agree that unsetting this would make anything more clear as garbage collection would do that normally for you and this even would imply that this is actually needed which will make things even more unclear.

I also think that there is a clear distinction between a SP we enter with and an SP which is the actual source of the request. I don't think renaming it to `sp` will make the process more transparent. Also there are throughout the project a lot variables you could use but you presumably should not use.

library/EngineBlock/Corto/Module/Service/StepupAssertionConsumer.php
pablothedude: See my answer above

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
pablothedude: It's only needed for the trusted proxy logic so I would like to keep it there because it's not needed to support the callout.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
pablothedude: Rephrased it, but added a part that it should ponly be used if configured in the proxied sp. Can you live with that?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
pablothedude: Done

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
pablothedude: Discussed IRL and I will change it for now to sp but iI think it would be clearer if some of the logic could be split up  in two steps:
* find actor sp
* process actor sp

MKodde: Conclusion of IRL discussion:

> I don't agree that unsetting this would make anything more clear as garbage collection would do that normally for you and this even would imply that this is actually needed which will make things even more unclear.

My intent for unsetting was not to improve performance, but to prevent further use of the variable.

> I also think that there is a clear distinction between a SP we enter with and an SP which is the actual source of the request. I don't think renaming it to sp will make the process more transparent. Also there are throughout the project a lot variables you could use but you presumably should not use.

We agreed on making this logic the same as the existing SSO setup. An we'll open a TD ticket to review how we can deal with trusted proxies in a more elegant manner.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
MKodde: Alright with me!

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
MKodde: I think the word `supported` is causing some of the confusion with me. EB supports second factor auth, but in the context of a scenario thats already assumed. Thats why I suggested `Step-up authentication should be requested`

I'm okay with the way it is now, but just to make my intention clear.

pablothedude: That was already changed

PR #781, opened 2019-10-15
Move the dev cache location to /tmp/eb_dev_cache

MKodde:
This to improve the application performance when clearing and warming up the dev caches.

https://www.pivotaltracker.com/story/show/169144846

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -62,6 +62,12 @@ class AppKernel extends Kernel
 
     public function getCacheDir()
     {
+        // In the dev environment use a folder outside the shared filesystem. This greatly improves cache clear
+        // and warmup time.
+        if ($this->getEnvironment() === 'dev') {
+            return '/tmp/eb_dev_cache';
+        }
+
         return $this->rootDir . '/cache/' . $this->environment;
     }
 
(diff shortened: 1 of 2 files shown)


app/AppKernel.php
```diff
@@ -62,6 +62,12 @@ public function registerContainerConfiguration(LoaderInterface $loader)
 
     public function getCacheDir()
     {
+        // In the dev environment use a folder outside the shared filesystem. This greatly improves cache clear
+        // and warmup time.
+        if ($this->getEnvironment() === 'dev') {
```
pablothedude: This is maybe something to include in the README.md or maybe even better:
`https://github.com/OpenConext/OpenConext-engineblock/wiki/Development-Guidelines`

pablothedude:
Nice addition!

MKodde:
Thanks for the review

> And what about test?

I was reluctant in moving the cache folder for the test env, this mainly because we can not write to that folder in Travis builds. I'm going to run a couple of experiments nonetheless.

app/AppKernel.php
MKodde: Readme did not provide an ample section to add this to, but adding it to the WIKI is a good idea. Adding it!

MKodde:
The test cache seems to work fine on Travis :fireworks:

app/AppKernel.php
pablothedude: :100:

PR #782, opened 2019-10-16
Cleanup unused column

pablothedude:
The response_processing_service_binding property is no longer used and
could be removed after the current release in order to support rolling
updates.

**Please validate before merging this back if we don't break old version because of Doctrine annotations!**

https://github.com/OpenConext/OpenConext-engineblock/pull/779
https://www.pivotaltracker.com/story/show/169178875

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -11,6 +11,8 @@ If you're using OpenConext-Deploy, Ansible is configured to do this for you and 
 
 If you don't use OpenConext-Deploy you should manually add the `ini_parameters.yml` variables to the`parameters.yml` and you're good to go.
 
+In version 6.2 the unused `response_processing_service_binding` column is removed from the `sso_provider_roles_eb5` table.
+
 # Remove Symfony bootstrap cache
 Among the cleanup tasks some unused scripts have been removed from the project.
  1. The bootstrap.cache.php file, previously used by Symfony was removed
--- a/database/DoctrineMigrations/Version20191011132428.php
+++ b/database/DoctrineMigrations/Version20191011132428.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types=1);
+
+namespace OpenConext\EngineBlock\Doctrine\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20191011132428 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 DROP response_processing_service_binding');
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 ADD response_processing_service_binding VARCHAR(255) DEFAULT NULL COLLATE utf8_unicode_ci');
+    }
+}


MKodde:
Personally, my schema did not have the column, and a test2 EB 6.1.3 also does not have the column. I start wondering if this migration makes sense. @pablothedude can you shed your light on this?

pablothedude:
The only light I can shine on this is: `https://github.com/OpenConext/OpenConext-engineblock/pull/779#issuecomment-542669284`.

That we have the column removed seems plausible, because we've tested the whole feature were the migration was still in. That test2 is missing this column is therefore rather strange because the removal of this column didn't seem to land in a release or as far as I can see on master.

MKodde:
@quartje or @thijskh can you explain why this field is already gone on test2. We assume it's still present on production.

thijskh:
```
MariaDB [eb]> describe sso_provider_roles_eb5;
+-------------------------------------+--------------+------+-----+---------+----------------+
| Field                               | Type         | Null | Key | Default | Extra          |
+-------------------------------------+--------------+------+-----+---------+----------------+
| id                                  | int(11)      | NO   | PRI | NULL    | auto_increment |
| entity_id                           | varchar(255) | NO   | MUL | NULL    |                |
| name_nl                             | varchar(255) | NO   |     | NULL    |                |
| name_en                             | varchar(255) | NO   |     | NULL    |                |
| description_nl                      | varchar(255) | NO   |     | NULL    |                |
| description_en                      | varchar(255) | NO   |     | NULL    |                |
| display_name_nl                     | varchar(255) | NO   |     | NULL    |                |
| display_name_en                     | varchar(255) | NO   |     | NULL    |                |
| logo                                | longtext     | NO   |     | NULL    |                |
| organization_nl_name                | longtext     | YES  |     | NULL    |                |
| organization_en_name                | longtext     | YES  |     | NULL    |                |
| keywords_nl                         | varchar(255) | NO   |     | NULL    |                |
| keywords_en                         | varchar(255) | NO   |     | NULL    |                |
| publish_in_edugain                  | tinyint(1)   | NO   | MUL | NULL    |                |
| certificates                        | longtext     | NO   |     | NULL    |                |
| workflow_state                      | varchar(255) | NO   |     | NULL    |                |
| contact_persons                     | longtext     | NO   |     | NULL    |                |
| name_id_format                      | varchar(255) | YES  |     | NULL    |                |
| name_id_formats                     | longtext     | NO   |     | NULL    |                |
| single_logout_service               | longtext     | YES  |     | NULL    |                |
| publish_in_edu_gain_date            | date         | YES  |     | NULL    |                |
| requests_must_be_signed             | tinyint(1)   | NO   |     | NULL    |                |
| response_processing_service_binding | varchar(255) | YES  |     | NULL    |                |
| manipulation                        | longtext     | NO   |     | NULL    |                |
| type                                | varchar(255) | NO   | MUL | NULL    |                |
| attribute_release_policy            | longtext     | YES  |     | NULL    |                |
| assertion_consumer_services         | longtext     | YES  |     | NULL    |                |
| allowed_idp_entity_ids              | longtext     | YES  |     | NULL    |                |
| allow_all                           | tinyint(1)   | YES  |     | NULL    |                |
| requested_attributes                | longtext     | YES  |     | NULL    |                |
| enabled_in_wayf                     | tinyint(1)   | YES  |     | NULL    |                |
| single_sign_on_services             | longtext     | YES  |     | NULL    |                |
| shib_md_scopes                      | longtext     | YES  |     | NULL    |                |
| support_url_en                      | varchar(255) | YES  |     | NULL    |                |
| support_url_nl                      | varchar(255) | YES  |     | NULL    |                |
| consent_settings                    | longtext     | YES  |     | NULL    |                |
| coins                               | longtext     | NO   |     | NULL    |                |
+-------------------------------------+--------------+------+-----+---------+----------------+
37 rows in set (0.00 sec)
```

thijskh:
On test2, can it be that we deployed an intermediate version of EB that at some point dropped this column already, but that migration was later removed again from EB (perhaps for rolling reasons)?

MKodde:
Allright! good to see this verified! Thanks @thijskh

pablothedude:
@thijskh are you able to see which migrations were executed on prod/acc? That way I would be able to determine what the impact of this migration would be so I can update the `upgrade.md` accordingly without the need to mention how to revert hypothetical situations on development environments.

thijskh:
Both acc and prod:

```
MariaDB [eb]> select * from migration_versions;
+----------------+
| version        |
+----------------+
| 20160307162928 |
| 20160412141621 |
| 20160721121856 |
| 20161123131704 |
| 20161209145942 |
| 20161209152354 |
| 20170331145533 |
| 20170912155800 |
| 20180118115853 |
| 20180202131013 |
| 20180215132859 |
| 20180219171531 |
| 20180220080549 |
| 20180221133212 |
| 20180618134736 |
| 20180804090135 |
| 20190425205743 |
| 20190504110100 |
| 20190703235333 |
| 20190710141146 |
+----------------+
20 rows in set (0.04 sec)
```

pablothedude:
@MKodde the column still has to be removed on acc and prod as far as I can see. So it seems the situation on test2 and your local dev box has to do with some local development efforts. This shouldn't effect a thing because it was seemed before.
https://www.pivotaltracker.com/story/show/168980646/comments/207623194

In the upgrading.md I've added a comment that the column is going to be removed. I'm not going to add a note on how to skip this migration because I don't see any value in doing so. I think that if you're deviating from the regular  upgrade path you're on your own and adding documentation on how to resolve such things would only make things less clear.

MKodde:
Thanks for adding the entry in the UPGRADING.md, you could be more explicit about having to run migrations. But that is implied when reading the text, most users will find out they can run a doctrine migration. Feel free to leave as is.

MKodde:
Also, we should do some JS dependency sec-upgrading as it seems.

PR #783, opened 2019-10-16
Display the actual EB IdPs metadata in the generated XML

MKodde:
This, again very large, PR will ensure the IdPs metadata endpoint renders the correct data.

**Worklog (WIP)**
* [x] Implement the SP entity id filtering
* [x] Ensure error reporting is implemented as requested in: [168258776](https://www.pivotaltracker.com/story/show/168258776)
* [x] Add shibmd:Scope elements if applicable
* [x] Make the logo an optional field of the entities
* [x] Increase test coverage and verify correct metadata generation

**Fixes**
https://www.pivotaltracker.com/story/show/168258709

**Created new issues:**
https://www.pivotaltracker.com/story/show/169199290
https://www.pivotaltracker.com/story/show/169199410

### DIFF
--- a/library/EngineBlock/Corto/Adapter.php
+++ b/library/EngineBlock/Corto/Adapter.php
@@ -354,8 +354,6 @@ class EngineBlock_Corto_Adapter
         $application = EngineBlock_ApplicationSingleton::getInstance();
         $settings = $application->getDiContainer();
 
-        $proxyServer->setHostName($settings->getHostname());
-
         $proxyServer->setConfigs(array(
             'debug' => $settings->isDebug(),
             'ConsentStoreValues' => $settings->isConsentStoreValuesActive(),
@@ -503,7 +501,7 @@ class EngineBlock_Corto_Adapter
             $keyPair,
             EngineBlock_ApplicationSingleton::getInstance()->getDiContainer()->getAttributeMetadata(),
             new Service(
-                $proxyServer->getUrl('provideConsentService'),
+                '/authentication/idp/provide-consent',
                 'INTERNAL'
             )
         );
--- a/src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
+++ b/src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
@@ -1,0 +1,157 @@
+<?php
+
+/**
+ * Copyright 2010 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlock\Metadata\MetadataRepository;
+
+use OpenConext\EngineBlock\Metadata\Entity\IdentityProvider;
+use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
+use OpenConext\EngineBlock\Metadata\Factory\Collection\IdentityProviderEntityCollection;
+use OpenConext\EngineBlock\Metadata\Factory\Decorator\EngineBlockIdentityProviderMetadata;
+use OpenConext\EngineBlock\Metadata\Factory\Decorator\IdentityProviderProxy;
+use OpenConext\EngineBlock\Metadata\Factory\Factory\IdentityProviderFactory;
+use OpenConext\EngineBlock\Service\Metadata\ServiceReplacer;
+use OpenConext\EngineBlockBundle\Url\UrlProvider;
+
+class IdpsMetadataRepository
+{
+    /**
+     * @var CachedDoctrineMetadataRepository
+     */
+    private $repository = array();
+
+    /**
+     * @var IdentityProviderFactory
+     */
+    private $idpFactory;
+
+    /**
+     * @var UrlProvider
+     */
+    private $urlProvider;
+
+    public function __construct(
+        CachedDoctrineMetadataRepository $repository,
+        IdentityProviderFactory $idpFactory,
+        UrlProvider $urlProvider
+    ) {
+        $this->repository = $repository;
+        $this->idpFactory = $idpFactory;
+        $this->urlProvider = $urlProvider;
+
+        // Can use SP Proxy decoration
+//        $idpEntityId = $proxyServer->getUrl('idpMetadataService');
+//        $spEntityId = $proxyServer->getUrl('spMetadataService');
+//        $keyPair = $proxyServer->getSigningCertificates();
+//
+//        $visitor = new EngineBlockMetadataVisitor(
+//            $idpEntityId,
+//            $spEntityId,
+//            $keyPair,
+//            EngineBlock_ApplicationSingleton::getInstance()->getDiContainer()->getAttributeMetadata(),
+//            new Service(
+//                $proxyServer->getUrl('provideConsentService'),
+//                'INTERNAL'
+//            )
+//        );
+//
+//        $this->getMetadataRepository()->appendVisitor($visitor);
+    }
+
+    /**
+     * @param string $entityId
+     * @return ServiceProvider
+     * @throws EntityNotFoundException
+     */
+    public function fetchServiceProviderByEntityId($entityId)
+    {
+        return $this->repository->fetchServiceProviderByEntityId($entityId);
+    }
+
+    /**
+     * @return IdentityProviderProxy[]
+     */
+    public function findIdentityProviders(EngineBlockIdentityProviderMetadata $engineBlockIdentityProvider, string $keyId)
+    {
+        return $this->convertIdentityProviders(
+            $engineBlockIdentityProvider,
+            $this->repository->findIdentityProviders(),
+            $keyId
+        );
+    }
+
+    /**
+     * @param array $identityProviderEntityIds
+     * @return IdentityProvider[]
+     */
+    public function findIdentityProvidersByEntityId(
+        EngineBlockIdentityProviderMetadata $engineBlockIdentityProvider,
+        array $identityProviderEntityIds,
+        string $keyId
+    ) {
+        return $this->convertIdentityProviders(
+            $engineBlockIdentityProvider,
+            $this->repository->findIdentityProvidersByEntityId($identityProviderEntityIds),
+            $keyId
+        );
+    }
+
+    /**
+     * @param IdentityProvider[] $idps
+     */
+    private function convertIdentityProviders(
+        EngineBlockIdentityProviderMetadata $engineBlockIdentityProvider,
+        array $idps,
+        string $keyId
+    ): IdentityProviderEntityCollection {
+        // Replace the services of the IdP's (SSO and SLO)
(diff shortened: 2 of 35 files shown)


library/EngineBlock/Corto/Adapter.php
```diff
@@ -503,7 +501,7 @@ protected function enrichEngineBlockMetadata(EngineBlock_Corto_ProxyServer $prox
             $keyPair,
             EngineBlock_ApplicationSingleton::getInstance()->getDiContainer()->getAttributeMetadata(),
             new Service(
-                $proxyServer->getUrl('provideConsentService'),
+                '/authentication/idp/provide-consent',
```
pablothedude: Won't this become a problem from absolute to relative?
Discussed IRL and opened a PR: https://www.pivotaltracker.com/story/show/169204880

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
```diff
@@ -0,0 +1,144 @@
...
+
+        $collection = new IdentityProviderEntityCollection();
+        foreach ($idps as $idp) {
+            // Don't add ourselves
+            if ($idp->entityId === $engineBlockIdentityProvider->getEntityId()) {
+                continue;
+            }
+
+            // Do not reveal hidden IdP's
+            if ($idp->getCoins()->hidden()) {
```
pablothedude: Would this still be necessary? Then we could clean this up later on?

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
```diff
@@ -0,0 +1,144 @@
...
+            // Use EngineBlock certificates
+            $idp->certificates = $engineBlockIdentityProvider->certificates;
+
+            // Ignore the NameIDFormats the IdP supports, any requests made on this endpoint will use EngineBlock
+            // NameIDs, so advertise that.
+            unset($idp->nameIdFormat);
+            $idp->supportedNameIdFormats = $engineBlockIdentityProvider->supportedNameIdFormats;
+
+            // Replace service locations and bindings with those of EB
+            $transparentSsoUrl = $this->urlProvider->getUrl('authentication_idp_sso', false, null, $idp->entityId);
```
pablothedude: I think when replacing this logic we should check the ResponseProcessingService, as the solution would be presumably the same.

src/OpenConext/EngineBlock/Service/Metadata/ServiceReplacer.php
```diff
@@ -16,6 +16,14 @@
  * limitations under the License.
  */
 
+namespace OpenConext\EngineBlock\Service\Metadata;
```
pablothedude: Do we want this legacy code in our new namespace?

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
```diff
@@ -51,18 +51,11 @@ class MetadataProvider
...
-     * @param IdentityProviderFactory $idpFactory
-     * @param MetadataRepositoryInterface $metadataRepository
-     * @param StepupEndpoint $stepupEndpoint
-     */
     public function __construct(
         MetadataRenderer $factory,
         ServiceProviderFactory $spFactory,
         IdentityProviderFactory $idpFactory,
-        MetadataRepositoryInterface $metadataRepository,
+        IdpsMetadataRepository $metadataRepository,
```
pablothedude: Consider renaming this to `IdpsMetadataRepository`.

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
```diff
@@ -0,0 +1,144 @@
...
+            }
+
+            // Do not reveal hidden IdP's
+            if ($idp->getCoins()->hidden()) {
+                continue;
+            }
+
+            // Use EngineBlock certificates
+            $idp->certificates = $engineBlockIdentityProvider->certificates;
+
```
pablothedude: All this logic is already implemented in the ProxyDecorator so this could easiliy be removed later on in the issue you created, thanks for that!

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
```diff
@@ -113,18 +106,42 @@ public function metadataForIdp(string $entityId, string $ssoLocation, string $ke
...
-    public function metadataForIdpsOfSp(string $entityId, string $keyId, string $serviceProviderEntityId = null): string
-    {
-        $identityProviders = $this->metadataRepository->findIdentityProviders();
+    public function metadataForIdps(
+        string $engineIdpEntityId,
+        string $ssoLocation,
+        ?string $spEntityId,
+        string $keyId
+    ): string {
+        // 1. Load the EngineBlock IdP entity (used to override the certificates and contact persons)
```
pablothedude: Consider making two methods one with `spEntityId` and one without.
1. metadataForIdps
2. metadataForSps (or something like that)

src/OpenConext/EngineBlock/Xml/MetadataRenderer.php
```diff
@@ -19,21 +19,25 @@
...
 use Twig\Environment;
 
 class MetadataRenderer
 {
     const ID_PREFIX = 'EB';
 
+    /**
+     * The number of seconds a Metadata document is deemed valid
+     */
+    const METADATA_EXPIRATION_TIME = 86400;
```
pablothedude: Good call to put this here!

src/OpenConext/EngineBlockBundle/Controller/MetadataController.php
```diff
@@ -98,22 +99,35 @@ public function spMetadataAction(string $keyId)
...
-        $metadataXml = $this->metadataService->metadataForIdpsOfSp(
-            $this->getAbsoluteUrlForRoute('metadata_sp'),
-            $keyId,
-            null
-        );
+        // 1. Determine if the IdP metadata is requested for a specific SP based on the 'sp-entity-id' parameter.
+        $spEntityId = $request->query->get('sp-entity-id', null);
+
+        // 2..5
+        $engineEntityId = $this->getAbsoluteUrlForRoute('metadata_idp');
```
pablothedude: Consider splitting this up to call a different method on the MetadataService based on if an sp-entity-id is set or not.

src/OpenConext/EngineBlockBundle/Url/UrlProvider.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * @param string $name The route name identifier as can be found in the routing configuration
+     * @param bool $processingMode When disabled, the keyId and remoteEntityId parameters will be added to the URL
+     * @param string|null $keyId
+     * @param string|null $remoteEntityId
+     * @return string
+     */
+    public function getUrl(string $name, bool $processingMode, ?string $keyId, ?string $remoteEntityId): string
```
pablothedude: This is much more cleaner also in terms of defence in depth

library/EngineBlock/Corto/Adapter.php
MKodde: Thanks, in my tests all seemed to work as intended. But let's verify this in greater detail in the opened story!

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
MKodde: Why would this rule be obsolete?

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
MKodde: Yes, replacing it completely is not required in my opinion. But making it more robust is. I've already created a ticket for this.

https://www.pivotaltracker.com/story/show/169199290

src/OpenConext/EngineBlock/Service/Metadata/ServiceReplacer.php
MKodde: I think we do! And as mentioned I opened a tech-debt ticket in:
https://www.pivotaltracker.com/story/show/169199290

The alternative was to use this logic from the library namespace, and that is even worse in my opinion.

pablothedude:
Great work!

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
MKodde: The Class is called that exactly! Or do you mean something different?

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
MKodde: Not all of that logic is implemented there. 

I was able to remove the setting of the nameIdFormats, the other overrides remain essential for correct metadata creation.

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
MKodde: I see little benefit in this, the methods would both be cleaner, but would be almost identical. Leaving this as-is for now.

src/OpenConext/EngineBlockBundle/Controller/MetadataController.php
MKodde: See https://github.com/OpenConext/OpenConext-engineblock/pull/783#discussion_r336017761

library/EngineBlock/Corto/Adapter.php
pablothedude: :+1:

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
pablothedude: If they are only used to hide the internal magic entities they could possibly be cleaned up.
https://www.pivotaltracker.com/story/show/168249058

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
pablothedude: :+1:

src/OpenConext/EngineBlock/Service/Metadata/ServiceReplacer.php
pablothedude: :+1:

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
pablothedude: `$metadataRepository` > `$dpsMetadataRepository`

src/OpenConext/EngineBlock/Metadata/MetadataRepository/IdpsMetadataRepository.php
pablothedude: This should already be on the proxy decorator

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
pablothedude: I think that when an spEntityId is set something entirely different is done. So code wise it's almost the same but flow wise it's something different. Also when 'spEntityId' is an empty string you presumably are getting unexpected behaviour. And if it was al clear enough why was the documentation added?

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
pablothedude: :+1:

PR #784, opened 2019-10-17
Create stepup metadata entity with factory

pablothedude:
The stepup metadata endpoint needed to be created. This change will make
sure that the entity is created.

Some minor adjustments where needed to allow optional fields.
Also the EB information is moved to a dedicated decorator as this is
won't be needed for all metadata endpoints.

https://www.pivotaltracker.com/story/show/169035362

I'll address adding decorator tests and the final implementation in a later PR
This is only a MVP to let the stepup metadata endpoint work.

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
@@ -121,7 +121,7 @@ class IdentityProviderEntity implements IdentityProviderEntityInterface
     /**
      * @return Organization
      */
-    public function getOrganizationNl(): Organization
+    public function getOrganizationNl(): ?Organization
     {
         return $this->entity->organizationNl;
     }
@@ -129,7 +129,7 @@ class IdentityProviderEntity implements IdentityProviderEntityInterface
     /**
      * @return Organization
      */
-    public function getOrganizationEn(): Organization
+    public function getOrganizationEn(): ?Organization
     {
         return $this->entity->organizationEn;
     }
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Decorator/AbstractIdentityProvider.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Decorator/AbstractIdentityProvider.php
@@ -113,7 +113,7 @@ class AbstractIdentityProvider implements IdentityProviderEntityInterface
     }
 
     /**
-     * @return Logo
+     * @return Logo|null
      */
     public function getLogo(): ?Logo
     {
@@ -121,17 +121,17 @@ class AbstractIdentityProvider implements IdentityProviderEntityInterface
     }
 
     /**
-     * @return Organization
+     * @return Organization|null
      */
-    public function getOrganizationNl(): Organization
+    public function getOrganizationNl(): ?Organization
     {
         return $this->entity->getOrganizationNl();
     }
 
     /**
-     * @return Organization
+     * @return Organization|null
      */
-    public function getOrganizationEn(): Organization
+    public function getOrganizationEn(): ?Organization
     {
         return $this->entity->getOrganizationEn();
     }
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Decorator/EngineBlockServiceProviderInformation.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Decorator/EngineBlockServiceProviderInformation.php
@@ -1,0 +1,82 @@
+<?php declare(strict_types=1);
+
+/**
+ * Copyright 2010 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlock\Metadata\Factory\Decorator;
+
+use OpenConext\EngineBlock\Metadata\ContactPerson;
+use OpenConext\EngineBlock\Metadata\Factory\ServiceProviderEntityInterface;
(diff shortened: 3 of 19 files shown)


src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
```diff
@@ -111,25 +111,25 @@ public function getDisplayNameEn(): string
     }
 
     /**
-     * @return Logo
+     * @return Logo|null
      */
-    public function getLogo(): Logo
+    public function getLogo(): ?Logo
```
MKodde: This change is also in my PR, let's be careful when merging these PR's

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/AbstractIdentityProvider.php
```diff
@@ -113,25 +113,25 @@ public function getDisplayNameEn(): string
     }
 
     /**
-     * @return Logo
+     * @return Logo||null
```
MKodde: Please use a single pipe

src/OpenConext/EngineBlock/Metadata/Factory/Factory/IdentityProviderFactory.php
```diff
@@ -51,55 +50,32 @@ public function __construct(KeyPairFactory $keyPairFactory, EngineBlockConfigura
...
-        string $keyId,
-        string $ssoBindingMethod = Constants::BINDING_HTTP_REDIRECT
-    ): IdentityProviderEntityInterface {
-        $entity = new IdentityProvider($entityId);
-        $entity->singleSignOnServices[] = new Service($ssoLocation, $ssoBindingMethod);
-        $entity->certificates[] = $this->keyPairFactory->buildFromIdentifier($keyId)->getCertificate();
+        $entity = $this->buildIdentityProviderOrmEntity($entityId, $ssoLocation, $keyId);
 
-        return $this->createEntityFromEntity($entity);
+        return new EngineBlockIdentityProviderMetadata(
```
MKodde: I'm not sure this is an improvement perse. Yes, we are moving implementation details to decorator instances. But this also makes it quite hard to see what the point of every decorator is. And creating a nested structure like this might not be favorable over a more composition based solution. I'm concerned this will become a great mindf*ck in no time.

What I like, is that the factories API is much cleaner now.

tests/unit/OpenConext/EngineBlock/Metadata/Factory/Factory/ServiceProviderFactoryTest.php
```diff
@@ -48,33 +48,25 @@ public function setup()
...
+        $entity = $this->factory->createEngineBlockEntityFrom(
+            'entityID',
+            'acsLocation',
+            'default'
+        );
 
         $this->assertInstanceOf(ServiceProviderEntityInterface::class, $entity);
     }
 
     public function test_create_proxy_from_entity()
```
MKodde: You need to update the title of this test

src/OpenConext/EngineBlock/Metadata/Factory/Factory/IdentityProviderFactory.php
```diff
@@ -51,55 +50,32 @@ public function __construct(KeyPairFactory $keyPairFactory, EngineBlockConfigura
...
+            new IdentityProviderProxy(  // Add EB proxy data
```
MKodde: Is this additional wrapping required? `IdentityProviderProxy`

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/EngineBlockServiceProviderInformation.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        return $this->engineBlockConfiguration->getDescription();
+    }
+
+    public function getDescriptionEn(): string
+    {
+        return $this->engineBlockConfiguration->getDescription();
+    }
+
+    public function getLogo(): Logo
```
MKodde: Must this overloaded method not conform to the interface definition?

src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
pablothedude: I saw that but the change is the same so this shouldn't be problematic at all.

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/AbstractIdentityProvider.php
pablothedude: :+1:

src/OpenConext/EngineBlock/Metadata/Factory/Factory/IdentityProviderFactory.php
pablothedude: A decorator is composition and at least it's now happening in one place and only during construction time as the objects are immutable. So you always know that it happened there. Unless someone introduces mutability again this will greatly insights as you can inspect the object and see where something did come from.

src/OpenConext/EngineBlock/Metadata/Factory/Factory/IdentityProviderFactory.php
pablothedude: Yes, as you have seen in the original code the visitor which mutated the original entity did the same.

src/OpenConext/EngineBlock/Metadata/Factory/Factory/IdentityProviderFactory.php
MKodde: Yes, a single decoration is a perfect example of composition. The thought behind the pattern, as I see it, is that you create a single decorator object for a single purpose. 

Jamming four or five decorators into one another starts smelling of inheritance to me.

My concern as noted before is that one would need to dig into the other decorators to discover where certain behavior comes from.

pablothedude: Composition or inheritance is a type of relation not how many times a relation occurs. I agree that it could become unclear if there are a lot of different decorators, but then we could merge some together or have multiple layers. You know at least now on forehand where it's happening because they are immutable.

Maybe we could test better what is created with the factory so this could be addressed and we could easily refactor this later on.

Feel free to discuss further IRL, because this is a very interesting discussion.

pablothedude:
Your concerns will be addressed in:

PR #787, opened 2019-10-18
Add metadata stepup signing

pablothedude:
The metadta stepup signing needed to be implmented. To make sure
the metadata xml stays the same some adjustments were needed
to support the stepup metadata because not all fields should be shown.

The fields shouldn't be shown because in case of stepup only the ACS
location and the enityId is relevant. So requested attributes, UiInfo and
organization information isn't relevant and should therefore be omitted.


pablothedude:
Merged back with: https://github.com/OpenConext/OpenConext-engineblock/pull/788

PR #795, opened 2019-10-24
Add test coverage for used keys

pablothedude:
The used keys are validated to be sure the EB keys are used
for signing and that the correct key is propagated as the one.
that should be used.


pablothedude:
I've updated the `.editorconfig` accordingly to prevent it from happening again.

PR #796, opened 2019-10-25
Adjust decorator naming to reflect purpose

pablothedude:
Some decorators and helpers naming didn't reflect the purpose of
the class properly.

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Factory/ServiceProviderFactory.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Factory/ServiceProviderFactory.php
@@ -21,9 +21,8 @@ namespace OpenConext\EngineBlock\Metadata\Factory\Factory;
 use EngineBlock_Attributes_Metadata as AttributesMetadata;
 use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
 use OpenConext\EngineBlock\Metadata\Factory\Adapter\ServiceProviderEntity;
+use OpenConext\EngineBlock\Metadata\Factory\Decorator\EngineBlockServiceProvider;
 use OpenConext\EngineBlock\Metadata\Factory\Decorator\EngineBlockServiceProviderInformation;
-use OpenConext\EngineBlock\Metadata\Factory\Decorator\EngineBlockServiceProviderMetadata;
-use OpenConext\EngineBlock\Metadata\Factory\Decorator\ServiceProviderProxy;
 use OpenConext\EngineBlock\Metadata\Factory\Decorator\ServiceProviderStepup;
 use OpenConext\EngineBlock\Metadata\Factory\ServiceProviderEntityInterface;
 use OpenConext\EngineBlock\Metadata\Factory\ValueObject\EngineBlockConfiguration;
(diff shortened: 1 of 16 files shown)


src/OpenConext/EngineBlock/Metadata/Factory/Factory/ServiceProviderFactory.php
```diff
@@ -75,7 +72,7 @@ public function createEngineBlockEntityFrom(string $keyId): ServiceProviderEntit
 
         $entity = $this->buildServiceProviderOrmEntity($entityId);
 
-        return new ServiceProviderProxy( // Add EB proxy data
+        return new EngineBlockServiceProvider( // Add EB proxy data
             new EngineBlockServiceProviderInformation(  // Add EB specific information
                 new ServiceProviderEntity($entity),
```
MKodde: You could update the comment, remove the `proxy data` part.

Suggestion: `// Set EngineBlock specific properties`

Or something along those lines.

And regarding line 79: What additional business value does the EngineBlockServiceProviderInformation have over the EngineBlockServiceProvider? 

I know this decorator provides the ini configured EB settings like name/description/logo and so forth. And from a single responsibility standpoint I think this should be a decorator in its own right. The name could be improved perhaps. For example include the fact that this decorator uses configured values to decorate the entity.

Suggestion: `EngineBlockConfiguredPropertiesServiceProvider`

Or even: `ServiceProviderEngineBlockConfiguredProperties`, but this implies that other decorators should also follow this format. Starting with their role descriptor (SP/IdP) followed by a suitable description of ther intention/purpose.

The same applies for the IdP counterpart.

src/OpenConext/EngineBlock/Metadata/Factory/Helper/IdentityProviderMetadataHelper.php
```diff
@@ -19,10 +19,9 @@
 namespace OpenConext\EngineBlock\Metadata\Factory\Helper;
 
 use OpenConext\EngineBlock\Metadata\Factory\Decorator\AbstractIdentityProvider;
-use OpenConext\EngineBlock\Metadata\Factory\IdentityProviderEntityInterface;
 use OpenConext\EngineBlock\Metadata\Service;
 
-class EngineBlockIdentityProviderMetadata extends AbstractIdentityProvider
+class IdentityProviderMetadataHelper extends AbstractIdentityProvider
```
MKodde: Good call renaming the helper classes :+1:

src/OpenConext/EngineBlock/Metadata/Factory/Factory/ServiceProviderFactory.php
pablothedude: The difference between the decorators is that  the `EngineBlockServiceProvider` only implements functionality to set EB specific data. This to override the default entity data and always use the EB (proxy) data. The `EngineBlockServiceProviderInformation` is only used to set information (for example name/displayname/organization etc.) which is non functional and thus only informational. Added a docblock to the decorator to make that more clear.

It's fine to have nested decorators because this is the composition over inheritance principle. So it doesn't need a own decorator because the decorator has it's own responsibility.

I'm not a big fan of the `Properties` suggestion because only informational data should be added and properties implies more than informational data could be set which shouldn't be done at all.

MKodde: Thanks, naming improved slightly and things became more easy to understand. 

Still not optimal in my opinion, but better than it was!

pablothedude:
The tests are failing because of a security warning in Symfony. I will create a separate PR to address this.

PR #797, opened 2019-10-25
Use correct language value in organization metadata

pablothedude:
The translations weren't working for the organization block and
only english translations were used. This is fixed.

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Helper/EngineBlockServiceProviderMetadata.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Helper/EngineBlockServiceProviderMetadata.php
@@ -19,14 +19,11 @@
 namespace OpenConext\EngineBlock\Metadata\Factory\Helper;
 
 use OpenConext\EngineBlock\Metadata\Factory\Decorator\AbstractServiceProvider;
-use OpenConext\EngineBlock\Metadata\Factory\ServiceProviderEntityInterface;
 use OpenConext\EngineBlock\Metadata\IndexedService;
-use OpenConext\EngineBlock\Metadata\Logo;
-use OpenConext\EngineBlock\Metadata\RequestedAttribute;
 
 class EngineBlockServiceProviderMetadata extends AbstractServiceProvider
 {
-    public function getOrganization() : string
+    public function getOrganizationNameEn() : string
     {
         if (!$this->entity->getOrganizationEn()) {
             return '';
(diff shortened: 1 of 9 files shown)


pablothedude:
As you can see this class could be used by all possible entities and is only a helper to present metadata. So the situation you're describing is only applicable for EB and for example not for the Stepup metadata endpoint.

The logic you're describing was earlier on moved to `\OpenConext\EngineBlock\Metadata\Factory\Decorator\EngineBlockServiceProviderInformation` were the configuration is injected and all configuration fields are mapped accordingly.

This way the responsibility of presenting metadata is the responsibility of the metadata helper and the responsibility how name and description fields are mapped are the responsibility of the `EngineBlockServiceProviderInformation`-decorator. So this was needed for separation of concerns.

In order to fix the decorator it was needed and was therefore related. Because it wasn't covered in tests you didn't see it was broken earlier on. Added a task to the PR.

src/OpenConext/EngineBlock/Metadata/Factory/Helper/EngineBlockServiceProviderMetadata.php
```diff
@@ -19,29 +19,42 @@
...
         if (!$this->entity->getOrganizationEn()) {
             return '';
         }
         return $this->entity->getOrganizationEn()->name;
     }
 
-    public function getOrganizationSupportUrl() : string
+    public function getOrganizationNameNl() : string
+    {
+        if (!$this->entity->getOrganizationNl()) {
```
MKodde: I do not recall that the organization data is optional. I thought that the EB org data is always provided. Judging by the code coverage reports these blocks are not covered. If they serve an important purpose then please cover them in test. If they are redundant then please remove them.

Maybe consult with the product owner on the expected behaviour. I think we'd rather display the EN values before if available before showing an empty string or omitting the value.

src/OpenConext/EngineBlock/Xml/MetadataRenderer.php
```diff
@@ -67,19 +67,24 @@ class MetadataRenderer
...
+     * @var TimeProvider
+     */
+    private $timeProvider;
 
     public function __construct(
         Environment $twig,
         EngineBlock_Saml2_IdGenerator $samlIdGenerator,
         KeyPairFactory $keyPairFactory,
         DocumentSigner $documentSigner,
+        TimeProvider $timeProvider,
```
MKodde: Good call, using DI here!

MKodde:
Not sure this was introduced in this PR. But looking at the IdPs metadata, the UIInfo of the individual IdPs is no longer filled with their actual information. But the EngineBlock values are displayed instead.

I'll try to find where this was introduced and come back to this PR.

src/OpenConext/EngineBlock/Metadata/Factory/Helper/EngineBlockServiceProviderMetadata.php
pablothedude: You could see for yourself that it's not required:
```https://engine.surfconext.nl/authentication/stepup/metadata```

MKodde: Roger, this is added for the Stepup metadata. Consider adding this information as a comment, as the regular EngineBlock SP metadata will never trigger that condition. And that is what one would expect judging by the name of this decorator.

pablothedude: The name will be changed in ```https://github.com/OpenConext/OpenConext-engineblock/pull/796``` to be more descriptive. I'll have a look if I could add a additional docblock in this PR.

pablothedude:
Note that this PR breaks on a JS securty warning that has already been resolved downstream. #805

PR #798, opened 2019-10-25
Remove ResponseProcessingService

pablothedude:
Remove ResponseProcessingService logic from ORM entities and move this
to the ProxyServer class so legacy code only exists in the legacy
namespace as this is only used there to temporarily inject the
ConsentService used for internal binding.

https://www.pivotaltracker.com/story/show/169204880

### DIFF
--- a/library/EngineBlock/Corto/ProxyServer.php
+++ b/library/EngineBlock/Corto/ProxyServer.php
@@ -472,8 +472,8 @@ class EngineBlock_Corto_ProxyServer
         // Change the destiny of the received response
         $inResponseTo = $receivedResponse->getInResponseTo();
         $newResponse->setInResponseTo($inResponseTo);
-        $newResponse->setDestination($serviceProvider->responseProcessingService->location);
-        $newResponse->setDeliverByBinding($serviceProvider->responseProcessingService->binding);
+        $newResponse->setDestination('/authentication/idp/provide-consent');
+        $newResponse->setDeliverByBinding('INTERNAL');
         $newResponse->setReturn($this->_server->getUrl('processedAssertionConsumerService'));
 
         $idp = $this->_server->getRepository()->fetchIdentityProviderByEntityId($receivedResponse->getIssuer());
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Helper/EngineBlockServiceProviderMetadata.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Helper/EngineBlockServiceProviderMetadata.php
@@ -19,10 +19,7 @@
 namespace OpenConext\EngineBlock\Metadata\Factory\Helper;
 
 use OpenConext\EngineBlock\Metadata\Factory\Decorator\AbstractServiceProvider;
-use OpenConext\EngineBlock\Metadata\Factory\ServiceProviderEntityInterface;
 use OpenConext\EngineBlock\Metadata\IndexedService;
-use OpenConext\EngineBlock\Metadata\Logo;
-use OpenConext\EngineBlock\Metadata\RequestedAttribute;
 
 class EngineBlockServiceProviderMetadata extends AbstractServiceProvider
 {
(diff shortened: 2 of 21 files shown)


library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -472,8 +472,8 @@ function sendConsentAuthenticationRequest(
         // Change the destiny of the received response
         $inResponseTo = $receivedResponse->getInResponseTo();
         $newResponse->setInResponseTo($inResponseTo);
-        $newResponse->setDestination($serviceProvider->responseProcessingService->location);
-        $newResponse->setDeliverByBinding($serviceProvider->responseProcessingService->binding);
+        $newResponse->setDestination('/authentication/idp/provide-consent');
```
MKodde: Not entirely sure, but consider turning the destination into a constant that is defined in a logical place.

src/OpenConext/EngineBlock/Metadata/Factory/Helper/EngineBlockServiceProviderMetadata.php
```diff
@@ -19,10 +19,7 @@
 namespace OpenConext\EngineBlock\Metadata\Factory\Helper;
 
 use OpenConext\EngineBlock\Metadata\Factory\Decorator\AbstractServiceProvider;
-use OpenConext\EngineBlock\Metadata\Factory\ServiceProviderEntityInterface;
```
MKodde: :1st_place_medal:

src/OpenConext/EngineBlock/Metadata/MetadataRepository/Visitor/EngineBlockMetadataVisitor.php
```diff
@@ -152,23 +141,4 @@ private function setRequestedAttributes(ServiceProvider $entity)
...
-     *
-     * It's unclear if the 'allow all entity ids' is still required for the
-     * consent service to function.
-     *
-     * @param ServiceProvider $entity
-     */
-    private function setConsentService(ServiceProvider $entity)
-    {
-        $entity->responseProcessingService = $this->consentService;
-        $entity->allowAll = true;
```
MKodde: Is the allowAll override still implemented in the new version? You might be introducing a side effect here.. The default value is false

pablothedude: The `ServiceProviderProxy` will be used and will make sure this value is true.

MKodde: Perfect!

library/EngineBlock/Corto/ProxyServer.php
pablothedude: I've only moved the code and I didn't bother because I didn't think that would be an improvement. Do you know recall why this wasn't done in the first place.

src/OpenConext/EngineBlock/Metadata/MetadataRepository/Visitor/EngineBlockMetadataVisitor.php
pablothedude: For certainty I added it to the `\OpenConext\EngineBlock\Metadata\MetadataRepository\Visitor\EngineBlockMetadataVisitor::visitServiceProvider` method and added the original comment to the method.

library/EngineBlock/Corto/ProxyServer.php
MKodde: I did not do this previously for an unknown reason. I agree that is not a huge improvement to turn this into a constant. Let's do that when we see reuse of that value in the future.

src/OpenConext/EngineBlock/Metadata/MetadataRepository/Visitor/EngineBlockMetadataVisitor.php
MKodde: Thanks for adding it there. Note that the comment you added to that method is not gramatically sound. Consider updating that.

MKodde:
Note that the Travis build fails on the Handlebars cve. This is acceptable for now and is already fixed in `master`

PR #800, opened 2019-10-31
Reconfigure LoA settings and verify received LoA

MKodde:
The first five commits address task 1 and 3 of https://www.pivotaltracker.com/story/show/169392199

Some remarks:
1. A LoaRepository was created, tasked with loading a LoA based on its identifier.
1. The existing Mapping has been kept intact but its API changed slightly as it now returns LoA value objects instead of string identifiers.

The final commit addresses task 2

Here the actual verification is performed.

### DIFF
--- a/application/configs/application.ini
+++ b/application/configs/application.ini
@@ -164,18 +164,29 @@ pdp.client_id = "EngineBlock"
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;; SFO SETTINGS ;;;;;;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
 ;; This PCRE regex is used to blacklist incoming AuthnContextClassRef attributes on. If an empty string is used
 ;; the validation is skipped. The validator will throw an exception if the used regex is invalid.
-stepup.authn_context_class_ref_blacklist_regex = "/http:\/\/test\.openconext\.nl\/assurance\/loa[1-3]/"
-;; The loa mapping from the internal used LoA's to the Stepup Gateway LOA's
-stepup.loa.mapping[http://test.openconext.nl/assurance/loa2] = "https://gateway.tld/authentication/loa2"
-stepup.loa.mapping[http://test.openconext.nl/assurance/loa3] = "https://gateway.tld/authentication/loa3"
+stepup.authn_context_class_ref_blacklist_regex = "/https:\/\/.example\.org\/assurance\/loa[1-3]/"
+
+;; The loa mapping from the internal used LoA's to the Stepup Gateway LOA's.
+;; Specification of the mapping: stepup.loa.mapping.[int 1|2|2].[string engineblock|gateway]
+;; The integer after the mapping field indicates the LoA level (1, 2 or 3 are supported).
+;; The engineblock or gateway keys specify the LoAs identifier as will be carried in the AuthnContextClassRef of an assertion.
+stepup.loa.mapping.1.engineblock = "http://example.org/assurance/loa1"
+stepup.loa.mapping.1.gateway = "https://gateway.tld/authentication/loa1"
+stepup.loa.mapping.2.engineblock = "http://example.org/assurance/loa2"
+stepup.loa.mapping.2.gateway = "https://gateway.tld/authentication/loa2"
+stepup.loa.mapping.3.engineblock = "http://example.org/assurance/loa3"
+stepup.loa.mapping.3.gateway = "https://gateway.tld/authentication/loa3"
+
 ;; The fallback LoA to return when the Stepup authentication fails but is not required
 stepup.loa.loa1 = "https://gateway.tld/authentication/loa1"
+
 ;; The EntityId (metadata URL) used in the callout to the SFO endpoint of the configured Stepup Gateway
-stepup.gateway.sfo.entityId = "https://engine.vm.openconext.org/authentication/stepup/metadata"
+stepup.gateway.sfo.entityId = "https://engine.example.org/authentication/stepup/metadata"
 ;; The single sign-on endpoint used for Stepup Gateway SFO callouts
-stepup.gateway.sfo.ssoLocation = "https://engine.vm.openconext.org/functional-testing/gateway/second-factor-only/single-sign-on"
+stepup.gateway.sfo.ssoLocation = "https://engine.example.org/functional-testing/gateway/second-factor-only/single-sign-on"
 ;; The public key from the Stepup Gateway IdP
 stepup.gateway.sfo.keyFile = "/etc/openconext/engineblock.crt"
 
--- a/src/OpenConext/EngineBlock/Metadata/Loa.php
+++ b/src/OpenConext/EngineBlock/Metadata/Loa.php
@@ -67,14 +67,12 @@ class Loa
     }
 
     /**
-     * @param int $level
+     * @param Loa $loa
      * @return bool
      */
-    public function levelIsHigherOrEqualTo($level)
+    public function levelIsHigherOrEqualTo(Loa $loa)
     {
-        Assertion::integer($level, 'Provide the integer value representing the LoA level');
-        Assertion::greaterThan($level, 0, 'Please provide a positive integer value');
-        $isHigherOrEqualTo = $this->level >= $level;
+        $isHigherOrEqualTo = $this->level >= $loa->getLevel();
         return $isHigherOrEqualTo;
     }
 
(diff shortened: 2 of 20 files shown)


pablothedude:
Did you mis `OpenConext\EngineBlockBundle\Tests\StepupDecisionTest` ?

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
```diff
@@ -0,0 +1,69 @@
+<?php
```
pablothedude: Is the path correct considering it is a repository and thus infrastructure?

application/configs/application.ini
```diff
@@ -164,18 +164,29 @@ pdp.client_id = "EngineBlock"
...
+;; The engineblock or gateway keys specify the LoAs identifier as will be carried in the AuthnContextClassRef of an assertion.
+stepup.loa.mapping.1.engineblock = "http://example.org/assurance/loa1"
+stepup.loa.mapping.1.gateway = "https://gateway.tld/authentication/loa1"
+stepup.loa.mapping.2.engineblock = "http://example.org/assurance/loa2"
+stepup.loa.mapping.2.gateway = "https://gateway.tld/authentication/loa2"
+stepup.loa.mapping.3.engineblock = "http://example.org/assurance/loa3"
+stepup.loa.mapping.3.gateway = "https://gateway.tld/authentication/loa3"
+
 ;; The fallback LoA to return when the Stepup authentication fails but is not required
 stepup.loa.loa1 = "https://gateway.tld/authentication/loa1"
```
pablothedude: This is probably not needed anymore

pablothedude:
Wouldn't it be better to have one single LOA with the identifiers for both EB and GW. This because sometimes the EB LOA is leading but you need the GW LOA identifier?

MKodde:
> Did you mis `tests/unit/OpenConext/EngineBlockBundle/Stepup/StepupDecisionTest.php`?

Yup, note the WIP commit, indicating it is still work in progress. The decision test was yet to be fixed up.

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
MKodde: Do you mean is the namespace correct?

I do not agree that a repository is infrastructural by definition. It is tasked with parsing configuration that allready passed an infrastructural parsing 'layer'. Granted, it is a Symfony service so puristically can be considered to be infrastructural. This might be a good candidate for a real life discussion.

application/configs/application.ini
MKodde: Yes it is

MKodde:
> Wouldn't it be better to have one single LOA with the identifiers for both EB and GW. This because sometimes the EB LOA is leading but you need the GW LOA identifier?

I don't understand. Can you provide more context please?

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
pablothedude: This seems like a concrete (in memory) implementation of a LoaRepository. Consider renaming it for example to LoaConfiguration.

application/configs/application.ini
pablothedude: But it would be better to then just use `1`as value as this is the internal used value. And the identifier used seems already be configured through the loa mapping itself.

pablothedude:
> > Did you mis `tests/unit/OpenConext/EngineBlockBundle/Stepup/StepupDecisionTest.php`?
> 
> Yup, note the WIP commit, indicating it is still work in progress. The decision test was yet to be fixed up.

My answer was a reaction on your thoughts in the PR description: `Provide test coverage for the policy decision class. That was omitted during initial development for some reason.` So I thought you directed that to me and was unaware of the existence of the test.

> > Wouldn't it be better to have one single LOA with the identifiers for both EB and GW. This because sometimes the EB LOA is leading but you need the GW LOA identifier?
> 
> I don't understand. Can you provide more context please?

Consider adding two helper methods `getEBIdentifier`/`getGWIdentifier` on the LOA VO or a single method with a parameter (`getIdentifier(string $for)`) for the environment were you would want the identifier for. Then when you have a LOA you also have both identifiers which would make the decision easier because you don't have to cast a EB LOA to a GW LOA.

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
MKodde: Personally I think a repository is a suitable name for the task this class performs. Even though it does not behave similarly to the repository found in the design pattern. If I were to rename this, I would tend to use `Collection`, `Store` or something along those lines.

application/configs/application.ini
MKodde: Leaving you with the introduction of matching logic. Note that a loa, in theory, does not have a numeric identifier perse. Government use indicators like low/medium/high. Those are not easily mapped back to a number.

MKodde:
> Consider adding two helper methods ... EB LOA to a GW LOA.

A tempting proposition, this might prevent injecting the Repository everywhere. I'll look into this.

src/OpenConext/EngineBlock/Metadata/Loa.php
```diff
@@ -0,0 +1,101 @@
...
+        $loa->identifier = $identifier;
+
+        return $loa;
+    }
+
+    /**
+     * @param int $level
+     * @return bool
+     */
+    public function levelIsHigherOrEqualTo($level)
```
pablothedude: Why not inject the LOA VO itself, than typehinting could be used and the additional validation could be omitted because that is already done on initialization of the VO itself.

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
```diff
@@ -0,0 +1,70 @@
...
+            );
+            Assertion::keysExist(
+                $mapping,
+                ['engineblock', 'gateway'],
+                'Both the engineblock and gateway keys must be present in every LoA mapping.'
+            );
+            Assertion::string($mapping['engineblock'], 'The EngineBlock LoA must be a string value');
+            Assertion::string($mapping['gateway'], 'The Gateway LoA must be a string value');
+
+            $this->store[$mapping['engineblock']] = Loa::create($level, $mapping['engineblock']);
```
pablothedude: Could this construction not have serious security implications? If one fetches a loa for EB but receives a LOA for GW?

src/OpenConext/EngineBlock/Metadata/Loa.php
MKodde: Yes, I copied this from StepUp GW, and have asked myself the same question some times. The one argument not to do it, is that you can now compare a GW and an EB LoA without having to fetch the counterpart.

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
MKodde: I still don't see how this would be a security issue. A misconfiguration could mess up the way the application behaves, and could introduce unexpected behaviour. Like it could with any other config option.

However I do like to be as safe as possible, and adding additional validity checks does not harm a thing. I'm adding a check on the LoaMapper that will check if no duplicate LoA level identifiers are used.

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
pablothedude: Yes, the name is perfectly suitable for the task it performs but it would imply something different. If the naming could make things much more clear I would opt to do so. So please be aware that it isn't clear for others that this repository isn't a repository as used in our ubiquitous language.

application/configs/application.ini
pablothedude: For our domain we seem to need numeric LOA's to compare if LOA is higher then another  or is the same as this would be needed in our domain. I assumed you thought the same as they were defined in the LOA VO itself. We also have a mapping to do just that. Map the identifier easily back to a number.

The only reason the configuration is currently there is to define the loa level to return if a stepup callout failed but was allowed to fail. And for our domain the GW identifier is totally irrelevant.

I see it's getting asserted in the LoaMapping so feel free to leave it in.

src/OpenConext/EngineBlock/Metadata/LoaRepository.php
pablothedude: Thanks for the LoaMapping assertions to make sure our domain is valid!

pablothedude:
All seems fine, would you only get back on:

> I don't like how the identifier is used in the LOA VO because it isn't truly an identifier. You would expect that if you are comparing two LOA's by identifier in the same domain they would be the same. Now you could compare them by identifier and they wouldn't be the same while they are.

MKodde:
Quote of the original post was abbreviated for brevity

> All seems fine, would you only get back on:
> > I don't like how the identifier is used ..... wouldn't be the same while they are.

Your tension is: the mapping allows us to map LoA levels to identifiers that do not match perse.  One could configure (in the engineblock.ini) a https://vm.openconext.org/assurance/loa1 to be level 2.

And I grant you that, that would indeed be potentially harmful, and be incorrect.

I don't see how we could improve on this in a workable manner, and we must be able to trust the users of the software configure it correctly. Otherwise, they can expect the software to malfunction.

PR #801, opened 2019-11-05
Cleanup dead code after metadata refactoring

pablothedude:
Some unused code could be removed becasue it's not needed anymore
after the metadata endpoint refactoring.


pablothedude:
This was a quick test to determine if the old metadata logic could be removed. This isn't possible because this is used for legacy AM code.
```https://www.pivotaltracker.com/story/show/168980881```

TLDR This won't be fixed now because the legacy code support will be dropped may 2020.

PR #802, opened 2019-11-06
Implement countermeasures against CVE 2019-3465

pablothedude:
This change will apply the countermeasures to harden against
CVE 2019-3465 and will effectively bump `robrichards/xmlseclibs` to
version 3.0.4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ The following changes where introduced in this release:
 * Update testing tools (PHPUnit, Mockery & Phake) #777
 
 
+## 5.13.3
+This is a security release that will harden the application against CVE 2019-3465
+* Implement countermeasures against CVE 2019-3465 #802
+
 ## 5.13.2
 A missing feature was implemented
 
--- a/composer.lock
+++ b/composer.lock
@@ -1,7 +1,7 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,11 @@ The following changes where introduced in this release:
 * Bump handlebars to version 4.3.1 #761
 * Update Composer settings for improved PHP 7.2 support #763
 * Update testing tools (PHPUnit, Mockery & Phake) #777
+* Implement countermeasures against CVE 2019-3465 #802
+
+
+## 5.13.3
+This is a security release that will harden the application against CVE 2019-3465
```
MKodde: I'm dumping all my feedback here
1. I think that PR 802 should be in 5.13.3?
2. Please rename the PR/Commit to something we agreed upon. For example: `Force upgrade of xmlseclibs to version 3.0.4`. At least something that makes explicit which package is updated to a certain version.

Other than that, looks good to go!

pablothedude: 1. I've got two separate commits because on master we have php72 and are too much diverged so I don't think we would want to merge the release with 5.6 changes back into a branch with 7.2 changes and strict typing. This way it's a lot cleaner.
2. I don't think your suggestion is better because it won't describe why the commit was needed. Could you elaborate how your suggestion would be more valuable over time? I'll change this for now to get this PR reviewed and get the job done but certainly don't agree.

MKodde: 1. Good call
2. I'm not going into detail here, and we talked about this yesterday at length. To summarize: the commit title describes the 'what' and here, we have a habit to describe what package is updated to what version. The why can explain the need for that change. And that portion of you commit was just fine. https://chris.beams.io/posts/git-commit/#seven-rules

This is not something I would block progress on this PR for, I'd just like to see our patching effor changes being alike.

PR #804, opened 2019-11-12
Stop disasembling the Edugain publication settings

MKodde:
This was first tagged to be removed in release 6.2, but we can rid ourselves of this dependency in 6.1. As EB should stop depending on these coins in 6.1

See: https://www.pivotaltracker.com/story/show/164925138


MKodde:
Build breaks on handlebars JS dev dependency. I suggest to not fix this in this PR. But will open a PR on master to address this

PR #806, opened 2019-11-12
Defense in depth SAML Response validation

MKodde:
This PR adds two important safety features to EngineBlock

## 1. A test to verify we are not vulnerable to the Hackmanit signature validation bypass exploit
This simulates the H01 finding described in detail in [1]. Effectively verifying we are not vulnerable to this exploit.

The test intercepts the SAML Response sent from the IdP and injects malicious XML to the Signature element, exploiting the vulnerability. The response is then sent to EngineBlock to process.

Depending on the presence of the safe xmlseclibs, EngineBlock would then process this assertion without a problem. Allowing the attacker to provide an assertion of their liking. From xmlseclibs 3.0.4, the SAML2 library is no longer vulnerable to this attack. And in this PR, EngineBlock will also provide defense in-depth protection against tampered with SAML Responses.

## 2. Defense in depth countermeasures 
Several checks are performed on the SAMLResponse. This allows us to reject clearly invalid responses before actually processing them. See this as an early, high level validation process.

The tests give insights into the actual validations that are performed in the SamlresponseValidator.

This addresses finding H01 in the Security audit performed by Hackmanit

See:
1. https://www.hackmanit.de/en/blog-en/82-xml-signature-validation-bypass-in-simplesamlphp-and-xmlseclibs
2. https://simplesamlphp.org/security/201911-01
3. https://nvd.nist.gov/vuln/detail/CVE-2019-3465


MKodde:
Target release pending. @thijskh @pmeulen @quartje what releases would you like to see these changes land in?

My gut feeling tells me it should be somewhere post 5.13 but before 6.1

pablothedude:
If this is the same as our forked repository, and you've just pushed to this remote just feel free to merge. Would you else elaborate what you want to have reviewed again?

MKodde:
> Would you else elaborate what you want to have reviewed again? 

I mentioned you as a reviewer, mainly to keep you in the loop on where this will land.

PR #809, opened 2019-11-13
Fix typos for stepup no token

pablothedude:
Some typos were on the no token feedback page. The typos are fixed and the missing link to the registration support page is also added.

https://www.pivotaltracker.com/n/projects/1425900/stories/169087364


pablothedude:
The build breaks on: ```https://github.com/OpenConext/OpenConext-engineblock/pull/808``` and is getting fixed there.

PR #810, opened 2019-11-13
Prevent double schema and hostname on logo

MKodde:
When the IdPs metadata is rendered, Manage configured IdP logo URLs are used. Which are URIs, in contrary to the configured EB logo which is a relative path to the locally sourced logo.

To get both situations to render the logo URI correctly, the EB config object was updated. It now uses the EB hostname to create a logo URI the schema is set to HTTPS as we can assume TLS is used for Engine.

Finally the ui_info.xml.twig was updated to no longer prepend the schema and hostname.

A IdPs Metadata.feature scenario is used to verify that the logos are rendered without double hostnames.

https://www.pivotaltracker.com/story/show/169724838

### DIFF
--- a/theme/material/templates/modules/Authentication/View/Metadata/partial/ui_info.xml.twig
+++ b/theme/material/templates/modules/Authentication/View/Metadata/partial/ui_info.xml.twig
@@ -4,6 +4,6 @@
     <mdui:Description xml:lang="nl">{{ metadata.descriptionNl }}</mdui:Description>
     <mdui:Description xml:lang="en">{{ metadata.descriptionEn }}</mdui:Description>
     {% if metadata.logo is not null %}
-    <mdui:Logo height="{{ metadata.logo.height }}" width="{{ metadata.logo.width }}">{{ app.request.getSchemeAndHttpHost() ~ metadata.logo.url }}</mdui:Logo>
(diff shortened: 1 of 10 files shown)


MKodde:
Sorry I was a bit quick with pressing the request review button :) I've been fixing the failing tests. And should come through with a passing Travis build soon..

theme/material/templates/modules/Authentication/View/Metadata/partial/ui_info.xml.twig
```diff
@@ -4,6 +4,6 @@
     <mdui:Description xml:lang="nl">{{ metadata.descriptionNl }}</mdui:Description>
     <mdui:Description xml:lang="en">{{ metadata.descriptionEn }}</mdui:Description>
     {% if metadata.logo is not null %}
-    <mdui:Logo height="{{ metadata.logo.height }}" width="{{ metadata.logo.width }}">{{ app.request.getSchemeAndHttpHost() ~ metadata.logo.url }}</mdui:Logo>
+    <mdui:Logo height="{{ metadata.logo.height }}" width="{{ metadata.logo.width }}">{{ metadata.logo.url }}</mdui:Logo>
```
pablothedude: If it would be a problem not to use https as default you could consider passing a schema parameter to the Logo VO?

MKodde: Good suggestion, I'm keeping this in the back of my mind if this turns out to be a problem.

PR #817, opened 2019-11-21
Remove migration to support 5.13

pablothedude:
This migration landed in 6.0 and resulted in a regression in the
previous release (5.13) because this release was still relying
on those columns.

This hotfix will make sure that the columns are kept in order to
support the previous release and will be targeted on a release > 6.0.

https://www.pivotaltracker.com/story/show/168565818


MKodde:
Some concerns:

1. If I understand correctly: 
   - in 5.13 we stopped using the old COIN fields
   - in 6.0 we thought we could remove the database columns, but we could not as this would make a rollback to 5.13 impossible, as that version still had the doctrine column annotations
   - in 6.1 we can safely remove the database columns, as the annotations have been removed in 6.0
1. There was no warning about this migration in the 6.0 changelog entry or in the upgrading notes. This change also eliminates the need for that. But in the future, we should be more carefull to update those files! And I think we should update the release notes for this 6.0 release, and clearly state the intended deprecation path.

pablothedude:
To be precise:
* 5.12 serialized coin support was added
* ~~5.13~~ unrelated sfo/stepup release
* 6.0  entity properties were removed and migrations were dropped the latter introduced the regression in 5.13 

If the migrations would be moved to 6.1 this should be safe as far as I could see.
The problem was that a single PR was created with all the efforts into one with all changes which needed to be done but this wasn't later on being split as we didnt know the roadmap.

I don't think a changelog entry would have prevented this from happening and would prevent such things in the future. I think the best option would be to test this in our QA-tooling or build pipeline. An other option would be to test this in OC-Deploy else we would still have a change this could happen.

MKodde:
> I don't think a changelog entry would have prevented this from happening and would prevent such things in the future.

The fact that we did not have an entry advertises this (faulty) change is what concerns me. And we should be more careful in that regard in the future. I agree with the rest of your reasoning.

PR #818, opened 2019-11-25
Add dockerized actions testing

pablothedude:
This commit will add dockerized testing with github actions.
This is a POC and need some TLC in order to let it function properly.

- [x] Database hostnames in application.ini
- [x] Selenium hostname in behat.yml
- [ ] NPM should not be ran as root in php-fpm container (use separate container or multistage dockerfile)
- [x] Add matrix to test supported PHP versions
- [ ] Discuss adding build stage to create and validate a release
- [ ] Discuss adding a test to validate the previous major release against the current DB schema


docker/php-fpm/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# Install dependencies
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+RUN apt-get update && apt-get install -y \
+    git \
+    nodejs \
+    npm \
+    zip \
+    chromium \
+    libpng-dev \
+    && docker-php-ext-install pdo_mysql exif gd \
```
MKodde: Do the ampersands serve a practical purpose?

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/Debug.php
```diff
@@ -46,10 +46,10 @@ public function getFunctions()
      * @param mixed $_
      * @return string
      */
-    public function varDump($expression, $_ = null)
+    public function varDump(...$expression)
     {
         ob_start();
-        var_dump(func_get_args());
+        var_dump($expression);
         return ob_get_clean();
```
MKodde: Was this required to get this PR to work? If not, it might be a good idea to strip it from this PR

pablothedude: The var_dump output differs per distro. Because it wouldn't be safe to rely on this function I've changed it a bit so the test would now also succeed.

docker/php-fpm/Dockerfile
pablothedude: Yes, to keep the intermediate container small. This shouldn't be build a lot often so a lot of intermediate steps (which could be cached) aren't needed.

application/configs/application.ini
```diff
@@ -117,13 +117,13 @@ logger.conf.handler.fingers_crossed.conf.activation_strategy.conf.action_level =
 ;;;;;;;;;;;; DATABASE SETTINGS ;;;;;;;;;;;;;;;;;
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-database.host = localhost
+database.host = db.vm.openconext.org
```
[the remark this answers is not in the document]
pablothedude: Personally I would like a separate CI config, but keep it in mind.

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,36 @@
...
+            - name: Init environment
+              run: cd docker && ./init.sh
+            - name: Run unit tests
+              run: |
+                  cd docker && docker-compose exec -T php-fpm.vm.openconext.org bash -c '
+                      ./app/console doctrine:schema:drop --force --env=test && \
+                      ./app/console doctrine:schema:create --env=test && \
+                      ./vendor/bin/phpunit --configuration=./tests/phpunit.xml --testsuite=eb4 && \
+                      ./vendor/bin/phpunit --configuration=./tests/phpunit.xml --testsuite=unit && \
+                      ./vendor/bin/phpunit --configuration=./tests/phpunit.xml --testsuite=integration
```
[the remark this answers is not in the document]
pablothedude: Good suggestion

docker/php-fpm/Dockerfile-php74
```diff
@@ -0,0 +1,35 @@
...
+    && pecl install apcu \
+    && pecl install apcu_bc-1.0.3 \
+    && docker-php-ext-enable apcu --ini-name 10-docker-php-ext-apcu.ini \
+    && docker-php-ext-enable apc --ini-name 20-docker-php-ext-apc.ini
+
+# Composer
+RUN curl -sL https://getcomposer.org/installer | php -- --install-dir /usr/bin --filename composer
+
+# Npm (this is a temporarily solution and should be fixed '--force --unsafe-perm')
+RUN npm cache clean --force --unsafe-perm
```
MKodde, line 25: Is the `--force --unsafe-perm` construction still required in the Docker env? If it is, maybe we can work around it by setting file permissions correctly? Something along the lines: `sudo chown -R $(whoami) ~/.npm` as this seems to be constructed to prevent write permission issues in the npm cache dir.

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/Debug.php
```diff
@@ -32,10 +30,27 @@ public function getFunctions()
...
+    /**
+     * Provides var dump functionality for use in Twig templates
+     *
+     * @SuppressWarnings(PHPMD.CamelCaseParameterName)
+     *
+     * @param mixed $expression
+     * @param mixed $_
+     * @return string
+     */
+    public function varDump(...$expression)
```
MKodde, line 47: This twig function was removed recently iirc? Why add it in this commit? It does not seem to be used in the changeset of this commit. Maybe you added it for some other reason? In that case maybe move it to a 'fixed the build' commit?

docker/php-fpm/Dockerfile-php74
pablothedude: The problem is that root is the current user in the Dockerfile. So we need to execute Npm as another user than root. Therefore we need to create a user in the dockerfile or use a multistage container with the appropriate user/rights to build the assets and then finally copy the files in the final container.

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/Debug.php
pablothedude: Thanks for the heads up. I've rebased the changes in this branch on top of develop and clearly missed the removal of the original function in develop. Will remove this.

PR #819, opened 2019-11-25
Prevent invalid assertion when stepup LoA is set

pablothedude:
When setting a Loa on a Saml2 Assertion, the LOA VO was set on the AuthnContextClassRef.  In certain distro's this resulted in a false positive when asserting the `is_string` method. Because the VO had a `__toString` method.

The method is now removed and explicitly a string is given as parameter while maintaining the VO in Corto itself.

Some additional boyscouting was done:
 * Travis branch regex fixed
 * Enabled `strict_variables` on test to prevent a silent failure and thus removed magic

https://www.pivotaltracker.com/story/show/169976060

**This needs also to be cherrypicked to 6.0**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,7 +84,9 @@ script:
 branches:
   only:
     - master
-    - /^feature\/metadata(.*)$/
+    - /^feature\/(.*)$/
+    - /^hotfix\/(.*)$/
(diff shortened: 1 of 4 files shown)


pablothedude:
Your concerns are addressed.

PR #820, opened 2019-11-27
Fix namespace after cherry picked bugfix

pablothedude:
Stepup was moved in 6.0 from the EngineBlockBundle namespace to the
EngineBlock namespace. After #819 was fixed for 6.1 this was cherry
picked into 6.0. This resulted in the namespace from the imported
Stepup classes being changed to the new namespace.


.travis.yml
```diff
@@ -87,6 +87,7 @@ branches:
     - /^feature\/(.*)$/
     - /^hotfix\/(.*)$/
     - /^bugfix\/(.*)$/
+    - /^release\/(.*)$/
```
MKodde: :100:

pablothedude:
Updated the relevant NPM and Symfony dependencies according to the advisories.

PR #823, opened 2020-01-15
Repair ChromeDriver issue and resolve NodeSass issue

MKodde:
We no longer use Chrome:stable on Travis xenial. This resolves the incompatibility between Chrome and the driver.

Finally Node SASS was updated. Even though the 'fix' mentioned above. It was still considered a production dependency by other packages..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,4 +100,3 @@ addons:
   hosts:
     - engine.vm.openconext.org
   mariadb: '10.2'
-  chrome: stable
--- a/theme/package-lock.json
+++ b/theme/package-lock.json
@@ -4843,9 +4843,9 @@
       "dev": true
     },
     "globule": {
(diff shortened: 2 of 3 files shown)


build.xml
```diff
@@ -144,7 +144,7 @@
 
     <target name="npm-audit" description="Run npm audit">
         <exec dir="${basedir}/theme" executable="npm" failonerror="true">
-            <arg line="run audit"/>
+            <arg line="run audit --package-lock-only"/>
```
pablothedude: Is the flag really needed, because according to the docs this only seem to work with the `fix` option? OPr am I missing something here?
`Run audit fix without modifying node_modules, but still updating the pkglock:`

MKodde: Yeah, I think it can be removed. Did not seem to have the desired effect.

PR #824, opened 2020-01-16
Metadata UI Info optimalisation

MKodde:
The displayname can always be set because name:en is a required field. Use this preference for the value:

* For EN: displayname:en, name:en
* For NL: displayname:nl, name:nl, name:en

Empty descriptions need not be displayed in the metadata.

Previously an empty tag was rendered, now the tag is omitted in its entirety. It was not a trivial task to verify this change in a functional test, but given the low complexity of this change it is a 'risk' I feel we can take.

https://www.pivotaltracker.com/story/show/170401452

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ProxiedIdentityProvider.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ProxiedIdentityProvider.php
@@ -98,4 +98,35 @@ class ProxiedIdentityProvider extends AbstractIdentityProvider
         $ssoLocation = $this->urlProvider->getUrl('authentication_idp_sso', false, null, $this->getEntityId());
         return [new Service($ssoLocation, Constants::BINDING_HTTP_REDIRECT)];
     }
+
+    /**
+     * Best effort to show a Dutch display name.
+     *
+     * Buisiness rule:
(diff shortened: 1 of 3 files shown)


src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ProxiedIdentityProvider.php
```diff
@@ -98,4 +98,35 @@ public function getSingleSignOnServices(): array
...
+
+    /**
+     * Best effort to show a Dutch display name.
+     *
+     * Buisiness rule:
+     * displayname:nl, name:nl, name:en
+     */
+    public function getDisplayNameNl(): string
+    {
+        if (empty(parent::getDisplayNameNl())) {
```
pablothedude: Please replace all `parent::` occurrences to `$this->entity->`

pablothedude:
Nice to have this tested separately!

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ProxiedIdentityProvider.php
MKodde: Very good catch. Fixing this!

PR #826, opened 2020-01-16
Show Organization data in IdP(s) metadata

MKodde:
Initially only SP's where treated with the Organization metadata (consisting of Name, DisplayName and an URL). Now also the IdP and IdPs metadata will display that data.

In addition, empty valued organization data will not be rendered as an empty xml element. But are omitted from the document.

https://www.pivotaltracker.com/story/show/170401380

### DIFF
--- a/theme/material/templates/modules/Authentication/View/Metadata/partial/organization.xml.twig
+++ b/theme/material/templates/modules/Authentication/View/Metadata/partial/organization.xml.twig
@@ -1,8 +1,20 @@
 <md:Organization>
+    {% if metadata.organizationNameNl is not empty %}
     <md:OrganizationName xml:lang="nl">{{ metadata.organizationNameNl }}</md:OrganizationName>
+    {% endif %}
+    {% if metadata.organizationNameEn is not empty %}
     <md:OrganizationName xml:lang="en">{{ metadata.organizationNameEn }}</md:OrganizationName>
+    {% endif %}
+    {% if metadata.organizationNameNl is not empty %}
     <md:OrganizationDisplayName xml:lang="nl">{{ metadata.organizationNameNl }}</md:OrganizationDisplayName>
+    {% endif %}
+    {% if metadata.organizationNameEn is not empty %}
     <md:OrganizationDisplayName xml:lang="en">{{ metadata.organizationNameEn }}</md:OrganizationDisplayName>
-    <md:OrganizationURL xml:lang="nl">{{ metadata.supportUrlNl }}</md:OrganizationURL>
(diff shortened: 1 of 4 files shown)


theme/material/templates/modules/Authentication/View/Metadata/partial/organization.xml.twig
```diff
@@ -1,8 +1,20 @@
...
+    {% endif %}
+    {% if metadata.organizationNameNl is not empty %}
     <md:OrganizationDisplayName xml:lang="nl">{{ metadata.organizationNameNl }}</md:OrganizationDisplayName>
+    {% endif %}
+    {% if metadata.organizationNameEn is not empty %}
     <md:OrganizationDisplayName xml:lang="en">{{ metadata.organizationNameEn }}</md:OrganizationDisplayName>
-    <md:OrganizationURL xml:lang="nl">{{ metadata.supportUrlNl }}</md:OrganizationURL>
-    <md:OrganizationURL xml:lang="en">{{ metadata.supportUrlEn }}</md:OrganizationURL>
+    {% endif %}
+    {% if metadata.organizationUrlNl is not null and metadata.organizationUrlNl is not empty %}
```
pablothedude: Could they be null?

pablothedude: Same for the logic a few lines below.

MKodde: Yes and no

Yes: the method getOrganizationUrlNl specifies a possible null result
No: The organization object itself should not be available when the Url is not presen.

I feel it best to keep things as they are now. Changing signatures at this point seems a bit risky..

MKodde:
> Are the getters below still used in the the IdentityProviderMetadataaHelper?

Yes they are. The organizatoin.xml.twig will use them, and that template is loaded for IdP's via IdPs.xml.twig. Wether or not they are actually present is up to the configuration of the proxied IdPs (configuration in Manage).

theme/material/templates/modules/Authentication/View/Metadata/partial/organization.xml.twig
pablothedude: `getOrganizationUrlNl` do return `strings`: 
https://github.com/OpenConext/OpenConext-engineblock/blob/855e3a78742cdb80719c312409e0068365811995/src/OpenConext/EngineBlock/Metadata/Factory/Helper/IdentityProviderMetadataHelper.php#L50

PR #827, opened 2020-01-30
According to the spec, EmailAddress needs to have a mailto: prefix

thijskh:


tests/unit/OpenConext/EngineBlock/Xml/MetadataRendererTest.php
```diff
@@ -169,6 +177,8 @@ public function the_metadata_factory_should_return_valid_signed_xml_for_sp()
         $entityDescriptor = new EntityDescriptor($dom->childNodes->item(1));
         $this->assertInstanceOf(EntityDescriptor::class, $entityDescriptor);
 
+        $this->assertContains('<md:EmailAddress>mailto:admin@example.org</md:EmailAddress>', $xml);
```
pablothedude: Consider a separate test for prepending the `mailto:` for an email address or at least add a comment that the `mailto` is required and therefore tested.

thijskh: Alright, added a comment for now. Certainly possibility to refactor further, but then I would test for presence of various metadata elements that are now set, seems out of scope for now.

PR #830, opened 2020-02-07
Add explicit reference to metadata xml signature.

thijskh:
This makes explicit what is signed. Otherwise validation fails
in some implemenations.

Would normally take the first element, the root, but our templates
have comment with TOS in them. Could just remove this? And then
change code to look for firstChild.

### DIFF
--- a/src/OpenConext/EngineBlock/Xml/DocumentSigner.php
+++ b/src/OpenConext/EngineBlock/Xml/DocumentSigner.php
@@ -32,13 +32,18 @@ class DocumentSigner
         $doc = new DOMDocument();
         $doc->loadXML($source);
 
+        // Find root element to sign. The firstChild is the TOS comment,
+        // so need to skip over that.
+        $rootNode = $doc->childNodes[1];
+
         // Create sign object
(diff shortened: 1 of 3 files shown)


src/OpenConext/EngineBlock/Xml/DocumentSigner.php
```diff
@@ -32,13 +32,18 @@ public function sign(string $source, X509KeyPair $signingKeyPair) : string
         $doc = new DOMDocument();
         $doc->loadXML($source);
 
+        // Find root element to sign. The firstChild is the TOS comment,
+        // so need to skip over that.
+        $rootNode = $doc->childNodes[1];
```
pablothedude: You could consider validating if the childnode is an instance of DOMElement and else throw an exception.

pablothedude: If you like I could commit a change on this branch to do so.

thijskh: Thanks, I have added this in db0ff5d

PR #834, opened 2020-02-26
Remove remaining eduGAIN code from EngineBlock

MKodde:
The AbstractRole (doctrine entity) still had some eduGAIN fields that where left in place for rolling updateability. They have now been made fully deprecated and are removed in this commit.

The migration definitively removes the column and indexes that where associated with those fields.

https://www.pivotaltracker.com/story/show/168893683


MKodde:
A very good catch! I was not aware that the annotations and the dropping of the columns needs to be in separate releases. But now that you mention this, we had similar issues with the removal of the coin columns.

pablothedude:
Thanks for splitting this up and creating a reminder! Would it be possible to label the Pivotal issue also 6.3 and add some links back and forth to the PR's and Pivotal issues. That would certainly be beneficial when releasing 6.3 and was also very helpful whit the Coin migration regressions in an earlier release.

MKodde:
Yes, updated the tickets! And for future reference. The 6.3 chore to add the migration can be found in PR #835

PR #836, opened 2020-03-04
Disassemble the EB debug feature

MKodde:
It was not used, and provided minimal additional information. Removing it seemed logical.

Regarding the debug Twig extension, the var_dump function was removed. But the print_r and var_export functions have been left in place. As the debug route (used to test EB from the home page) still uses it to create debug info for the helpdesk mail.

https://www.pivotaltracker.com/story/show/164757722

### DIFF
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -21,7 +21,6 @@ use OpenConext\EngineBlock\Metadata\Entity\IdentityProvider;
 use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
 use OpenConext\EngineBlock\Validator\AllowedSchemeValidator;
 use OpenConext\EngineBlockBundle\Exception\ResponseProcessingFailedException;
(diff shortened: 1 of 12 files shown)


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -692,16 +689,16 @@ public function send(
...
-    {
-        $schemaUrl = 'http://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd';
-        if ($this->_server->getConfig('debug') && ini_get('allow_url_fopen') && file_exists($schemaUrl)) {
-            $dom = DOMDocumentFactory::fromString($xml);
-            if (!$dom->schemaValidate($schemaUrl)) {
-                throw new Exception('Message XML doesnt validate against XSD at Oasis-open.org?!');
-            }
-        }
-    }
+//    protected function validateXml($xml)
```
pablothedude: Could this be removed?

MKodde: Thanks for seeing this. That function indeed had to be removed after commenting it out!

PR #839, opened 2020-03-27
Update npm dependencies

pablothedude:
- [x] Update all dependencies
- [x] Add consent test endpoint + visual regression test
- [x] Automatic resolutions with NPM

It's possible to test the pages with the visual regression tool.
The tool is already working on master. But be aware that two test will
fail. A broken test of the generic error page which triggers an error
of the same generic error page was the cause and is fixed here.


pablothedude:
Initially I only wanted to update the required dependencies. But I had too much conflicts, so the only way seemed to be the way up. The only problem was `tippy.js` and the `js-cookie` dependency because they are required but were not tested.

The reason I could upgrade them was because the NPM build pipeline only uses cli commands in the package.json so if their footprint would change they would fail.

Mabe we should reconsider their usage or add tests in order to allow easy upgrading of Npm assets so we could do that more actively instead of reactive on an event of a security warning. I'll create a ticket for that.

pablothedude:
https://www.pivotaltracker.com/story/show/172103012

PR #840, opened 2020-03-30
Add visual consent test

pablothedude:
Add a simple visual regression test for the consent screen.
So the upgrading of build tools/dependencies could be more easily
verified.

The security warnings will be fixed in: #839

### DIFF
--- a/theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
+++ b/theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
@@ -1,26 +1,22 @@
-const timeout = 15000;
+const timeout = 5000;
 const {toMatchImageSnapshot} = require('jest-image-snapshot');
 
 // Extend Jest expect
 expect.extend({toMatchImageSnapshot});
 
-const footerDifferences = [
-    {
-        name: 'all-buttons-visible',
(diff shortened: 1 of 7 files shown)


theme/material/javascripts/tests/visual-regression/error-page/Footer.test.js
```diff
@@ -1,68 +1,45 @@
...
+    ],[
+        'only-support-email-hidden',
+        'https://engine.vm.openconext.org/functional-testing/feedback?template=unable-to-receive-message'
+    ],[
+        'only-wiki-hidden',
+        'https://engine.vm.openconext.org/functional-testing/feedback?template=missing-required-fields&feedback-info={"requestId":"5cb4bd3879b49","artCode":"31914", "ipAddress":"192.168.66.98","serviceProvider":"https://current-sp.entity-id.org/metadata", "serviceProviderName": "OpenConext Drop Supplies SP","identityProvider":"http://mock-idp"}'
+    ],[
+        'support-email-and-wiki-button-hidden',
+        'https://engine.vm.openconext.org/functional-testing/feedback?template=missing-required-fields'
+    ],
```
MKodde: Maybe for in the future, but adding the different nameid format renderings and the IdP provided messag would be a nice addition to this visual regression test.

pablothedude: Thanks, I'll keep that in mind.

PR #841, opened 2020-04-01
Add Portuguese language support

pablothedude:
This will merge the Portuguese translations into a new feature branch
with all recent changes. This is mostly concerning the changed metadata
endpoints and SFO callout logic.

This PR will only add PT as language. Some tasks still need to be done
but will be addressed later on in separate PR's.

- [ ] Fix checks with only allowed languages (helper)
- [ ] Check and validate missing PT translations
- [ ] Update migrations and test and validate with metadata push
- [ ] Move translations from parameters.yml to translations


src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ProxiedIdentityProvider.php
```diff
@@ -98,35 +98,4 @@ public function getSingleSignOnServices(): array
...
-     * Buisiness rule:
-     * displayname:en, name:en
-     */
-    public function getDisplayNameEn(): string
-    {
-        if (empty($this->entity->getDisplayNameEn())) {
-            return $this->entity->getNameEn();
-        }
-        return $this->entity->getDisplayNameEn();
-    }
```
MKodde, line 131: The removal of these methods seem unrelated to the PT language addition? Please explain.

MKodde:
Are you planning on addressing the open todo's on this PR or on successive PR(s)?

pablothedude:
I will address this later on in separate PR's.

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ProxiedIdentityProvider.php
pablothedude: This logic was moved to the AbstractIdentityProvider as it seemed default behaviour to present a displayname. Also the displayname seems to be unrelated to proxying so therefore I removed it from this decorator.

MKodde: Did you check with the product manager if this behavioural change is expected behavior? It was added specifically for the IdP metadata which sometimes contained empty titles. Moving it to the abstract for all other IdP to benefit from this is probably fine. But might not?

MKodde:
All systems go, except for this question: https://github.com/OpenConext/OpenConext-engineblock/pull/841#discussion_r405556312

src/OpenConext/EngineBlock/Metadata/Factory/Decorator/ProxiedIdentityProvider.php
pablothedude: Validated the expected behaviour with @thijskh and this is always expected behaviour.

MKodde: In that case it's fine with me!

PR #842, opened 2020-04-06
Add supported language provider

pablothedude:
In order to configure supported languages a helper is introduce to
enable only a subset of available languages in EB.

https://www.pivotaltracker.com/story/show/162414108

- [x] Fix checks with only allowed languages (helper)
- [x] Check and validate missing PT translations
- [x] Update migrations and test and validate with metadata push
- [ ] Move translations from parameters.yml to translations [PR](https://github.com/OpenConext/OpenConext-engineblock/pull/844)


src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
```diff
@@ -0,0 +1,59 @@
...
+            throw new UnsupportedLanguageException('No supported languages found');
+        }
+
+        $this->supportedLanguages = array_values($languages);
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getSupportedLanguages()
```
MKodde, line 55: Are we able yet to use PHP 7 return type hints? I do not recall from what version we are allowed to fully drop PHP 5.

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
```diff
@@ -0,0 +1,59 @@
...
+            if (in_array($language, $availableLanguages)) {
+                $languages[$language] = $language;
+            } else {
+                throw new UnsupportedLanguageException('A supported language is enabled while not available');
+            }
+        }
+
+        if (empty($languages)) {
+            throw new UnsupportedLanguageException('No supported languages found');
+        }
```
MKodde, line 47: I suggest you add an additional validation to the start of the constructor where you verify that AvaliableLanguages is not empty. There you can throw a meaningfull message like:

`No active languages are configured, please configure the available languages in 'fffoobbar.yaml'`

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
```diff
@@ -0,0 +1,59 @@
...
+     * @param string[] $enabledLanguages
+     */
+    public function __construct(array $availableLanguages, array $enabledLanguages)
+    {
+        $languages = [];
+        foreach ($enabledLanguages as $language) {
+            if (in_array($language, $availableLanguages)) {
+                $languages[$language] = $language;
+            } else {
+                throw new UnsupportedLanguageException('A supported language is enabled while not available');
```
MKodde, line 41: Please improve the readability of this exception message by stating the unexisting language(s). Resulting in something along these lines:

`This language can not be enabled: %s. Please check foobar.yaml.`

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * @param string[] $availableLanguages
+     * @param string[] $enabledLanguages
+     */
+    public function __construct(array $availableLanguages, array $enabledLanguages)
+    {
+        $languages = [];
+        foreach ($enabledLanguages as $language) {
+            if (in_array($language, $availableLanguages)) {
```
MKodde, line 38: Instead of iterating over the enabled languages, you might be able to use one of the array `intersect` or `diff` functions to get a full overview of which languages are not supported. That way you can make the exception message much more specific.

src/OpenConext/EngineBlockBundle/Resources/config/services.yml
```diff
@@ -181,10 +181,16 @@ services:
             - "%maximum_authentication_procedures_allowed%"
             - "%time_frame_for_authentication_loop_in_seconds%"
 
+    engineblock.language_support_provider:
+        class: OpenConext\EngineBlockBundle\Localization\LanguageSupportProvider
+        arguments:
+            - ['en', 'nl', 'pt']
+            - ['en', 'nl']
```
MKodde: Surely these arguments should be configurable parameters. Or was this a hard requirement?

tests/integration/OpenConext/EngineBlockBundle/Localization/LanguageSupportProviderTest.php
```diff
@@ -0,0 +1,78 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlockBundle\Localization;
+
+use OpenConext\EngineBlockBundle\Exception\UnsupportedLanguageException;
+use PHPUnit\Framework\TestCase;
+
+class LanguageSupportProviderTest extends TestCase
```
MKodde, line 24: I love this test. Maybe add a `test data provider` based test case to further test different input options?

tests/unit/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/LocaleTest.php
```diff
@@ -38,6 +39,11 @@ public function test_get_locale_falls_back_on_default()
         $this->assertEquals('nl', $this->buildLocale('fr', true)->getLocale());
     }
 
+    public function test_get_supported_locales()
+    {
+        $this->assertEquals(['en', 'nl'], $this->buildLocale('fr', true)->getSupportedLocales());
+    }
```
MKodde, line 45: This seems to be a copy of an edge case test using the fallback on default language when an unsupported language is requested. Perhaps for this new test you base base your test on the happy flow? It was rather confusing to me while reviewing this.

database/DoctrineMigrations/Version20200422145000.php
```diff
@@ -0,0 +1,29 @@
...
+final class Version20200422145000 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sso_provider_roles_eb5 ADD name_pt VARCHAR(255) NOT NULL AFTER name_en, ADD description_pt VARCHAR(255) NOT NULL AFTER description_en, ADD display_name_pt VARCHAR(255) NOT NULL AFTER display_name_en, ADD organization_pt_name LONGTEXT DEFAULT NULL COMMENT \'(DC2Type:object)\' AFTER organization_en_name, ADD keywords_pt VARCHAR(255) NOT NULL AFTER keywords_en, ADD support_url_pt VARCHAR(255) DEFAULT NULL AFTER support_url_en
+        ');
+    }
```
MKodde, line 20: Was this migration added accidentally? Or are we actually adding i18n translation fields to the roles table? Also please review if we should use PHP 7 constructs in the files we are adding. If we do, please review your other code to also use PHP 7 constructs.

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
pablothedude: Even if we are allowed I still would like to have support for both versions till EB can choose for itself which version to use with the latest OC-Deploy. 
https://www.pivotaltracker.com/story/show/170014115

Because we are the ones that maintain the codebase I would like to be able to do so instead of being cutting edge just for the sake of it. I think that currently support for both PHP5 and 7 still outweighs strict typing in a small part of our codebase when we still have to cope with the legacy codebase.

pablothedude: If OC-deploy is fixed and only supports PHP7 I will drop PHP5 support.

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
pablothedude: The available languages are hardcoded and should only be updated by us when we add support for a new language so it won't be that valuable. See: `servcies.yml`.

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
pablothedude: Updated the message to: `No active languages are configured, please check your configuration`.  I think the users who see this will be smart enough to know what they've just changed themselves.

src/OpenConext/EngineBlockBundle/Resources/config/services.yml
pablothedude: This was something I did hard-code to prevent merge conflicts because the ini parameter config still needed to be merged back. But I forgot to implement them afterwards. So, thanks!

tests/integration/OpenConext/EngineBlockBundle/Localization/LanguageSupportProviderTest.php
pablothedude: Maybe later on, I think for now this will be sufficient.

tests/unit/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/LocaleTest.php
pablothedude: Added a more happy flow test.

database/DoctrineMigrations/Version20200422145000.php
pablothedude: No this wasn't added accidentally. This was lacking in the original PT PR. I've first fixed all merge conflicts there but I couldn't test the migrations then because of changes done in the meantime. This PR was the first possibility to thoroughly test this and therefore I've added it here.

I think that after this sprint it will be safe to stop using PHP5 completely.

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
MKodde: Allright, I was slightly confused as you did introduce some php 7 constructs in other files. 

I fully agree with you by the way.

app/config/parameters.yml.dist
```diff
@@ -22,6 +22,11 @@ parameters:
         - 192.168.1.1
         - 10.0.0.1
 
+    ## The anebaled languages, currently 'en', 'nl' and 'pt' are supported
```
MKodde, line 25: typo: ~anebaled~ enabled

src/OpenConext/EngineBlockBundle/Localization/LanguageSupportProvider.php
MKodde: I don't mind leaving it like this. Using the built in array functions would decrease cyclomatic complexity. Which might be a good practice. But complexity is low enough, so I'm fine with the current situation.

database/DoctrineMigrations/Version20200422145000.php
MKodde: Okay, for the sake of BC, then please remove the PHP 7 lang constructs from this migration.

src/OpenConext/EngineBlockBundle/Resources/config/services.yml
```diff
@@ -181,10 +182,16 @@ services:
             - "%maximum_authentication_procedures_allowed%"
             - "%time_frame_for_authentication_loop_in_seconds%"
 
+    engineblock.language_support_provider:
+        class: OpenConext\EngineBlockBundle\Localization\LanguageSupportProvider
+        arguments:
+            - ['en', 'nl', 'pt']
```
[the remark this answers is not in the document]
pablothedude: This was how it originally was implemented. We could easily change it later on so I'll leave it here for now.

database/DoctrineMigrations/Version20200422145000.php
MKodde: I still see the `: void` return type definition. Maybe you forgot to push?

src/OpenConext/EngineBlockBundle/Resources/config/services.yml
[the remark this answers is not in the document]
MKodde: Okay, it just felt out of place. No problem leaving it here for now.

PR #845, opened 2020-04-29
Log original NameID and given LOA on successful login

pablothedude:
The original NameID and attained LOA were missing and are now added
to the 'login granted' log.

https://www.pivotaltracker.com/story/show/171404731

The node-sass security warning is fixed in https://github.com/OpenConext/OpenConext-engineblock/pull/847.

### DIFF
--- a/library/EngineBlock/Corto/Filter/Command/LogLogin.php
+++ b/library/EngineBlock/Corto/Filter/Command/LogLogin.php
@@ -48,12 +48,16 @@ class EngineBlock_Corto_Filter_Command_LogLogin extends EngineBlock_Corto_Filter
         // Remove the SP that is our next hop
         array_pop($requesterChain);
 
+        $originalResponse = ($this->_response->getOriginalResponse() ? $this->_response->getOriginalResponse() : $this->_response);
+
         $this->authenticationLogger->logLogin(
             $this->_serviceProvider,
             $this->_identityProvider,
             $this->_collabPersonId,
             $this->_request->getKeyId(),
-            $requesterChain
+            $requesterChain,
+            $originalResponse->getIntendedNameId(),
+            $originalResponse->getAttainedLoa()
         );
     }
 }
--- a/library/EngineBlock/Corto/Module/Service/ProcessedAssertionConsumer.php
+++ b/library/EngineBlock/Corto/Module/Service/ProcessedAssertionConsumer.php
@@ -81,11 +81,6 @@ class EngineBlock_Corto_Module_Service_ProcessedAssertionConsumer implements Eng
         }
 
          // Done processing! Send off to SP
-        $nextResponse = $currentProcess->getResponse();
-        $response->setDestination($nextResponse->getDestination());
-        $response->setDeliverByBinding($nextResponse->getOriginalIssuer());
-        $response->setOriginalIssuer($nextResponse->getOriginalBinding());
-
         $this->_server->unsetProcessingMode();
 
         $sentResponse = $this->_server->createEnhancedResponse($receivedRequest, $response);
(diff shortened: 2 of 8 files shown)


library/EngineBlock/Corto/Filter/Command/LogLogin.php
```diff
@@ -48,12 +48,16 @@ public function execute()
         // Remove the SP that is our next hop
         array_pop($requesterChain);
 
+        $originalResponse = ($this->_response->getOriginalResponse() ? $this->_response->getOriginalResponse() : $this->_response);
```
MKodde, line 51: In which situations would the OriginialResponse *not* be available? You might want to make this more clear. Now it looks like a fallback mechanism that simply ensures a name id is set regardless of it's origin.

library/EngineBlock/Corto/Module/Service/StepupAssertionConsumer.php
```diff
@@ -233,6 +233,7 @@ private function updateProcessingStateLoa(EngineBlock_Saml2_AuthnRequestAnnotati
     {
         $processStep = $this->_processingStateHelper->getStepByRequestId($receivedRequest->getId(), ProcessingStateHelperInterface::STEP_STEPUP);
         $processStep->getResponse()->getAssertion()->setAuthnContextClassRef($loa->getIdentifier());
+        $processStep->getResponse()->setAttainedLoa($loa->getIdentifier()); // set attained loa so this could be logged later on
```
MKodde, line 236: Correct me if I'm wrong. The attained LoA is the 'EngineBlock' LoA, mapped back from the received LoA from the remote step up provider. If that's the case, this code looks to be doing just that.

pablothedude: It's indeed the internal EB LOA it has attained after handling the Stepup callout.

library/EngineBlock/Corto/Filter/Command/LogLogin.php
pablothedude: When the `eb.run_all_manipulations_prior_to_consent` feature is enabled the response isn't decorated with the original response because there is no response. This is happening because of the output filtering then taking place in the input filtering this will log a successfull login on the acs location before consent. So yes, this is a fallback needed to support just that.

I've consulted Thijs about logging this prior to consent and that shouldn't happen after all so I will file an issue in Pivotal to fix this bug. So then this fallback could be removed completely.

library/EngineBlock/Corto/Module/Service/ProcessedAssertionConsumer.php
```diff
@@ -83,8 +83,8 @@ public function serve($serviceName, Request $httpRequest)
          // Done processing! Send off to SP
         $nextResponse = $currentProcess->getResponse();
         $response->setDestination($nextResponse->getDestination());
-        $response->setDeliverByBinding($nextResponse->getOriginalIssuer());
-        $response->setOriginalIssuer($nextResponse->getOriginalBinding());
+        $response->setDeliverByBinding($nextResponse->getOriginalBinding());
+        $response->setOriginalIssuer($nextResponse->getOriginalIssuer());
```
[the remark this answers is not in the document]
pablothedude: When I did refactor the legacy code to be able to support Stepup callouts I've mixed them up. I've done some testing and this doesn't seems to be needed at all because the `MessageAnnotationDecorator::sspMessage` seem to be used for the issuer and the destination and the `RepsonseAnnotationDecorator` is doing that for the Binfing method based on the initial AuthnRequest. So I've removed them completely.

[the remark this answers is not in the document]
MKodde: Okay, thats a somewhat unexpedted conclusion, +1 for cleaning it up!

PR #846, opened 2020-05-01
Remove old browser support

pablothedude:
There was still browser support in sass/less for very old browsers.
This support could be dropped now.

The node-sass security warning is fixed in https://github.com/OpenConext/OpenConext-engineblock/pull/847.

See:
https://www.pivotaltracker.com/story/show/165716857


theme/material/stylesheets/application/components/_modal.css.sass
```diff
@@ -11,14 +11,9 @@
...
-    margin: 1em
 
 #request-access
   @extend %box
   margin: auto
   padding: 0.5rem
   max-width: $container-width
-  .lt-ie9 &
-    padding: 0.5em
-    max-width: $container-width-em
```
MKodde, line 24: According to some greppery, `container-width-em` is no longer used after your cleanup. You should remove the definition as well.

`OpenConext-engineblock/theme/material/stylesheets/shared/_vars.css.sass:37`

theme/material/stylesheets/application/modules/_app-search.css.sass
```diff
@@ -61,18 +52,7 @@
     border: none
     background: url('../images/search.svg') no-repeat center center
       size: 1rem auto
-    .lt-ie9 &
-      text-indent: -9999em
-      width: 3.5em
-      height: 3.5em
-    .lt-ie10 &
-      top: 0.5em*2
-      margin-top: 40px
```
MKodde, line 70: Personally, I'm totally fine with also dropping IE9 support. [The story](https://www.pivotaltracker.com/story/show/165716857) this PR seems to be related to only mentions dropping IE8 support. Did you discuss this scope widening? Or should the `.lt-ie10 ` removals be reverted?

theme/material/templates/layouts/scripts/default.html.twig
```diff
@@ -1,10 +1,6 @@
 {% import "@themeLayouts/scripts/language-switcher.html.twig" as languageSwitcher %}
 <!DOCTYPE html>
-<!--[if lt IE 7]> <html class="no-js lt-ie10 lt-ie9 lt-ie8 lt-ie7" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
-<!--[if IE 7]> <html class="no-js ie7 lt-ie10 lt-ie9 lt-ie8" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
-<!--[if IE 8]> <html class="no-js ie8 lt-ie10 lt-ie9" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
-<!--[if IE 9]> <html class="no-js ie9 lt-ie10" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
-<!--[if gt IE 9]> <html class="no-js" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
+<!--[if IE 10]> <html class="no-js" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
```
MKodde, line 3: Removing the `[if gt IE 9]` condition and replacing it with an IE10 specific condition, results in IE 11 not getting the `no-js` class. This might interfere in no js behavior for IE 11 users (which we do support as far as I know).

Which makes me wonder if we need this `no-js` support for edge browsers. A quick grep on `no-js` shows me a JS removes that class. If the class is not removes this indicates there is no JS support. In that case we do some additional styling to show submit buttons and such. I'd say that we want that behavior on all browsers. 

By the way the current setup seems to only add the `<html>` opening element on IE10, and non IE browsers. Older IE browsers would now render an erroneous html document.

theme/material/stylesheets/application/modules/_app-search.css.sass
pablothedude: FYI: `https://www.pivotaltracker.com/story/show/165377239/comments/201708132`
I shall validate this with Thijs to be sure.

theme/material/templates/layouts/scripts/default.html.twig
pablothedude: IE10 does support conditional comments only in standards mode and IE11 doesn't support this at all, so this seemed completely safe to do. I've considered to drop it altogether but left it in place for IE10 to support IE10 not in standards mode.

I've tested this in IE11 and IE11 is using the `html` element with `no-js` class. Also in IE10 that is the expected behaviour.

Could you elaborate how you've tested this so I could reproduce the situation you're describing?

MKodde: I have only code reviewed this. But looked further into this just now. I overlooked the `<!--[if !IE]><!-->` conditional comment. That should allow other (non IE) browsers to use the default html opening element. 

Still having no html opening element on older IE browsers seems harsh in my opinion.

I tested this in IE9 using Browserstack. Concider the following code snippet

```
<!DOCTYPE html>
<!--[if lt IE 10]> <html id="oldie" class="no-js" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
<!--[if IE 10]> <html id="wosname" class="no-js" lang="{{ locale()|escape('html_attr') }}"> <![endif]-->
<!--[if !IE]><!--> <html id="ie10" class="foobar no-js {{ userAgent('safari')|escape('html_attr') }}" lang="{{ locale()|escape('html_attr') }}"> <!--<![endif]-->
<head>
    <script>
        window.onload = function() {
            console.log(document.getElementById('ie10'));
            console.log(document.getElementById('wosname'));
            console.log(document.getElementById('oldie'));
        }
    </script>
....
```

The console output would be:

```
LOG: null
LOG: null
LOG: HtmlElement
```

I'm not certain what kind of fallback behavior kicks in in these older browsers when no HTML opening tag is set, especially using conditional comments. But I think we should add an `<!--[if lt IE 10]>` comment.

MKodde:
See my [reaction](https://github.com/OpenConext/OpenConext-engineblock/pull/846#discussion_r423055022).

Also, please rephrase the `f` commit.

PR #847, opened 2020-05-01
Fix NPM security warning

pablothedude:
Update used assets to latest and resolve some problems with a
resolution.

### DIFF
--- a/theme/package.json
+++ b/theme/package.json
@@ -39,7 +39,7 @@
     "jest": "^25.2.4",
     "jest-image-snapshot": "^3.0.1",
     "jshint": "^2.11.0",
-    "node-sass": "^4.13.1",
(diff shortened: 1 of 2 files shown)


theme/package.json
```diff
@@ -54,6 +54,7 @@
     "tippy.js": "^6.1.0"
   },
   "resolutions": {
-    "minimist": "1.2.3"
+    "minimist": "1.2.3",
```
MKodde, line 57: Did you by chance tets if the old resolution could be removed? If not this would be a good talking point at the pro-active NPM meet we have sceduled for wednesday. Periodically testing if resolutions can be removed would be golden.

pablothedude: The first thing I did was remove the resolution but unfortunately the resolution was still needed.

PR #848, opened 2020-05-01
If the SP provided a RequestedAuthnContext in the AuthnRequest, replicate this to the IdP

thijskh:
This allows the SP to communicate e.g. desired authentication parameters (strength) to the IdP in a generic way.


thijskh:
Since ADFS does not return a related AuthnContextClassRef in the returned assertion, the value of this addition is (too) limited for now.

thijskh:
Use cases have been identified.

thijskh:
That is indeed the case but is fine for now. If a use case emerges where we want to combine both, we can redesign then.

PR #851, opened 2020-05-13
Make support urls translatable

pablothedude:
Some urls still needed to be translatable after
the `application.ini` changes. This is done now.


pablothedude:
I've removed it I'll add a task to the backlog to see if more could then be cleaned up. That way 2.6 could be released and then I'll start a new sprint with this task.

MKodde:
The removal of the Edugain document class might deserved a dedicated commit. As it relates to, but not fully fits the commit it is in now. Feel free to leave as is.

Issue #852, opened 2020-05-14
Error on omitting Stepup-certificate (v5.13)


tvdijen:
This may already been covered by the migration from `/etc/openconext/engineblock.ini` to YAML config file, but in that case this issue may be kept for future reference to other people bumping into this issue. It probably exists as of 5.13 where the new stepup-functionality was added.

Case:
- I upgraded from 5.12 to 6.1.3
- I didn't add the new Stepup-config, simply because I'm not using Stepup for this EB-instance.
- My certificate is not in the default `/etc/openconext/engineblock.crt` location.

Composer install went fine, but on accessing EB from the browser in an authflow the following error occured:

```
May 14 20:16:25 sv1811042 EBLOG[11331]: [2020-05-14 20:16:25] app.DEBUG: Caught Exception "OpenConext\EngineBlock\Exception\InvalidArgumentException":"Keyfile '/etc/openconext/engineblock.crt' should be a valid file" {"session_id":"h7c9ljfn4q0k13erctj8ohsnt7","request_id":"5ebd8af9728c2"} []
May 14 20:16:25 sv1811042 EBLOG[11331]: [2020-05-14 20:16:25] engineblock.ERROR: Keyfile '/etc/openconext/engineblock.crt' should be a valid file | Caught Unhandled generic exception [..]
```

I'd argue that EB should be perfectly usable without any of the Stepup-stuff configured..
I eventually solved it by setting `stepup.gateway.sfo.keyFile` to an actual certificate.. The other stepup-settings are still left out and it's working fine now.

Perhaps changing the default for this setting to the value of `encryption.keys.default.publicFile` would be a solid fix?

pablothedude:
Hi Tim, thanks for the report. I've created a PR with a fix for the described issue.
https://github.com/OpenConext/OpenConext-engineblock/pull/853

Are you able to verify if this will work for you? I'll try to do my best to let this land in the next patch version of 6.2.

tvdijen:
No bueno.. If left out from the configuration it will try and load `/etc/openconext/engineblock.crt`.
It's still the same assertion failing, so that means it must be fed with some default value elsewhere.

pablothedude:
Hi Tim, the validation is moved from the constructor completely. The validation now only happens JIT when an entity is configured to use the Stepup callout functionality. So if you don't configure it it won't get validated. This should help you and would suit the majority of the Openconext users.

[...]

PR #853, opened 2020-05-15
Allow empty sfo key

pablothedude:
Allow empty SFO key to support EB without SFO. This change will allow SFO configuration without a valid key needed in the configuration.

https://github.com/OpenConext/OpenConext-engineblock/issues/852

### DIFF
--- a/docs/stepup_callout.md
+++ b/docs/stepup_callout.md
@@ -1,0 +1,33 @@
+# Engineblock Stepup second factor integration
+
+It's possible to require second factor authentication in EngineBlock using OpenConext StepUp in second factor only (SFO) mode. When configured correctly, EngineBlock will utilize StepUp Gateways SFO to do a second factor callout. In order to do so the EB configuration needs to be configured in a certain way to allow this on a per IdP or per SP basis. The following coin metadata attributes need to be passed from the Manage (or other software) instance that is pushing EngineBlock metadata into EngineBlock.
+
+
+## Engineblock metadata configuration
+### SP
+#### metadata:coin:stepup:allow_no_token
+**Type:** boolean
+
+To continue with LOA 1 if no second factor token found
+
+#### metadata:coin:stepup:requireloa
+The LOA minimal required
+
+**Type:** boolean
+
+
+
+
+### IdP
+#### metadata:coin:stepup_connections
+
+**Type:** object
+* name: _entityId_,
+* level: _requiredLoa_
+
+An entry per SP which should use the SFO capabilities.
+
+
+## Engineblock global configuration
+The EngineBlock installation also needs additional configuration in order to facilitate the SFO second factor authentications. For details on these configuration settings, please review the SFO section in the [app/config/parameters.yml.dist](app/config/parameters.yml.dist) file.
+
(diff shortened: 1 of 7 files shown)


MKodde:
This looks like a good and pragmatic way to solve the issue described in #852. We can maybe take this a step further:

1. Remove/unset the default SFO settings in the parametesr.yml.dist file. So EB with SFO is not configured out of the box. I think using this feature is not common for EB users except for ourselves. Please ask Thijs for input on this if you haven't done so already.
2. Add a documentation entry in the `docs` folder explaining the configuration of the SFO featuer (basically moving the already documented `dist` file documentation to a markdown file.

tests/unit/OpenConext/EngineBlock/Stepup/StepupGatewayLoaMappingTest.php
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-namespace OpenConext\EngineBlockBundle\Tests;
+namespace OpenConext\EngineBlock\Stepup;
```
MKodde, line 19: Did you intentionally remove the /Tests portion from the namespace? I think the other tests all reside in a `Test` namespace. I think you removed the trailing `Tests` namespace to prevent the PSR4 namespace warning. Which is a good! I'm opening a low priority ticket in Pivotal to look into this at some later stage.

docs/stepup_callout.md
```diff
@@ -0,0 +1,34 @@
+# Engineblock Stepup second factor integration
+
+It is possible to have second factor support in EngineBlock. This is done by utilizing the SFO capabilities in the Stepup Gateway to do a callout
+to Stepup to perform second factor validation. In order to do so the EB configuration needs to be configured with coins to allow this
+on a per IdP or per SP basis.
```
MKodde, line 5: Proposal to make the intro text somewhat clearer:

_It's possible to require second factor authentication in EngineBlock using OpenConext StepUp in second factor only (SFO) mode. When configured correctly, EngineBlock will utilize StepUp Gateways SFO to do a second factor callout. In order to do so the EB configuration needs to be configured in a certain way to allow this on a per IdP or per SP basis. The following coin metadata attributes need to be passed from the Manage (or other software) instance that is pushing EngineBlock metadata into EngineBlock._

End suggestion for the introduction.

Then I would move the final paragraph into this introduction text. Or give it a heading like you did with the other paragraphs to draw more attention to it. 

Maybe this paragraph can be added to the introduction:

_The EngineBlock installation also needs additional configuration in order to facilitate the SFO second factor authentications. For details on these configuration settings, please review the `SFO` section in the `app/config/parameters.yml.dist` file._

tests/unit/OpenConext/EngineBlock/Stepup/StepupGatewayLoaMappingTest.php
pablothedude: Yes, it was indeed to  prevent the PSR4 warnings, thanks for opening up a PR for this!

docs/stepup_callout.md
pablothedude: Thank you for the suggestion, I've changed it accordingly

tests/unit/OpenConext/EngineBlock/Stepup/StepupGatewayLoaMappingTest.php
MKodde: FYI: `https://www.pivotaltracker.com/story/show/172935406`

MKodde:
Thanks for addressing the doc suggestions, feel free to merge when Travis is also in agreement.

PR #857, opened 2020-05-27
Change symfony cache path to reflect deploy path

pablothedude:
In deploy a different cache and log path are used. When performing a
clean install this will result in errors because an unknown path is
being used. This change will resolve that.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -64,7 +64,7 @@ class AppKernel extends Kernel
         // In the dev & test environments use a folder outside the shared filesystem. This greatly improves cache clear
         // and warmup time.
         if ($this->getEnvironment() === 'dev' || $this->getEnvironment() === 'test') {
-            return sprintf('/tmp/eb_cache/%s', $this->getEnvironment());
+            return sprintf('/tmp/engineblock/cache/%s', $this->getEnvironment());
         }
 
         return $this->rootDir . '/cache/' . $this->environment;


app/AppKernel.php
```diff
@@ -65,14 +65,20 @@ public function getCacheDir()
...
 
         return $this->rootDir . '/cache/' . $this->environment;
     }
 
     public function getLogDir()
     {
+        // In the dev & test environments use a folder outside the shared filesystem. This greatly improves cache clear
+        // and warmup time.
+        if ($this->getEnvironment() === 'dev' || $this->getEnvironment() === 'test') {
+            return sprintf('/tmp/engineblock/log/%s', $this->getEnvironment());
```
MKodde, line 79: Is this really required for logging? In `dev` mode I really appreciate the logs being in the app/log folder. This enables me to view the logs without having to ssh into my box.

I also think the performance penalty regarding logs is very small compared to cache.

If this changes ends up being merged. Please change the `.gitignore` accordingly. My personal preference would be to at least keep the `dev` logs in application.

pablothedude: Will a symlink work?

MKodde: It probably would, but what's the point of moving the dev logs outside the project in the first place? Maybe this should be fixed in Ansible? A dev only task could perhaps change the default behaviour?

pablothedude: I'm still wondering why things are logged in your ptojectroot because EB default will log all to syslog? I'll remove the log part from this PR and will remove the creation of the log dir in Deploy.

pablothedude: Reverted the log path change and changed the path to the projectroot path in Deploy:
https://github.com/OpenConext/OpenConext-deploy/pull/284

MKodde: For develop, it is logged to syslog, but the test env writes to ./app/logs/test

MKodde (approved):

PR #859, opened 2020-07-02
Assemble authcontextclassref combinations

MKodde:
Added ability to read the `mfa_entities` data from the Manage metadata push.

For some reason Travis no longer reports back to Github on the build status. But the link below can be used for details about the QA build.

Note: Build breaks on security issue that should be resolved in a separate PR
See: https://www.pivotaltracker.com/story/show/172718043
See: https://travis-ci.org/github/OpenConext/OpenConext-engineblock/builds/704306933

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
@@ -615,7 +615,15 @@ class PushMetadataAssembler implements MetadataAssemblerInterface
         if (!isset($connection->mfa_entities)) {
             return [];
         }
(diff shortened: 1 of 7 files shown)


src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
```diff
@@ -607,4 +609,15 @@ private function assembleAttributeReleasePolicy(stdClass $connection)
             )
         );
     }
+
+    private function assembleMfaEntities(stdClass $connection): array
+    {
+        if (!isset($connection->mfa_entities)) {
+            return [];
+        }
+        $entities = json_decode(json_encode($connection->mfa_entities), true);
```
pablothedude: It's not very clear why this is needed and makes the code less readable.
I will remove this.

PR #861, opened 2020-07-24
Add authncontextclassref to SP if configured in IdP

pablothedude:
If a configured authncontextclassref is configured in the
IdP for the current SP the authncontextclassref needs to be
set to the response.

https://www.pivotaltracker.com/story/show/172718137

PR #863, opened 2020-07-24
Test unsolicited presence of authcontextclassref

pablothedude:
**Be aware this is build on top of #861 so this has te be merged back first!**

This will prove that for the unsolicted flow the authncontextclassref
is set when configured for the IdP SP combination.

https://www.pivotaltracker.com/story/show/172719874

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/MfaAuthncontextclassref.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/MfaAuthncontextclassref.feature
@@ -23,3 +23,9 @@ Feature:
     Then the url should match "functional-testing/SSO-IdP/sso"
       And the response should not contain "http://schemas.microsoft.com/claims/multipleauthn"
 
+  Scenario: The configured authncontextclassref should be set if configured with the IdP configuration mapping also for unsolicited logins
+    Given the IdP "SSO-IdP" is configured for authncontextclassref "http://schemas.microsoft.com/claims/multipleauthn" for SP "SSO-SP"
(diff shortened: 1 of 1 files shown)


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/MfaAuthncontextclassref.feature
```diff
@@ -23,3 +23,9 @@ Feature:
     Then the url should match "functional-testing/SSO-IdP/sso"
       And the response should not contain "http://schemas.microsoft.com/claims/multipleauthn"
 
+  Scenario: The configured authncontextclassref should be set if configured with the IdP configuration mapping also for unsolicited logins
```
MKodde, line 26: This sentence needs some work, I find it hard to understand what message is conveyed here.

Maybe something along the lines of:

`The configured AuthnContextClassRef should be present on unsolicited single sign on` 

In my opinion the other scenarios already make clear that the ref should only be present when configured. Alternatively you could describe that portion in the feature description. As all scenarios will probably adhere to this principle.

pablothedude: Changed it in https://github.com/OpenConext/OpenConext-engineblock/pull/864. Will discus IRL if I get there if that change is sufficient.

MKodde: As found in #864
> Scenario: The configured authn method should also be set for unsolicited logins if configured in the IdP configuration mapping

Sounds good to me, personally I wouldn't use the `authn method. I'd simply use `authentication method`. No need to use unnecessary abbreviations.

PR #868, opened 2020-07-29
added new Portuguese translations

domgon:
We notice that were missing some messages from newer messages


pablothedude:
Hi @domgon, thanks for your contribution. While reviewing the changes I saw that it seems that something was wrong and that it looked like the translations were copied into the original file and that the old ones didn't seemed to be correctly overwritten.

I therefore opened up a new PR with the fix and with the changes in this PR. In the bottom of the file some translations were added that were added in the meantime for the footers on the error pages. Could you verify if the word `Mail` which is used on a button on the bottom of the page is correct.

Also the translation for `error_stepup_callout_user_cancelled_desc` seem to have a duplicate.

https://github.com/OpenConext/OpenConext-engineblock/pull/869

pablothedude:
Closed in favour of https://github.com/OpenConext/OpenConext-engineblock/pull/869

PR #872, opened 2020-07-30
Bump yargs-parser and minimist to mitigate vulnerability

pablothedude:
minimist 1.2.5
yargs-parser 13.1.2


MKodde:
Also, have you investigated why the PHP 7.2 fails on the solicited/unsolicited test? On 7.4 it seems to work as intended

PR #875, opened 2020-08-06
Show proxied SP and proxy in feedback info

MKodde:
proxy-sp-feedback-info

When no ids are connected to a proxied SP, previously the SP info of the proxy was displayed on the feedback info. This change will ensure both the SP and proxy SP are represented on the feedback info page.

@thijskh Two questions: 
1) Do you agree with the proposed labels in the feedback info table? __Proxy SP__ and _Proxy SP Name_ ? Shorter abbreviations might be more economical.

2) The columns might need reorganization to get optimal screen realestate for the proxy sp name. For now I'd like to show you this for verification.

See: https://www.pivotaltracker.com/story/show/174036058

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
+++ b/library/EngineBlock/Corto/Module/Service/SingleSignOn.php
@@ -81,6 +81,7 @@ class EngineBlock_Corto_Module_Service_SingleSignOn implements EngineBlock_Corto
         // When dealing with an SP that acts as a trusted proxy, we should perform SSO on the proxying SP and not the
         // proxy itself.
         if ($sp->getCoins()->isTrustedProxy()) {
+            $proxySp = $sp;
             // Overwrite the trusted proxy SP instance with that of the SP that uses the trusted proxy.
             $sp = $this->_server->findOriginalServiceProvider($request, $log);
         }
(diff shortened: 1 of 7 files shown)


thijskh:
I propose to drop “Proxy SP name” since the proxy is relevant for internal use and we know the names of proxies

thijskh:
I propose to drop “Proxy SP name” since the proxy is relevant for internal use and we know the names of proxies

MKodde:
The Proxy SP name was dropped from the feedback info, and I've spreaded the proxy entity id over 2 columns.

library/EngineBlock/Corto/Module/Service/SingleSignOn.php
```diff
@@ -167,6 +168,12 @@ public function serve($serviceName, Request $httpRequest)
 
         // 0 IdPs found! Throw an exception.
         if (count($candidateIDPs) === 0) {
+            // When a trusted proxy is used, the currentServiceProvider is set to the entityId of the original issuing
+            // SP. This prevents the display of the Proxy in the feedback information.
+            if (isset($proxySp) && $proxySp->getCoins()->isTrustedProxy()) {
```
johanib, line 173: Maybe instanceof check instead of isset?

MKodde: Personally I find this statement more expressive and should not be less performant as far as I know.

MKodde: > Code looks good. But build seems to fail.

Thanks again for the review! The failing build seems to be a inconsistency on the PHP 7.4 build that pops up every now and then. I'm triggering a new build and will put a maintenance ticket on our backlog to investigate/fix this issue.

PR #877, opened 2020-08-11
Remove non conforming SHO oid from config

MKodde:
~The urn:oid:1.3.6.1.4.1.1466.115.121.1.15 attribute was previously released by default when the SHO was in the ARP. This oid is not valid and was added for historic reasons. It's not clear which SP's still rely on this attribute. So by default the attribute is not released by EB, but can be enabled using the 'eb.arp_remove_non_conforming_sho_attribute' feature flag.~

Simply removing the alias from the config was a more pragmatic solution. Federations relying on the alias can manually put it back if need be.

See: https://www.pivotaltracker.com/story/show/164237578

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,21 @@
 # UPGRADE NOTES
 
+## 6.2 > 6.3
+# Removal of incorrect schacHomeOrganization ARP alias
+
+If your federation relies on the release of the urn:oid:1.3.6.1.4.1.1466.115.121.1.15 sHO alias in the ARP. Then please
+put the entry back manually. This config can be found in `application/configs/attributes.json` near line 330.
+
+The following alias can be added back (or review the git history):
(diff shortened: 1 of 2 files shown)


MKodde:
Yes that should work too, the only drawback is that you'd have to keep track of an alternative configs/attributes.json file in you deploy scripts. If that works for you, I'd love to revert this change!

PS the build 72 build break on a visual regression test that times out. It ran on the 74 test so this should not be a blocker.

thijskh:
We already ship our own version of the attributes.json config file in https://github.com/OpenConext/OpenConext-deploy/blob/master/roles/engineblock/files/attributes.json

At least it's then out of the EB product. Of course everyone using OpenConext-deploy will still have it. But we can consider to move the SURF-specific items in attributes.json to our ansible environments, we can do that separately.

PR #878, opened 2020-08-11
Move metadata organization business rules away from metadata assembler

MKodde:
1. Organization data is now always assembled in the PushMetadataAssembler
2. Still only displayed on the metadata field when all data is complete.
3. An additional bugfix was applied, where the displayname was never rendered

https://www.pivotaltracker.com/story/show/170744290

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
@@ -124,7 +124,7 @@ class IdentityProviderEntity implements IdentityProviderEntityInterface
         return $this->entity->logo;
     }
 
-    public function hasValidOrganizationData(string $locale): bool
+    public function hasCompleteOrganizationData(string $locale): bool
     {
         $organization = $this->entity->getOrganization($locale);
         if ($organization
--- a/src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
+++ b/src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
@@ -124,6 +124,19 @@ class IdentityProviderEntity implements IdentityProviderEntityInterface
         return $this->entity->logo;
     }
 
+    public function hasValidOrganizationData(string $locale): bool
+    {
+        $organization = $this->entity->getOrganization($locale);
+        if ($organization
+            && is_string($organization->displayName) && $organization->displayName !== ''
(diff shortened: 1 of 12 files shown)


src/OpenConext/EngineBlock/Metadata/Factory/Adapter/IdentityProviderEntity.php
```diff
@@ -124,6 +124,19 @@ public function getLogo(): ?Logo
         return $this->entity->logo;
     }
 
+    public function hasValidOrganizationData(string $locale): bool
```
thijskh, line 127: I think I prefer "C0omplete" over "Valid". The individual items may be valid and we will just use them if they are. _Only for metadata_ we need the set to be complete.

I would also definitely comment very clearly what this is about. That it's only used and needed for Metadata XML conformance.

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/Compat/Container.php
```diff
@@ -177,4 +177,14 @@ public static function formatXml($xml)
         $xml = $dom->saveXml();
         return $xml;
     }
+
+    public function getTempDir(): string
+    {
+        throw new NotImplementedException(sprintf('getTempDir is not implemented! URL'));
```
thijskh, line 183: Is this correct sprintf usage?

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/Compat/Container.php
```diff
@@ -177,4 +177,14 @@ public static function formatXml($xml)
...
+    }
+
+    public function writeFile(string $filename, string $data, int $mode = null): void
+    {
+        throw new NotImplementedException(sprintf('writeFile is not implemented! URL'));
```
thijskh, line 188: As above

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/Compat/Container.php
MKodde: No, it was not. 

Also I dropped the commit where I introduced this change. I incorrectly added this change on this PR where is should be part of an upcoming SAML2 library upgrade.

src/OpenConext/EngineBlockFunctionalTestingBundle/Saml2/Compat/Container.php
MKodde: Dropped this commit

PR #879, opened 2020-08-12
Add trusted proxy signing verification

MKodde:
When the trusted proxy does not sing its requests, we error out of the single sing on flow. The error is generic and has a non descriptive error message is raised.

See https://www.pivotaltracker.com/story/show/153995599

### DIFF
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -193,10 +193,22 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
             );
         }
 
+        $requestsMustBeSigned = $serviceProvider->requestsMustBeSigned;
+        $isTrustedProxy = $serviceProvider->getCoins()->isTrustedProxy();
+
+        if ($isTrustedProxy && !$requestsMustBeSigned){
+            throw new EngineBlock_Corto_Module_Bindings_Exception(
+                sprintf(
+                    'SP "%s" is marked as Trusted Proxy but does not have "redirect:sign" enabled.',
+                    $spEntityId
+                )
+            );
+        }
+
         // Determine if we should check the signature of the message
         $wantRequestsSigned = (
             // If the destination wants the AuthnRequests signed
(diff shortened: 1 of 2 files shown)


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -193,10 +193,19 @@ public function receiveRequest()
             );
         }
 
+        $requestsMustBeSigned = $serviceProvider->requestsMustBeSigned;
+        $isTrustedProxy = $serviceProvider->getCoins()->isTrustedProxy();
+
+        if ($isTrustedProxy && !$requestsMustBeSigned){
+            throw new EngineBlock_Corto_Module_Bindings_Exception(
+                'Trusted Proxies are required to sign their requests'
```
thijskh, line 201: Does this exception come with all relevant metadata in the logs?
Or should we already need to make the message something like "SP '%s' is marked as Trusted Proxy but does not have 'requests.sign' enabled."

MKodde: The relevant metadata would probably be present (scattered) in the logs. I like your suggestion to make this message more verbose.

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -193,10 +193,22 @@ public function receiveRequest()
             );
         }
 
+        $requestsMustBeSigned = $serviceProvider->requestsMustBeSigned;
+        $isTrustedProxy = $serviceProvider->getCoins()->isTrustedProxy();
+
+        if ($isTrustedProxy && !$requestsMustBeSigned){
+            throw new EngineBlock_Corto_Module_Bindings_Exception(
+                sprintf(
+                    'SP "%s" is marked as Trusted Proxy but does not have "requests.sign" enabled.',
```
thijskh, line 202: Sorry, it’s called “redirect.sign” I think

MKodde: Yup, you are absolutely right. The assembler reads the: metadata.redirect:sign. Updated the exception message accordingly.

PR #883, opened 2020-08-18
Add visual regression snapshots

MKodde:
And improve js_testing.md while at it. Amending a chrome + chrome driver installation instruction.

See for details: https://www.pivotaltracker.com/story/show/174377003


pablothedude:
Nice catch! Maybe this would opt even more for the usage of Cypress.

A bit off topic but I was wondering if the additional feedback info was also added also in the tests, I was missing them in the PR below, but maybe updated now while at it.
https://github.com/OpenConext/OpenConext-engineblock/pull/875

MKodde:
The new plan is to move the javascript tests to the Cypress test framework. With that, this PR will be closed.

PR #884, opened 2020-08-24
Try Cypress for Consent tests

MKodde:
- ~Note that Cypress shouldb e dev dependency (todo)~
- npm install should yield a working Cypress installation. Running
./theme/node_modules/.bin/cypress open gives you the GUI test runner.
- Only a simnple consent test was added.
- Todo: implement a visual regression test setup possibly using Percy.io


pablothedude:
Nice work, I think the use of Cypress in favour of Puppeteer will be beneficial in the long run because it has better E2E support and that's were we currently are using Jest/Puppeteer for.

MKodde:
I'm merging this, and will open a new PR that is aimed at running the Cypress tests on CI and porting the existing tests

Issue #39, opened 2014-05-21
Add support for IsPassive="true"


pmeulen:
The current implementation of IsPassive consists of passing this value thru to the home IdP (similar to ForceAuthn). 
From os-core section 3.4.1: "If [IsPassive is] 'true', the identity provider and the user agent itself MUST NOT visibly take control of the user interface from the requester and interact with the presenter in a noticeable fashion"

The implementation if IsPassive in OpenConext is not complete:
- OpenConext can show a WAYF if the user has no session, this should not happen. Instead;
- when passive is not possible OpenConext should respond with a SAML Status, since we cannot show an error to a user:
  samlp:Status
      <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Responder">
          <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:NoPassive" />
      /samlp:StatusCode
  /samlp:Status

Also in case of other errors we should pass a status code to the IdP.

Because OpenConext passes IsPassive to the home IdP (as it should), OpenConext should expect and be able to handle a status code from the IdP.

thijskh:
I researched this issue and these are the conclusions:

When receiving an IsPassive authnrequest, EB:
- sends a correct NoPassive response when >1 IdP is connected to this SP;
- sends a correct IsPassive authnrequest when 1 IdP is connected to this SP;
- processes an assertion correctly when this IdP decides to return one;
- throws an exception and shows the error screen "Identity Provider error" when the IdP returns a NoPassive response.

Only the last bullet is incorrect behaviour. The correct behaviour is that if EB receives a NoPassive from the IdP, it should similarly send out a NoPassive message to the SP.

PR #885, opened 2020-08-24
Send NoPassive status response back to issuing SP

MKodde:
These NoPassive status responses are sent by the IdP who is not
able/allowed to process a Passive authentication.

Previously we would show a custom error page. But on this occasion it's
better to let the SP decide how to proceed in this situation.

See: https://www.pivotaltracker.com/story/show/164238276
Closes: https://github.com/OpenConext/OpenConext-engineblock/issues/39

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
+++ b/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
@@ -25,6 +25,7 @@ use OpenConext\EngineBlockBundle\Authentication\AuthenticationState;
 use OpenConext\Value\Saml\Entity;
 use OpenConext\Value\Saml\EntityId;
 use OpenConext\Value\Saml\EntityType;
+use SAML2\Constants;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Session\Session;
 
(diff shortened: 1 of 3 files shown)


library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -89,6 +90,14 @@ public function serve($serviceName, Request $httpRequest)
         $application = EngineBlock_ApplicationSingleton::getInstance();
         $log = $application->getLogInstance();
 
+        // Test if we should return a no passive status response back to the SP
```
pablothedude, line 93: Doesn't this have to take place after the response signature method, requester chain validation and the key check?

pablothedude (changes requested):
Maybe the response should happen later on in the process, could you verify this?

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
MKodde: I'm not sure that would be wise if at all possible. My rationale is to exit as soon as possible. But maybe @thijskh, @phavekes  or @pmeulen can answer your question?

pablothedude: @MKodde it seems possible according to our tests if I move the logic down (after those checks).

MKodde: Yes, it might work. But given the situation in the Bindings.php::receiveResponse method where I return the ResponseAnnotationDecorator from the catch block. I can't be sure what other problems are present in that response. It might have failed on any precondition. Processing the SAML Response after preconditions failed, in the AssertionConsumer might result in unexpected behaviour that might not be simulated in the Behat tests.

thijskh: I prefer that we take action only after having performed the checks "is this even a legitimate message". Doing it early might enable a (yet unkown) security issue and we might miss useful things like 'additional logging' that are futher below in the code, on this type of response.

pablothedude (approved):
:+1: thanks!

PR #887, opened 2020-08-26
Migrate existing tests to cypress

MKodde:
The goal of this PR is to integrate all JavaScript tests to Cypress test framework.

This includes the:
- Unit tests
- E2E tests (Wayf and Consent)
- Visual Regression tests (error pages)

Two test stages are added to the GitHub Actions test-integration
 - Unit and E2E tests run on every build (PR and Commit)
 - Visual regression tests only run when a tag is pushed or when a nightly build is triggered


.github/workflows/test-integration.yml
```diff
@@ -128,6 +128,14 @@ jobs:
...
               env:
                   SYMFONY_ENV: ci
+            - name: Run Cypress visual regression tests
+              if: ${{ github.event_name == 'schedule' && matrix.php == env.PROD_PHP }}
+              run: |
+                  cd docker && docker-compose exec -T cypress bash -c '
+                      CYPRESS_integrationFolder=cypress/visual-regression cypress run --browser=chrome --headless
+                  '
+              env:
+                  SYMFONY_ENV: ci
```
MKodde: Isn't this block a duplication of lines 97 through 103?

MKodde (commented):
Changes look good to me

1. The unit and e22 tests do not yet run correctly on CI, have to look why they started failing
2. I'd have to look into a way to run the unit & e2e tests on every builld, but run the visual regression tests only on releases. Should be easy enough to do.

PR #889, opened 2020-09-03
Store entityId of issuer, not the value object

MKodde:
The Issuer object was stored in the CachedResponses session. Causing problems with successive use of the
CachedResponses data which is expected to be a string representation of the SP/IdP entityID.

https://www.pivotaltracker.com/story/show/174627328

### DIFF
--- a/library/EngineBlock/ApplicationSingleton.php
+++ b/library/EngineBlock/ApplicationSingleton.php
@@ -251,9 +251,6 @@ class EngineBlock_ApplicationSingleton
         if (isset($_SESSION['currentServiceProvider'])) {
             $feedbackInfo['serviceProvider'] = $_SESSION['currentServiceProvider'];
             $spEntityId = $_SESSION['currentServiceProvider'];
-            if ($spEntityId instanceof Issuer) {
-                $spEntityId = $spEntityId->getValue();
-            }
             $feedbackInfo['serviceProviderName'] = $this->getServiceProviderName($spEntityId);
         }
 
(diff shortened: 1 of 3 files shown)


pablothedude (changes requested):
That seems to do the job. When grepping I was able to find a couple of other suspicious lines of code.

https://github.com/OpenConext/OpenConext-engineblock/blob/e6ff14e0a19c03347567d750b57f6182a3114da3/library/EngineBlock/Corto/Module/Bindings.php#L156
https://github.com/OpenConext/OpenConext-engineblock/blob/e6ff14e0a19c03347567d750b57f6182a3114da3/library/EngineBlock/Corto/Module/Bindings.php#L171
https://github.com/OpenConext/OpenConext-engineblock/blob/e6ff14e0a19c03347567d750b57f6182a3114da3/library/EngineBlock/Corto/Module/Bindings.php#L183

Could you verify this?

tvdijen:
All three definitely need a `->getValue()` appended

MKodde:
Thanks for finding these @pablothedude. I updated the code. Luckily the code did not break by not adding the ->getValue() method. See my commit message for more details.

pablothedude:
For me it's still somewhat confusing that in `Bindings.php` `$spEntityId` holds an Issuer VO and not the result of `->getValue()` itself. I'ts maybe a bit out of scope for now but I think changing that will make the code better readable in terms of ubiquitous language.

pablothedude (approved):

MKodde:
Rationale was to leave the value object as is in that method, and call the getValue method when required. Granted the other methods are not used in the scope of the method. 

I will leave it as is for now.

PR #891, opened 2020-09-07
Print the key-id in the SSO locations of the IdP metadata

MKodde:
The rollover key was not present in the IdP metadata of EngineBlock. This is taken care of in this commit.

An additional challenge arose: The Key Id is not always explicitly defined when opening the IdP metadata, in fact it is only used in roll-over scenarios. Setting the default by default caused the SSO URl to be filled with key:default. Where this is not a requirement.

https://www.pivotaltracker.com/story/show/174672018

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
+++ b/src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
@@ -41,8 +41,11 @@ class KeyPairFactory
      *
      * @throws RuntimeException
      */
-    public function buildFromIdentifier(string $identifier) : X509KeyPair
+    public function buildFromIdentifier(?string $identifier) : X509KeyPair
     {
+        if ($identifier === null) {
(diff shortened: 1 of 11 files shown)


src/OpenConext/EngineBlock/Metadata/X509/KeyPairFactory.php
```diff
@@ -41,8 +41,11 @@ public function __construct(array $keyPairConfiguration)
      *
      * @throws RuntimeException
      */
-    public function buildFromIdentifier(string $identifier) : X509KeyPair
+    public function buildFromIdentifier(?string $identifier) : X509KeyPair
     {
+        if ($identifier === null) {
+            $identifier = self::DEFAULT_KEY_PAIR_IDENTIFIER;
+        }
```
MKodde, line 48: Instead of doing this in the controller, what was previously the case. We now set the default identifier in the KeyPairFactory. This prevents other users of the Key from using the default key-id

thijskh (approved):

PR #893, opened 2020-09-11
Whether MFA AuthnContext must be added should be based on original SP

thijskh:
So this also works when the SP in question is behind a trusted proxy.

### DIFF
--- a/library/EngineBlock/Corto/ProxyServer.php
+++ b/library/EngineBlock/Corto/ProxyServer.php
@@ -417,7 +417,7 @@ class EngineBlock_Corto_ProxyServer
         }
 
         // Add authncontextclassref if configured
-        $service = $identityProvider->getCoins()->mfaEntities()->findByEntityId($serviceProvider->getEntityId()->getEntityId());
+        $service = $identityProvider->getCoins()->mfaEntities()->findByEntityId($this->findOriginalServiceProvider($spRequest, $this->_logger)->entityId);
         if ($service instanceof MfaEntity) {
             $sspMessage->setRequestedAuthnContext([
                 'AuthnContextClassRef' => [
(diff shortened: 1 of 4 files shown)


thijskh:
Unit tests fail, something with missing ProxyServer, help wanted?

library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -416,8 +417,15 @@ public function sendAuthenticationRequest(
             throw new EngineBlock_Corto_ProxyServer_Exception(sprintf('Unknown message type: "%s"', get_class($sspMessage)));
         }
 
+        try {
+            $originalSpEnitytId = $this->findOriginalServiceProvider($spRequest, $this->_logger)->entityId;
+        } catch (EntityNotFoundException $e) {
+            // On debug requests, the entity ID can not be found in the database as this is the EngineBlock internal
+            // entity which does not reside in the database.
+            $originalSpEnitytId = $spEntityId;
+        }
```
MKodde, line 426: @thijskh I did this in order to allow debug authentications to succeed. The repository that would try to find the SP in the database throws an EntityNotFoundException as the internal entity (EngineBlock SP) is not retrievable from the database. As we no longer import it during the metadata assembly.

I personally don't see any issue with my change, it kind of falls back on the previous behavior which should yield the correct result. As we can safely say there is no proxy at work here. At least not a trusted proxy. Do you agree with this?

thijskh: Yes

thijskh:
:+1:

MKodde (approved):

MKodde:
Merging this PR
Releasing v6.3.4

PR #896, opened 2020-09-17
Improve consent test endpoint

MKodde:
The low hanging fruit for overriding the default consent functional testing controller has been updated here.

What remains to do is setting the attributes and more importantly the attribute sources.

The new query parameters available on this endpoint can be found in the js-testing readme file.

NameID overriding is a somewhat strange override. When you use the `name-id` query param, the name id is set to be `unspecified`. Even if you use the `persistent-name-id` feature flag.


Badlapje:
los van een blanco lijn die PHPCS vond ziet het er mij allemaal dik ok uit :-)

Badlapje (approved):
dik ok :-)

MKodde:
Allright, I'll fix the breaking build and will merge this feature

PR #897, opened 2020-09-22
Move the material theme to the base theme folder

MKodde:
## About the setup
The base theme will be the simplest version of the theme. It can not be rendered by itself. But an empty extension shows how basic the styling is.

The `openconext` theme is the old css/js that resided in the `theme/material` folder. It comes WITH the `material/openconext` theme. Which only differs from the material theme by adding the surfboard to the error page (background.css)

The `skeune` theme extends the base theme. Adding the color overrides. Note that the wayf, index page and the error page need to be styled in this theme. Or alternatively,  an effort should be put into leveraging the old index and error designs into the base theme.

## Building
Building stayed unchanged for now. The temporal `-theme` buid targets have been removed. Building, using the theme prefix should work for both themes. When omitting the `EB_THEME` env var, the `openconext` theme is built by default.

In order to configure Twig. Please set the `theme.name` parameter in `parameters.yml` according to the theme you'd like to build.

Examples:

```
$ EB_THEME=skeune npm run build
$ EB_THEME=skeune npm run build:css
$ npm run build
```

## Todo
- Update OpenConext-deploy tasks
- Decide how we want to go about theme specific tests
- The readme regarding theming must be rewritten
- Test if the watch tasks still work


theme/base/stylesheets/vars.overrides.scss
```diff
@@ -5,41 +5,41 @@
...
-$brownGray: #ababab;
-$hardBorderGray: #919ba4;
-$brownishGray: #5b5b5b;
-$skeuneGray: #f4f6f8;
-$white: #fff;
+$black: #212121 !default;
+$hoverBlue: #ddf2fd !default;
+$loginBlue: #b3e5ff !default;
+$lightBlue: #94d6ff !default;
+$skeuneLightBlue: #7cc8e1 !default;
```
MKodde, line 12: Renaming this might be a good idea, we don't want any skeune references in the base theme right?

Badlapje: agreed, will fix that and the filename in one go :-)

app/config/parameters.yml.dist
```diff
@@ -235,4 +235,7 @@ parameters:
 ##########################################################################################
     ## THEME SETTINGS
     ##########################################################################################
-    theme.name: 'openconext'
+    # When changing the default theme name, be aware that the CI build will no longer be able to select our default
```
Badlapje, line 238: tof dit 👍

MKodde: A warning for a construction that needs definite refinement!

MKodde:
**Sitrep spt 28st**
I attempted to create a theme switch in the e2e tests. Basically updating the parameters.yml, clearing the cache, and rebuilding the theme. This resulted in very hard to debug issues on the GH actions pipeline. So I've postponed that feature in favour of only running the 'skeune' e2e tests and skipping all material tests for now.

Note that the a11y tests are failing for now on some a11y and html errors. I don't think this should hold back a merge at this point.

Badlapje:
> The base theme will be the simplest version of the theme. It can not be rendered by itself. But an empty extension shows how basic the styling is.

It can be rendered by itself, no?  We just advise against it.

theme/base/javascripts/wayf.js
```diff
@@ -11,19 +9,9 @@ import {IdpListElementFactory} from "./modules/IdpListElementFactory";
...
-        return;
-    }
-
-    if (document.getElementById('engine-main-page') !== null) {
-        replaceMetadataCertificateLinkTexts();
-        return;
-    }
+export function initializeWayf() {
 
+    document.body.className = document.body.className.replace('no-js', '');
```
Badlapje, line 14: deze is iets om in de openconext-specifieke js te steken als we de JS refactoren.  In skeune hoeft dit niet meer.  Ik heb'm daar toegevoegd.

Badlapje:
Looks good to me, aside from the failing check.  I'll run that one locally to see what's up.

MKodde:
Merging :+1:

PR #898, opened 2020-09-22
Clean up unused usage of authnrequest destination

thijskh:
This is actually a bugfix. Sending an AuthnRequest to EB >= 6.3.1 that has no `Destination` parameter (which is indeed optional in the spec):

```xml
<samlp:AuthnRequest xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
                    xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
                    ID="CORTO7b58fe48a7c62ea17e02b50a68b0dfa021d69280"
                    Version="2.0"
                    IssueInstant="2020-09-22T14:43:47Z"
                    >
    <saml:Issuer>http://example.org</saml:Issuer>
</samlp:AuthnRequest>
```

Would yield a fatal error:
```
Argument 2 passed to EngineBlock_Corto_Module_Bindings::_verifyKnownSP() must be of the type string, null given, called in /opt/openconext/OpenConext-engineblock-6.3.4/library/EngineBlock/Corto/Module/Bindings.php on line 188
```

Type `string` was added to the `$destination` parameter of `_verifyKnownSP()` in 6.3.1. Simple solution is to also allow null, but closer inspection shows that the destination is only passed down to a possible exception but there also not used at all (anymore?). So the even better solution seems to just clean it all up.

### DIFF
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -183,9 +183,7 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
         $_SESSION['currentServiceProvider'] = $ebRequest->getIssuer()->getValue();
 
         // Verify that we know this SP and have metadata for it.
-        $serviceProvider = $this->_verifyKnownSP(
-            $spEntityId->getValue()
-        );
+        $serviceProvider = $this->_verifyKnownSP($spEntityId->getValue());
 
         if (!$serviceProvider instanceof ServiceProvider) {
             throw new EngineBlock_Corto_Module_Bindings_Exception(
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -184,8 +184,7 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
 
         // Verify that we know this SP and have metadata for it.
         $serviceProvider = $this->_verifyKnownSP(
-            $spEntityId->getValue(),
-            $ebRequest->getDestination()
+            $spEntityId->getValue()
         );
 
         if (!$serviceProvider instanceof ServiceProvider) {
@@ -513,11 +512,10 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
      * Verify if a message has an issuer that is known as an SP to us. If not, it
      * throws a Corto_Module_Bindings_VerificationException.
      * @param string $messageIssuer
-     * @param string $destination
      * @return AbstractRole Remote Entity that issued the message
      * @throws EngineBlock_Exception_UnknownServiceProvider
      */
-    protected function _verifyKnownSP(string $messageIssuer, string $destination = '')
+    protected function _verifyKnownSP(string $messageIssuer)
     {
         $remoteEntity = $this->_server->getRepository()->findServiceProviderByEntityId($messageIssuer);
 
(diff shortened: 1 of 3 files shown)


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -184,8 +184,7 @@ public function receiveRequest()
 
         // Verify that we know this SP and have metadata for it.
         $serviceProvider = $this->_verifyKnownSP(
-            $spEntityId->getValue(),
-            $ebRequest->getDestination()
+            $spEntityId->getValue()
         );
```
MKodde, line 188: Nitpick
Consider formatting it back to a single line statement, now that the second argument is no longer present.

src/OpenConext/EngineBlockBundle/EventListener/RedirectToFeedbackPageExceptionListener.php
```diff
@@ -160,7 +160,6 @@ public function onKernelException(GetResponseForExceptionEvent $event)
 
             $redirectParams  = [
                 'entity-id'   => $exception->getEntityId(),
-                'destination' => $exception->getDestination()
             ];
```
MKodde, line 163: Two comments here:
1. The controller action that is requested for this exception still reads the destination from the request. See: `https://github.com/OpenConext/OpenConext-engineblock/blob/a79112cbaee1cc7557ae6bccfe226e0ffda1484a/src/OpenConext/EngineBlockBundle/Controller/FeedbackController.php#L193` That should be cleaned up too
2. Again, this can be converted back into a one-liner statement now that the destination is removed. But feel free to leave as is.

MKodde (changes requested):
Thanks for fixing this @thijskh, see two minor and one more pressing review comment below. Have you thought of covering this in an acceptance (behat) test? Maybe this is not quite worth the effort, but if you are interested I could assist you in this.

MKodde:
> Type `string` was added to the `$destination` parameter of `_verifyKnownSP()` in 6.3.1. Simple solution is to also allow null, but closer inspection shows that the destination is only passed down to a possible exception but there also not used at all (anymore?). So the even better solution seems to just clean it all up.

The destination was used to display in the feedback info block of the feedback page. But removing it is fine with me personally.

thijskh:
> The destination was used to display in the feedback info block of the feedback page. But removing it is fine with me personally.

"Was used" as in something the past? Because I believe it's not used currently/recently.

src/OpenConext/EngineBlockBundle/EventListener/RedirectToFeedbackPageExceptionListener.php
thijskh: 1. This is the unknown identity provider action. Not the unknown service provider action.

We do more processing on the Destination when receiving a message from an IdP (Assertion). And there's no problem there currently. So I left that code to be. In an Assertion/Response (from remote IdP) it's mandatory. In an AuthnRequest (from remote SP) it's not.

thijskh: 2. This matches other lines that have one parameter in the file. So I'll leave it as is.

MKodde: 1. Right you are, my bad.
2. I'm fine with that.

MKodde:
> "Was used" as in something the past? Because I believe it's not used currently/recently.

No, I was mistaken. I was looking at the IdP not found action as you mentioned. 

LGTM, lets ship it!

MKodde (approved):

PR #899, opened 2020-09-22
Add support for mobile devices on consent page

Badlapje:


theme/skeune/stylesheets/helpers.scss
```diff
@@ -12,8 +12,9 @@
  ******************************************************************/
 
 @import '~@csstools/normalize.css';
+//@import 'a11y-tests';
```
MKodde, line 15: If this commented out import turns out to be really unused, then please remove this line altogether

theme/skeune/stylesheets/breakpoints.scss
```diff
@@ -0,0 +1,64 @@
...
+$laptopAndTablet: '(min-width: 768px) and (max-width: 1339px)';
+$tabletAndSmaller: '(max-width: 1023px)';
+$laptopAndSmaller: '(max-width: 1339px)';
+
+@mixin screen($size) {
+    @if $size == desktop {
+        @media only screen and #{$desktop} {
+            @content;
+        }
+    } @else if $size == laptop {
```
MKodde, line 19: Nice mixin! I love the use of the @content `sass` feature here!

theme/skeune/templates/modules/Authentication/View/Proxy/Partials/Consent/idpRow.html.twig
```diff
@@ -7,12 +7,16 @@
     {% set id %}
         incorrect_{{ attributeSource }}
     {% endset %}
-    {% include '@theme/Default/Partials/toggleCheckbox.html.twig' with {
+    <input type="checkbox" tabindex="-1" role="button" aria-pressed="false" class="modal visually-hidden" id="{{ id|trim }}" name="{{ id|trim }}" />
+    {# note: the div is here only for IE11, as soon as IE11 support is scrapped, this can be removed along with the CSS/JS for it  #}
```
MKodde, line 11: :muscle:

languages/messages.en.php
```diff
@@ -163,6 +163,14 @@
...
      'consent_disclaimer_secure_onemodal' => <<<'TXT'
 These data will be sent securely from your institution via <input type="checkbox" tabindex="-1" role="button" aria-pressed="false" class="modal visually-hidden" id="consent_disclaimer_about" name="consent_disclaimer_about"><label class="modal" tabindex="0" for="consent_disclaimer_about">%suite_name%</label> %modal_about% using a number that uniquely identifies you for this service.
+TXT
+     ,
+     'consent_disclaimer_secure_onemodal' => <<<'TXT'
+These data will be sent securely from your institution via <input type="checkbox" class="modal" id="consent_disclaimer_about"><label class="modal" for="consent_disclaimer_about">%suite_name%</label> %modal_about% using a number that uniquely identifies you for this service.
+TXT
+     ,
+     'consent_disclaimer_secure_onemodal' => <<<'TXT'
+These data will be sent securely from your institution via <input type="checkbox" class="modal" id="consent_disclaimer_about"><label class="modal" for="consent_disclaimer_about">%suite_name%</label> %modal_about% using a number that uniquely identifies you for this service.
```
MKodde, line 173: Overlooked this in one of my previous reviews. But is this the nicest way to add the modal trigger? Maybe the HTML should be offloaded to the template? Granted, other (more style oriented) html elements are in the translation files. Maybe we should look into a way to remove those too?

Whats your take on this?

MKodde (commented):
I see good things here! Some smaller feedback points can be found below.

theme/skeune/stylesheets/helpers.scss
Badlapje: done and clarified it in the readme

languages/messages.en.php
Badlapje: I had the same objection, but opted to put it in here in the end because splitting the string into two parts works only for English.  If we translate it to Dutch or to Portuguese (or other languages later), the input might need to be in a different part.  So it's intrinsically tied to the translation, sadly.

Some of the other html-elements we can move indeed.  I'll add it to the leftover ticket.

MKodde: Thanks for the clarification. I'm fine with this! And thanks for adding the task.

MKodde (approved):

MKodde:
:shipit:

PR #903, opened 2020-09-29
Break up Cypress tests for different themes

MKodde:
## Cypress theme segregation
The goal was to be able to run Cypress e2e tests against the different themes.  The challenge is that the theme must be built and the Twig theme engine must be initialized to use the correct templates.

In the end, creating integration folders in every theme was not implemented as it goes against the Cypress philosophy so I refrained from doing that. The main `theme/cypress/integrations` folder was used instead. In that folder, every theme can install a set of tests.

In order to run those tests against the correct assets, a test script was created. This script:

1. Updates the parametes.yml `theme.name` value to the theme name that was passed to the script
2. Clears the CI application cache
3. Builds the theme
4. Runs the tests found in the `themename` folder of the `cypress/integration` folder

## Todo (in this PR?)
- [x] Update the readme to instruct the user on how to use the test.js file
- ~Think about other test uses for the test.js file. Maybe it can also run the visual regression tests?~
- ~Investigate if the cli output can be outputted directly from the `child process` or maybe run the script in the 'parent' process.~


theme/README.md
```diff
@@ -66,6 +66,17 @@ if however you prefer to use the openconext theme, it'd read like this:
...
+**Switching themes on CI**
+A helper script was included in the `theme/scripts` folder to assist in building theme assets.
+
+This script changes the Twig theme and builds the chosen frontend theme assets. To use this script you simply run:
+
+```bash
+$ EB_THEME=skeune ./scripts/prepare-test.js
+```
+
+The script must be run on the php-fpm instance on your CI environment as it also clears the application cache in order to let the correct Twig templates to be included in the cache.
```
Badlapje, line 78: nitpicky, maar ik zou deze zin herschrijven naar:

The script must be run on the php-fpm instance on your CI environment as it also clears the application cache in order to include the correct Twig templates in the cache.

theme/cypress/integration/openconext/consent/consent.material.spec.js
```diff
@@ -4,7 +4,7 @@ context('Consent on Material theme', () => {
     cy.visit('https://engine.vm.openconext.org/functional-testing/consent');
```
Badlapje, line 4: lijkt me dat de .material (en bij de skeune test) de .skeune uit de filenaam mag?

Badlapje:
Nice work! 💯

PR #905, opened 2020-09-29
Move missing templates to base

MKodde:
The form and redirect templates, used to carry the AuthNRequest and the
SAML response through EngineBlock, where not yet in the base templates,
resulting in broken authentications.


MKodde:
@thijskh The broken authentication was caused by missing Twig templates that are not yet in the 'new' base template system.

This PR simply copies over the tried and tested templates. Maybe @Badlapje kan verify if these templates could be made better from a no-js accessability standpoint? But for now this seems a fix for the problem at hand.

MKodde:
Disclaimer: I noted during my tests that the debug authentication does not succeed, it fails on a `NotOnOrAfter in SubjectConfirmationData is in the past`

Reported that in [Pivotal](https://www.pivotaltracker.com/story/show/174968005)

Badlapje (approved):

Badlapje:
Looks good to me!

PR #908, opened 2020-09-30
Copy fonts to web folder in build script

Badlapje:
Fonts weren't working as expected because they were not being copied over to the web folder.

There are also two small bugfixes in here.  If you prefer them in a different branch, i'll take them out.  
1. remove margin on idp list (as the margin is already correctly implemented on the header of the preselection).
2. move the exit code to the end of the executecommand function, as else we can't see the output from the build script.


MKodde (commented):
Nice to see theme fonts take precedence over the base fonts!

Some questions:

1. I think this is an opportunity to add the web/fonts to the `.gitignore` file. Before doing so, you should remove the font files that are already committed to this folder? And maybe add a gitkeep folder to ensure it exists?
2. Are all fonts currently present in the base font folder required? I think not, as the ProximaNova font is replaced with the Nunito font IIRC. I'd love to see as little as possible font files in VCS.
3. Are we allowed to remove the NunitoOFL.txt file? At least maybe you could prevent it being copied to the web folder?

theme/scripts/build.js
```diff
@@ -26,7 +26,7 @@ try {
 
     console.log(`Using theme ${theme} to run the build.\nOutput will be printed once the build is finished.\n`);
     executeShellCommand(`cd ${__dirname}/.. && EB_THEME=${theme} npm run buildtheme`);
-    return process.exit(0);
+    // return process.exit(0);
```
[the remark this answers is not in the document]
Badlapje: done

Badlapje:
Superb feedback 💯   Sadly we can't remove the .txt file, but i excluded it from being copied.

.gitignore
```diff
@@ -7,6 +7,8 @@ bin/ignore_me.php
 /web/javascripts/*
 /web/stylesheets/*
 /web/images/*
+/web/fonts/*
+!/web/fonts/.gitkeep
```
MKodde, line 11: This certainly works, personally I use one commit to add the .gitkeep file, and then set the git ignore rule. That way the file is tracked, but new contents are ignored. But this is fine with me too!

MKodde (approved):
Nice work! Let's move forward with this!

.gitignore
Badlapje: i kinda botched it, because the file wasn't added automatically to git, so reckoned i'd fix it this way rather than doing some complicated rebase 😊

PR #909, opened 2020-09-30
Add automatic a11y test and fix problems

Badlapje:
Added a new command to get rid of the symfony debugbar during testing.

- There was a contrast issue with the edit button, because the font wasn't large enough.  Increased the font size to resolve.
- There was an issue with an extra open strong tag.  Removing it fixed that.
- The search form lacked a submit button.  Fixed that while keeping the styling the same.  Also added an aria-role as per best practice.
- The focus styles for the footer didn't show clearly as it was blue on blue.  Changed those to white.
- Buttons lacked a pointer when hovering above them, added those generically.
- The Idp lacked a type for it's button: added that.  It also has a button nested in a link.  So added a comment to disable event bubbling on the button to avoid accidentally triggering the link.

### DIFF
--- a/theme/cypress/support/commands.js
+++ b/theme/cypress/support/commands.js
@@ -22,3 +22,7 @@ Cypress.Commands.add('matchImageSnapshots', (viewport, pageDetails) => {
 Cypress.Commands.add('buildTheme', (themeName) => {
   cy.exec(`EB_THEME=${themeName} npm run buildtheme`);
 });
+
+Cypress.Commands.add('hideDebugBar', () => {
+  cy.get('.sf-toolbar').invoke('attr', 'style', 'display: none');
+});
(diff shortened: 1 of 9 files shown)


theme/cypress/support/commands.js
```diff
@@ -22,3 +22,7 @@ Cypress.Commands.add('matchImageSnapshots', (viewport, pageDetails) => {
 Cypress.Commands.add('buildTheme', (themeName) => {
   cy.exec(`EB_THEME=${themeName} npm run buildtheme`);
 });
+
+Cypress.Commands.add('hideDebugBar', () => {
+  cy.get('.sf-toolbar').invoke('attr', 'style', 'display: none');
+});
```
MKodde, line 28: :1st_place_medal:

MKodde (approved):
LGTM!

theme/cypress/support/commands.js
Badlapje: dit zit nog niet helemaal lekker: als ik het in de test zet dan lukt het lokaal, maar niet op de dev-omgeving. Dus je moet'm voor de a11y tests telkens toevoegen in de beforeEach en erna weer verwijderen.

PR #910, opened 2020-10-01
Add integration tests for the new WAYF

Badlapje:
Added preliminary integration tests for the WAYF screen in the new design.
Fixed them as far as i could for now (without interactivity) and commented on those which need to be fixed in the future.  Skipped those that can't work atm.

### DIFF
--- a/theme/base/stylesheets/pages/wayf.scss
+++ b/theme/base/stylesheets/pages/wayf.scss
@@ -140,7 +140,7 @@
                 @include grid-position(1, 2, 2, 3);
                 color: $buttonBlue;
                 cursor: pointer;
-                font-size: $f-normal;
+                font-size: $f-large;
                 font-weight: $bolder;
                 line-height: 1.5;
                 margin: auto 0 calculateRem(34px) calculateRem(5px);
@@ -244,12 +244,10 @@
             }
 
             > .wayf__search {
+                @include relative;
                 margin: 2rem 0 1.25rem;
 
                 > .search__field {
-                    @include poirot;
-                    background-position: right .5rem center;
-                    background-size: 1.5rem 1.5rem;
                     border: 1px solid $gray;
                     border-radius: 6px;
                     box-shadow: 0 1px 0 2px #5e6873;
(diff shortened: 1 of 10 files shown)


theme/cypress/integration/skeune/wayf/UnconnectedIdpPicker.spec.js
```diff
@@ -0,0 +1,7 @@
+import {UnconnectedIdpPicker} from "../../../../base/javascripts/modules/UnconnectedIdpPicker";
+
+it('Check if the UnconnectedIdpPicker constructor was called', () => {
+    const unconnectedIdpPicker = new UnconnectedIdpPicker(null, null, null);
+
+    expect(unconnectedIdpPicker).instanceof(UnconnectedIdpPicker);
+});
```
MKodde, line 7: This feels like code duplication to me. This code is already tested in the 'openconext' theme. That should be enough don't you think? 

What would be even better is to run the unit tests in another 'suite' which can be theme independent. Not something to fix here and now..

I suggest removing it from `skeune` for now

theme/cypress/integration/skeune/wayf/WayfMouseBehaviour.spec.js
```diff
@@ -0,0 +1,49 @@
+context('WayfMouseBehaviour', () => {
```
MKodde, line 1: I feel this test is also very `openconext` theme related. This test covers a bugfix that is specific for that theme. When we are changing behaviour of WAYF, this test probably becomes obsolete. 

I suggest removing it from `skeune`

MKodde (commented):
Other than the two requests for removal, this looks good to go!

theme/cypress/integration/skeune/wayf/WayfMouseBehaviour.spec.js
Badlapje: ah, makes sense to me if it's openconext specific :-)

theme/cypress/integration/skeune/wayf/UnconnectedIdpPicker.spec.js
Badlapje: Removed it for now, but added a task in the leftover ticket for this.

Even though it's already tested in the openconext theme, changes in the skeune theme might break it.  So it's better to test it for every theme.  It is code duplication though, so fully agree with that point.  We shall find a way ;-)

MKodde (approved):

PR #911, opened 2020-10-01
Style rememberchoice form

Badlapje:
After conferring with Arnout and Edwin, a spot was chosen for the rememberchoice feature.  This PR moves the form to that spot (between search and idp-list) & adds some styles for it.  The styles are a proposal.

### DIFF
--- a/theme/package.json
+++ b/theme/package.json
@@ -29,7 +29,6 @@
     "create-theme": "node scripts/create-theme.js",
     "watch:js": "watchify ${EB_THEME:=base}/javascripts/application.js --poll=100 --debug -t [ babelify ] ${EB_THEME:=base}/javascripts/application.js -o ../web/javascripts/application.min.js -v",
     "watch:css": "node-sass -r ${EB_THEME:=openconext}/stylesheets -w -o ../web/stylesheets",
-    "watch:css:single": "node-sass -r ${EB_THEME:=openconext}/stylesheets -o ../web/stylesheets",
     "release": "npm run build"
   },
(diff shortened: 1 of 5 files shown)


theme/package.json
```diff
@@ -29,6 +29,7 @@
     "create-theme": "node scripts/create-theme.js",
     "watch:js": "watchify ${EB_THEME:=base}/javascripts/application.js --poll=100 --debug -t [ babelify ] ${EB_THEME:=base}/javascripts/application.js -o ../web/javascripts/application.min.js -v",
     "watch:css": "node-sass -r ${EB_THEME:=openconext}/stylesheets -w -o ../web/stylesheets",
+    "watch:css:single": "node-sass -r ${EB_THEME:=openconext}/stylesheets -o ../web/stylesheets",
```
MKodde, line 32: I see that you like this! However, the name does not sit well with me. The behavior I see is a quick compilation of the scss/sass code to the web/stylesheets folder. It does not actually watch for changes. 

Maybe you where looking for: `npm run build:css:sass` ?

MKodde (commented):
Nice work again! See feedback below. The remember choice feature looked good and functioned as expected with the keyboard and mouse.

Would adding an e2e test for this be worth it?

theme/package.json
Badlapje: gah, can't believe i missed that.  Removed it again 🤦

MKodde: Not a problem!

MKodde (approved):

PR #912, opened 2020-10-01
Put temporary code in place to make WAYF function

MKodde:
This change adds support for the Skeune WAYF design, which is WIP to
begin with. So the change to the wayf.js must be considered temporary
code.

In order to continue to the IdP, an entity Id must be set on the form.
This form consists of hidden form fields (request id, entityId and so
forth) the form is then submitted.

The other code is only accessible when using the old theme.

### DIFF
--- a/theme/base/javascripts/wayf.js
+++ b/theme/base/javascripts/wayf.js
@@ -16,6 +16,20 @@ export function initializeWayf() {
         return;
     }
 
+  // Temporary code to get wayf to function. Only submit by clicking on a IdP functions for now
+  const form = document.querySelector('.wayf__search');
+  if (form !== null) {
+    // Set a quick and dirty click listener on every IdP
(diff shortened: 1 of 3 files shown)


theme/base/javascripts/wayf.js
```diff
@@ -16,6 +16,20 @@ export function initializeWayf() {
         return;
     }
 
+  // Temporary code to get wayf to function. Only submit by clicking on a IdP functions for now
+  const form = document.querySelector('.wayf__search');
+  if (form !== null) {
+    // Set a quick and dirty click listener on every IdP
+    document.querySelectorAll('.wayf__idp').forEach((article) => {
```
Badlapje, line 23: for the final code we can rewrite this to one click handler on the UL, to save resources.  A tad more complicated though, so for now this works great 👍

Badlapje (approved):

PR #913, opened 2020-10-01
Style the WAYF for mobile

Badlapje:
Turns out tuning the paddings and margins was enough :-)


MKodde:
Looked good on the viewports I tested in Chrome and Firefox. Was unable to use real mobile browsers in Browserstack.

MKodde (approved):

PR #915, opened 2020-10-01
Ensure the WAYF works cross-browser

Badlapje:
Fix various cross-browser issues

- label needs a container to be a grid item in IE11
- margin-top auto does not work in IE11
- add font-display swap to prevent content flashes or missing text while the font is loading
- on webkit browsers the input cannot get a height unless apearance none is added
- added a background size for the account button, so it's always completely visible
- increased size of 2nd column for previousselection section to avoid the focus style being messed up in chrome/ie11

Add normalize.css as scss

Removed it as a dependency because it caused an extra request & could not be found.
Instead added it inline as per our other css

NOTE: this PR was branched off from the mobile branch. I've put the base to that one so it's easier to review.  Before merging it needs to be changed to feature/ui-redesign

### DIFF
--- a/theme/base/stylesheets/helpers.scss
+++ b/theme/base/stylesheets/helpers.scss
@@ -11,7 +11,7 @@
     You also need to import these at the start of each file that uses a variable/mixin/general class.
  ******************************************************************/
 
-@import '~@csstools/normalize.css';
+@import 'normalize';
 @import 'breakpoints';
 @import 'classes';
 @import 'fonts';
--- a/theme/base/stylesheets/fonts.scss
+++ b/theme/base/stylesheets/fonts.scss
@@ -3,54 +3,61 @@
  ******************************************************************/
 
 @font-face {
+    font-display: swap;
     font-family: 'Nunito';
     font-style: normal;
     font-weight: 200;
     src: url('/fonts/nunito-extralight-webfont.woff2') format('woff2'),
-    url('/fonts/nunito-extralight-webfont.woff') format('woff'), url('/fonts/Nunito-ExtraLight.ttf') format('truetype');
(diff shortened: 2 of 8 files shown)


theme/base/stylesheets/helpers.scss
```diff
@@ -11,7 +11,8 @@
     You also need to import these at the start of each file that uses a variable/mixin/general class.
  ******************************************************************/
 
-@import '~@csstools/normalize.css';
+//@import '~@csstools/normalize.css';
```
MKodde: Perhaps remove the commented import?

#Nitpick

MKodde (changes requested):
Code wise I have little to argue with this! Looking good!

In my browser tests I found some stuff that looked slightly off. But not sure that should be part of this PR. My findings:

Firefox & Chrome, large viewport:
* Using keyboard navigation. The hover style of the search bar, the search button and the IdP selection buttons look off in regards to the rendering of the dotted lines.
* The 'i' image(svg?) in the eduID blue-bar moved to the baseline of the bar. It should be aligned in the heart of the text.
* I recon the same goes for the 'loupe' icon that is shown before the 'Nothing found' text at the bottom of the page.

Sadly BrowserStack did not allow me to view the page in IE11. So I hope you're able to verify this on your Windows machine?

It seemed a BS issue, so maybe that's fixed later today.

theme/base/stylesheets/helpers.scss
Badlapje: hehe, saw this when rebasing and removed it 💯

Badlapje:
Fixed the keyboard navigation issue (except for the IdP selection buttons as i wasnt sure what you meant (asked on slack), the i alignment and the loupe icon alignment.

Below is a screenshot of the page in IE.  Looks the same after the rework :-) 

image

Badlapje:
For the focus style of the idp itself, i'll address that when i refactor the markup to incorporate the form thought and get rid of the anchor.

PR #916, opened 2020-10-02
Refactor JS code for consent page

Badlapje:
Made a generic initializePage function.
Split up all js functions in seperate files for the consent page, put those in a consent subfolder.
Created a initializeConsent function for both openconext / base (skeune) which contains all relevant JS with updated selectors.
Removed the inline JS from default.html.twig.
Imported the JS for consent in application.js for both openconext / base (skeune).

### DIFF
--- a/theme/openconext/javascripts/consent.js
+++ b/theme/openconext/javascripts/consent.js
@@ -1,0 +1,17 @@
+import {focusAcceptButton} from '../../base/javascripts/consent/focusAcceptButton';
+import {initTooltips} from './consent/initTooltips';
+import {initAttributesToggle} from './consent/initAttributesToggle';
+import {initSlideIns} from './consent/initSlideIns';
+import {initializePage} from '../../base/javascripts/page';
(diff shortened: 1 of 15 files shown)


theme/openconext/javascripts/consent.js
```diff
@@ -0,0 +1,17 @@
...
+import {initializePage} from '../../base/javascripts/page';
+import {hideNoScript} from './hideNoScript';
+
+export const initializeConsent = () => {
+  const callbacksAfterLoad = () => {
+    focusAcceptButton('#accept_terms_button');
+    initTooltips();
+    initAttributesToggle();
+    initSlideIns();
+  };
```
MKodde, line 14: I like the modulized approach! Is your intention to also use some of these modules in the WAYF? That would be nice (if possible)!

MKodde (approved):

theme/openconext/javascripts/consent.js
Badlapje: It is indeed :-)

PR #919, opened 2020-10-03
Response without NameID is valid behaviour

tvdijen:
Should fix:
Oct  3 13:52:09 sv1810952 EBLOG[1527]: [2020-10-03 13:52:09] app.ERROR: Call to a member function getValue() on null {"session_id":"o6v872d3nh68rrlt1tn40jrv2g","request_id":"5f7865e9c3ebe"} {"exception":"[object] (Error(code: 0): Call to a member function getValue() on null at /apps/installation/OpenConext-engineblock/OpenConext-engineblock-6.3.5/library/EngineBlock/Corto/Filter/Abstract.php:75)"}

Although I haven't tested this yet..

### DIFF
--- a/library/EngineBlock/Corto/Filter/Abstract.php
+++ b/library/EngineBlock/Corto/Filter/Abstract.php
@@ -72,7 +72,7 @@ abstract class EngineBlock_Corto_Filter_Abstract
         else if ($response->getCollabPersonId()) {
             $collabPersonId = $response->getCollabPersonId();
         }
-        else if ($responseNameId->getValue()) {
+        else if (isset($responseNameId) && $responseNameId->getValue()) {
             $collabPersonId = $responseNameId->getValue();
         }
         else {


tvdijen:
I've just verified the suggested patch and it seems to do what it should..

MKodde:
Hi Tim, thanks for this PR and for already verifying that this should work! Would you be kind to target this PR on the `master` branch? We have not merged any post 6.3 work, so it's safe to keep working on master for these bugfixes

MKodde (approved):
Change looks good!

Instead of using isset, you could have used a somewhat more expressive `$responseNameId !== null`, `is_null`, or even a 'instanceof' type check.

This works too (I'm bugged by the fact that isset tests if a variable is set AND is not null) PHP is not super single responsibility-ish that way.

Feel free to leave this as is, as this is a matter of taste.

PR #921, opened 2020-10-03
Also add c14n method to the reference transforms in XML metadata.

thijskh:
This fixes parsing of our generated XML with some remote entities, notably ADFS. This was present in the old metadata (pre-twig) and is also present in e.g. the signed Responses we send and in metadata generated by other products.

XML before:

```xml
<ds:Reference URI="#EB...">
    <ds:Transforms>
        <ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
    </ds:Transforms>
    <ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"/>
    <ds:DigestValue>...</ds:DigestValue>
</ds:Reference>
```

XML after:
```xml
<ds:Reference URI="#EB...">
    <ds:Transforms>
        <ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
        <ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
    </ds:Transforms>
    <ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"/>
    <ds:DigestValue>...</ds:DigestValue>
</ds:Reference>
```

Other users do the same, see e.g. https://github.com/simplesamlphp/simplesamlphp/blob/d6d30c87235c3dc3c9690753d7da487b02613972/lib/SimpleSAML/Metadata/Signer.php#L284

Thanks @tvdijen for finding an analyzing the issue.

### DIFF
--- a/src/OpenConext/EngineBlock/Xml/DocumentSigner.php
+++ b/src/OpenConext/EngineBlock/Xml/DocumentSigner.php
@@ -42,12 +42,13 @@ class DocumentSigner
         $rootNode = $doc->childNodes[1];
 
         // Create sign object
+        $canonicalMethod = XMLSecurityDSig::EXC_C14N;
         $objDSig = new XMLSecurityDSig();
-        $objDSig->setCanonicalMethod(XMLSecurityDSig::EXC_C14N);
+        $objDSig->setCanonicalMethod($canonicalMethod);
         $objDSig->addReference(
             $rootNode,
             self::SIGN_ALGORITHM,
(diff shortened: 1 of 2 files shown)


MKodde:
Looks good! We could further cover this in test by adding some assertions to the `Metadata.feature` acceptance test. But seeing this verified in the unit test should be sufficient test coverage.

MKodde (approved):

PR #922, opened 2020-10-05
Remove the remaining eduGAIN metadafields

MKodde:
This PR consists of a migration that drops the last database table related eduGAIN references.

The commit warns: `To be released post EngineBlock 6.2 for rollbackability` Now that we are in the 6.3 tier, this can be merged now.

### DIFF
--- a/database/DoctrineMigrations/Version20191011132428.php
+++ b/database/DoctrineMigrations/Version20191011132428.php
@@ -15,6 +15,11 @@ final class Version20191011132428 extends AbstractMigration
         // this up() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
 
+        $this->skipIf(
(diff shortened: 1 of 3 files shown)


thijskh (approved):

thijskh (changes requested):
On production we have publish_in_edugain but not publish_in_edu_gain_date. So deploying this would break on the migration I think

tvdijen:
I have both on all my instances..

MKodde:
I was not aware of an awesome Doctrine Migration feature where you can check if a certain column or index exists before altering it.

I used that on both Migrations that where causing issues on the 6.3.x migration path.

tvdijen:
This should probably be used for adding / removing colums as well, instead of fiddling around with SQL queries..
The only downside I've seen so far is that it doesn't support a specific location for the new column..

thijskh (approved):

MKodde:
@tvdijen that's certainly worth looking into for the future. I usually record a migration using the diff tool. But (re)writing them using the provided methods might help creating more robust migrations.

For now I'm merging these changes.

PR #924, opened 2020-10-06
Ensure you can select an idp both with and without JS

Badlapje:
Added tabindexes, focus & hover styles.
Added a description for the article, as it get's focus.
Cleaned up.


MKodde (commented):
The keyboard navigation (using TAB) works as expected. The previous WAYF also had support for using the up and down keys. 

Pressing down would select the first IdP item. And from there the arrow keys would allow you to navigate the list of IdP's

It might be nice to integrate that keyboard behaviour in this PR?

Regarding the no-js solution. I love it, having the login button for each IdP entry is very intuitive. The keyboard navigation can maybe be optimized. As tabbing through the list first selects the IdP list item, and then the button. Requiring double tabbing to reach the next IdP. Is there a nice way to streamline this?

And I observed a style related issue with the no-js version of the page. The no-js warning text is rendered outside the main content. Making it hard to read (blue on black, see screenshot) Please feel free to fix this another time!
Screenshot_2020-10-06 OpenConext - Select an organisation to login to the service

Badlapje:
I've added the keyboard listener to go up and down in the list.

Sadly: we can't optimize the tab behaviour of the no-js version.  Because of the lack of js we need that button to be focusable.
The idp itself is focusable because in the JS version, that's what is focused on.

We could make two versions: one for no-js without the tabindex on the idp, but that'd add quite a lot of markup to the page.  So i reckon this is a minor annoyance for keyboard users without JS turned on but a bigger annoyance for larger html size for all other users.

Regarding the styling of the no-js: i'm waiting for a design from Edwin still.  I'll remind him of that.

MKodde:
> I've added the keyboard listener to go up and down in the list.

Going to test drive that right now!

> We could make two versions: one for no-js without the tabindex on the idp, but that'd add quite a lot of markup to the page. So i reckon this is a minor annoyance for keyboard users without JS turned on but a bigger annoyance for larger html size for all other users.

I see, looks like a good trade-off this way to me. Maybe ask our PM's to vent their opinion on this point?

theme/base/javascripts/wayf/keyboardBehaviour.js
```diff
@@ -28,6 +40,37 @@ export const keyboardBehaviour = () => {
...
+        lastIdp.focus();
+      } else if (isFocusOn(firstIdp)) {
+        eduId.focus();
+        return;
+      } else if (isFocusOn(eduId)) {
+        searchBar.focus();
+      }
+
+      focusOnPreviousIdp();
+      return;
```
MKodde: Nice, we almost have the same behaviour as the 'openconext' WAYF. One distinction is that on that WAYF, the search bar is focussed on load. Maybe we should copy that behaviour (some time later)

You probably figured this out already, but if you'd like to test a real world WAYF against a tried and tested build from before our UI refactor work, simply visit: `https://profile.surfconext.nl`

MKodde (approved):

theme/base/javascripts/wayf/keyboardBehaviour.js
Badlapje: I've been thinking about that focus.  We can do that in the html, so no JS needed there.  But i need to check with Thijs on this, as the searchbar isn't always visible.  I need to know where the focus goes if we have a previous selection list.

Badlapje: good tip on the profile.surfconext 💯

PR #925, opened 2020-10-06
Add and refactor integration tests for the WAYF

Badlapje:
Split the tests into three files:
- generic behaviour: does not depend on clicking / entering
- keyboard behaviour: depends on enter input
- mouse behaviur: depends on clicks


theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/search.html.twig
```diff
@@ -3,8 +3,9 @@
     <input type="hidden" id="form-idp" name="idp" value=""/>{# needed for the ContinueToIdp module #}
     <label for="wayf_search" class="visually-hidden">{{ 'search'|trans }}</label>
     <input
-        id="wayf_search"
+        autofocus
```
MKodde, line 6: Didn't know about this attribute! 

I see it is safe to use for us. But will not work on older IE browsers. We do not support those, but we do best effort support. Don't think this important enough to kick that principle into play here.

theme/cypress/integration/skeune/wayf/WayfMouseBehaviour.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+context('WAYF when using the mouse', () => {
+  beforeEach(() => {
+    cy.visit('https://engine.vm.openconext.org/functional-testing/wayf');
+  });
+
+  it('Should login when selecting an idp', () => {
+    cy.get('.wayf__remainingIdps .wayf__idp')
+      .eq(1)
+      .click({force: true});
```
MKodde, line 12: Ah, the good ol clicke force!

MKodde (approved):
No objections on this one!

PR #928, opened 2020-10-06
Style noscript messages

Badlapje:


theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/noScript.html.twig
```diff
@@ -1,3 +1,6 @@
 <noscript>
-    {{ 'wayf_noscript_warning'|trans|raw }}
+    {% include '@theme/Default/Partials/warning.html.twig' with {
+        className: 'wayf__noScript',
+        text: 'wayf_noscript_warning'|trans|raw
```
MKodde, line 4: This `raw` filter is later filtered again as it seems. I fear you have to filter the value in the `warning.html.twig ` partial.
Screenshot_2020-10-07 OpenConext - Select an organisation to login to the service

MKodde (commented):
One small point of feedback, other then that, LGTM!

theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/noScript.html.twig
Badlapje: nice catch!  

Fixed it, and in the same go also removed _skeune prefix from the translation name + ensured the html is always correct for the text content.

MKodde (approved):
Yeah! :shipit:

PR #929, opened 2020-10-07
Add all previous selection functionality

Badlapje:
Add selected idp to cookie
Remove deleted idp from cookie
Increase count on selected "previous" selected idp
Remove deleted idp from previous selection
Move deleted idp to remaining idps
Reindex idps
Put focus on the first previously selected idp remaining
Hide previous selection when there's no more elements & focus on searchbar


theme/base/javascripts/utility/setCookie.js
```diff
@@ -0,0 +1,13 @@
...
+ * Save a cookie, by default for 365 days
+ *
+ * @param content       the content you want to put in it
+ * @param cookieName    the name for the cookie
+ * @param expires       when the cookie expires (number)
+ * @param path          on which domain to save it
+ */
+export const setCookie = (content, cookieName, expires = 365, path = '/') => {
+  Cookies.set(cookieName, content, { expires: expires, path: path });
+};
```
MKodde, line 13: Nice way to set a cookie! Maybe specify additional parameters to set cookie security related options (httponly, secure, and samesite even?)

For now this reflects the current situation. We might want to discuss if the current situation is still good enough, given this was written 4 years ago.

theme/base/javascripts/wayf/deleteDisable/deleteIdp.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Delete an idp from the previous selection in both html & the previous selection list
+ *
+ * @param element               HTMLElement   the deleted idp
+ * @param previousSelection     Array         the list of previously selected idps
+ */
+export const deleteIdp = (element, previousSelection) => {
+  const configuration = JSON.parse(document.getElementById('wayf-configuration').innerHTML);
+  const cookieName = configuration.previousSelectionCookieName;
```
MKodde, line 11: :+1:

theme/base/javascripts/wayf/handleDeleteDisable.js
```diff
@@ -1,18 +1,40 @@
...
 
-  // in case of click the origin is the span, so setting the element needs to be done differently
+  // in case the origin is the span setting the element needs to be done differently
   if (e.target.tagName === 'SPAN') {
     element = e.target.closest('.idp__deleteDisable');
   }
 
   // handle clicking disabled button
   if (element.querySelector('.idp__delete').style.display === 'none') {
+    return; // todo work this out
```
MKodde, line 24: Something you want to work out at a later point? Then maybe add it to the 'leftovers' ticket

theme/cypress/integration/skeune/wayf/WayfKeyboardBehaviour.js
```diff
@@ -19,29 +16,32 @@ context('WAYF when using the keyboard', () => {
...
-    cy.pressArrowOnIdpList('up', 'search__field');
-    cy.pressArrowOnIdpList('up', 'wayf__idp', '5');
-    cy.pressArrowOnIdpList('up', 'wayf__idp', '4');
-    cy.pressArrowOnIdpList('up', 'wayf__idp', '3');
-    cy.pressArrowOnIdpList('up', 'wayf__idp', '2');
-    cy.pressArrowOnIdpList('up', 'wayf__idp', '1');
-    cy.pressArrowOnIdpList('up', 'wayf__eduIdLink');
-    cy.pressArrowOnIdpList('up', 'search__field');
+  describe('Should be able to traverse the remaining idp section with arrow keys', () => {
+    it('check if pressing down works as expected', () => {
```
MKodde, line 20: Out of curiosity, what difference does this make? Adding a `describe` around the specification.

MKodde (approved):
Looks very good! See some thoughts in the comments below. Nothing that should block a merge.

theme/base/javascripts/wayf/handleDeleteDisable.js
Badlapje: Handling disabled Idps is on my list right after adding search functionality.  By disabled i mean the ones with no access.  It's on my todo list right after search (so hopefully by monday you'll see a PR about it)

theme/cypress/integration/skeune/wayf/WayfKeyboardBehaviour.js
Badlapje: Well, we had several tests where we test 'complicated' behaviour with lots of steps.  We put those in the tests files as one single test, but it's actually a suite of tests to test one "set" of related behaviour.  

By adding a describe and then using `it('name of step being tested ... ` it gives a nice overview.  Check it out:
image

By making it easier to see what sub-steps there are, it's a lot easier to debug a failing test.  Instead of seeing what command was given, you can just go to the specific sub-step test.

Badlapje: one thing to take note of though is that the beforeEach hook does run for every "it" instance.

MKodde (approved):
Screenshot_2020-10-08 OpenConext - Select an organisation to login to the service

:arrow_up: 

Two small findings that probably should not be addressed in this PR.

theme/base/javascripts/utility/setCookie.js
Badlapje: aha, i like it :-)  Putting it on the to-do list.  Can take it up with the remember my choice functionality.

theme/cypress/integration/skeune/wayf/WayfKeyboardBehaviour.js
MKodde: Ah, yes this way the tests are organized way nicer! I now understand your statement on Slack just now.

theme/base/javascripts/wayf/handleDeleteDisable.js
MKodde: Ah, I think my functional review finding below relates stronly to this TODO.

Badlapje:
Those are bugs.  I added some logic in the twig templates on when to show the no-access version (which i dubbed "disabled") and when to show the delete button.  But i wasn't sure yet how to determine the no-access state.  You clarified that for me yesterday, so i'll take these two up along with the no-access functionality.

Nice find ^__^

PR #931, opened 2020-10-09
Fix sequence / schema violation

tvdijen:
As per Slack-discussion, the recent changes in https://github.com/OpenConext/OpenConext-engineblock/pull/878 cause a schema violation, because a specific order of elements is demanded.

Also see par [2.3.2.1](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) of the specs


tvdijen:
As per Slack-discussion, the recent changes in https://github.com/OpenConext/OpenConext-engineblock/pull/878/files cause a schema violation, because a specific order of elements is demanded.

thijskh (approved):

MKodde (approved):

PR #932, opened 2020-10-09
Refactor previous selection for speed

Badlapje:
By handling the visibility of the previous selection in JS, we had a flash of content (FOC) briefly, which on a slow connection was incrementally worse.  The visbility is now handled in the twig template.  This means:
- we no longer have a FOC
- we have a faster page load
- we have less JS => faster page load again & faster time to interactive


theme/base/javascripts/wayf/utility/sortIdpList.js
```diff
@@ -0,0 +1,31 @@
+import {sortByTitle} from './sortIdpMethods';
+
+const TITLE = 'title';
+
+/**
+ * Sort an idpList.  By default it's by displayTitle.
+ * No other sorts exist atm, but this is anticipated once the search is implemented.
```
MKodde, line 7: Mental note to remove this once other methods are implemented.

theme/base/javascripts/wayf/utility/sortIdpMethods.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param firstElement
+ * @param secondElement
+ * @returns {number}
+ */
+export const sortByTitle = (firstElement, secondElement) => {
+  const titleOne = getTitle(firstElement.children[0]);
+  const titleTwo = getTitle(secondElement.children[0]);
+
+  return titleOne.localeCompare(titleTwo);
```
MKodde, line 14: Reading in the `localeCompare` man pages, you could pass the current locale as a parameter to get more language specific ordering results. Not sure this is valuable at this point, but certainly possible.

theme/base/javascripts/wayf/utility/sortAndReindex.js
```diff
@@ -0,0 +1,57 @@
...
+ * @parameter sortBy  string    the attribute to sort by
+ * @parameter focus   boolean   whether or not to focus the first item
+ */
+import {sortIdpList} from './sortIdpList';
+
+const PREVIOUS = 'previous';
+const REMAINING = 'remaining';
+
+export const sortAndReindex = (list = REMAINING, sortBy = 'title', focus = false) => {
+  let idpListSelector = '.wayf__remainingIdps .wayf__idpList';
```
MKodde, line 19: Not sure, but would injecting the query selectors not decrease coupling to the dom elements? And hence improve quality/maintainability of this function?

It might not be practical, and in that case maybe this is the best way.

theme/cypress/integration/skeune/wayf/WayfMouseBehaviour.js
```diff
@@ -85,13 +85,17 @@ context('WAYF when using the mouse', () => {
...
       });
 
-      it('Test the last deleted idp is at the top of the list', () => {
-        cy.get('.wayf__remainingIdps .wayf__idp[data-index="1"]').should('have.attr', 'data-entityid', 'https://example.com/entityId/2');
+      it('Test the last deleted idp is in the remaining list', () => {
+        cy.get('.wayf__remainingIdps .wayf__idp[data-entityid="https://example.com/entityId/2"]').should('exist');
+      });
+
+      it('Test the remaining list is sorted alphabetically', () => {
+        cy.get('.wayf__remainingIdps li:first-of-type .wayf__idp').should('have.attr', 'data-index', '1');
```
MKodde, line 98: Adding tests: 

:woman_dancing:

theme/base/javascripts/wayf/handleDeleteDisable.js
```diff
@@ -28,13 +31,15 @@ export const handleDeleteDisable = (e, previouslySelectedIdps) => {
...
   // If no items are left: do what's needed.
   if (!hasSelectedIdps()) {
     handleNoneLeft();
     return;
   }
 
-  // If there are items left: reindex & focus first one
-  reindexAndFocus();
+  // If there are items left: sort, reindex & focus first one
+  sortAndReindexPrevious('title', true);
```
MKodde, line 44: Not sure on the latest and greatest JS best practices. But it may be better to separate the sorting and indexing into two methods? This is merely based on a smelly function name. 

Looking at the actual implementation of the sortAndReindex methods, they seem to do a cohesive task on the IdP list. Maybe renaming this method would remove the smell. Ore simply leaving it like this is fine with me too!

MKodde (commented):
Interesting change set. I added some thoughts and concerns in the comments below. I think this is good enough to move forward with, but maybe you want to change some stuff based on my review. It's up to you.

theme/base/javascripts/wayf/utility/sortIdpMethods.js
Badlapje: Good catch.  It's a small speed cost for potentially better results.  
On the other hand: it's not yet universally supported (mostly on mobile) and I'm not sure whether passing the options argument makes the method fail in those mobile browsers.  Added is the fact that our supported languages would have very little benefit from it.

Is adding this to the leftover ticket ok with you?  If we have time it would be nice to learn more on if this breaks mobile or not.

theme/base/javascripts/wayf/utility/sortAndReindex.js
Badlapje: Without a state store it would've had to go through too many layers & cluttered up the methods in between with too many parameters.

But i made those first three lines into a separate function, so it's a tad less tightly coupled & easier to overwrite if desired.  Well spotted!

theme/base/javascripts/wayf/handleDeleteDisable.js
Badlapje: Good one :-)  Will refactor it in the search PR as i already made reinsert into a function there.  Saves a few rebase issues that way if it's ok with you?

Badlapje: done

PR #933, opened 2020-10-10
Implement search functionality

Badlapje:
Filter by searchterm & sort by weight.  Autohide all items with a weight of 0.


theme/base/javascripts/wayf/search/searchAndSortIdps.js
```diff
@@ -0,0 +1,40 @@
...
+ * @param     searchTerm     string
+ * @returns   node[]
+ */
+export const searchAndSortIdps = (idpArray, searchTerm) => {
+  // reset weights by removing them
+  clearWeight(idpArray);
+
+  if (typeof searchTerm !== 'undefined' && searchTerm.length) {
+    findWeight(idpArray, searchTerm.toLowerCase());
+    idpArray.sort(sortByWeight);
```
MKodde, line 31: Is it common practice in JS land to use data attributes (with a search-match-weight) to sort a list? It looks like a pragmatic solution to the problem, but it does keep on stowing properties on the list items. Would using some sort of a data store not be 'better'?

theme/base/javascripts/wayf/search/findWeight.js
```diff
@@ -0,0 +1,59 @@
...
+import {setWeight} from './setWeight';
+import {
+  checkPartialMatchId,
+  checkPartialMatchKeywords,
+  checkPartialMatchTitle
+} from './checkPartialMatch';
+
+const EXACT_TITLE = 100;
+const EXACT_ID = 60;
+const EXACT_KEYWORD = 60;
```
MKodde, line 11: A good starting point, but maybe these values should be configurable to more easily demand changes to the algorithm in the future.

theme/base/javascripts/wayf/search/findWeight.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * A weight is set for each entityId according to how well it matches:
+ * - perfect match: title = 100, id & keyword = 60
+ * - partial match: title = 30, id = 20, keyword = 25
+ * If a perfect match is found, the search stops for that idp.
+ *
+ * @param     idpArray        array of idp nodes
+ * @param     searchTerm     string
+ */
+export const findWeight = (idpArray, searchTerm) => {
```
MKodde, line 30: I strongly feel this function containing quite a lot of business logic should be covered in unit tests.

theme/cypress/integration/skeune/wayf/WayfGeneralBehaviour.js
```diff
@@ -2,45 +2,82 @@
...
+      cy.get('.wayf__search').type('OllekebollekeKnol');
+      cy.get('.wayf__noResults').should('be.visible');
+    });
+
+    it('Should be able to search for an idp', () => {
+      cy.visit('https://engine.vm.openconext.org/functional-testing/wayf');
+      cy.get('.wayf__search').type('IdP 4');
+
+      // After filtering the search results, verify one result is visible
+      cy.get('.wayf__remainingIdps .wayf__idp:not([data-weight="0"])').should('have.length', 1).should('contain.text', 'Connected IdP 4 en');
```
MKodde, line 43: Maybe make these assertions multi-line statements for improved readability.

:1st_place_medal:  for these tests!

I still think we should add unit tests to cover the test BL.

MKodde (changes requested):
Superb work. You managed to neatly divide an conquer the search requirements. I think increasing the unit test coverage should be added to this PR though. 

Feel free to IM me on what I have in mind or to brainstorm on how we want to go about this.

I'm going to test this in-browser now. Might come back with additional findings later.

**Edit after functional review**
- Searching works real nice!
- The text matching when using multiple search terms could be improved. See example below.

IdP List: `IdP 1, IdP 2, IdP 3`
Search term: 'IdP 2'
Yields one 100% match: IdP 2 :heavy_check_mark: 

Search term: '2 IdP'
Yields no search result :disappointed: 

Suggestion: explode the search terms on whitespaces, calculate weight for each search term. Set the average of the weight matches as the combined weight. This can be further tweaked if need be.

theme/base/javascripts/wayf/search/searchAndSortIdps.js
MKodde: FYI I'm not advocating adding a costly data store to this project, but would like to hear about your motivation on choosing data attributes.

Badlapje: As mentioned on the slack call: it's a common way to do this for those who prefer to not use a framework as a sort of lightweight data store.  In our case i don't think a data store makes sense as it's really only needed for a few things & we care about snappiness.

theme/base/javascripts/wayf/search/findWeight.js
Badlapje: will ask PM tomorrow

theme/cypress/integration/skeune/wayf/WayfGeneralBehaviour.js
Badlapje: ❤️ that suggestion :-)  Refactored for readability.

theme/base/javascripts/wayf/search/findWeight.js
```diff
@@ -0,0 +1,59 @@
...
+  idpArray.forEach(li => {
+    const idp = li.children[0];
+    const title = getData(idp, 'title');
+    if (title === searchTerm) {
+      console.log('exact match on title');
```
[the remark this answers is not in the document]
Badlapje: removed

theme/base/javascripts/wayf/search/findWeight.js
Badlapje: refactored to allow for easier unit testing

Badlapje: Added unit tests

theme/cypress/integration/skeune/wayf/unit-tests/assignWeight.spec.js
```diff
@@ -0,0 +1,50 @@
+import {assignWeight} from '../../../../../base/javascripts/wayf/search/assignWeight';
+
+context('unit test the assignWeight function', () => {
+  describe('Test multiple hits', () => {
+    // should show 100 + 30 + 25 + 18 + 17 = 187
```
MKodde, line 5: Sounds fair! My intuition said: going past 100 is a bad thing, but thinking about it, this would allow us to express a full match, and add additional matching score to the additional matches.

MKodde (commented):
I'm very happy with the new unit tests! Thanks for adding those.

Next up, another quick test drive in the browser :racehorse:

MKodde:
Loved the search experience much better than on the previous test drive. Additional findings:

The whitespace itself is also included in the matching logic so it seems. This might not be expected behavior? Let me try to explain: 

1) There are three IdP's 'Mujina IdP', 'The Mended Drum' and 'ROC van Twente'. 
Searching for ROC yields a single search result
Adding a whitespace to the search results gives me back all IdP's, ROC van Twente is still up top which is very good, but I'd rather not yet see the other IdPs. They should appear once an actual second search term is entered. 

2) Given the same IdP setup described in 1). Searching for 'Mended Twe' shows ROC van Twente before 'The Mended Drum'. In my opinion the match on the title is stronger for 'The Mended Drum'. There migth be a match on the entity id going on that is taken into account. 

3) The console is spammed with debug messages again.

Badlapje:
1. fixed.  Some weird string.split behaviour there, but easily resolved with a trim.
2. that sounds right to me: R comes before C, so with equal weight the ROC van Twente would come before The Mended Drum.  We are searching:
- for a match with the entire title/keyword/entityId
- for a partial match with the title / a keyword / the entityId
So we are not yet looking for a full match with part of the title or part of a keyword.  Your logic is sound though, i'll see if i can add that without reducing the snappiness.
3. removed those

MKodde (approved):
Cool, nice to see that changes in the search algo can be made with relatively little effort

PR #934, opened 2020-10-11
Add remember choice support

Badlapje:
Also added some security features

### DIFF
--- a/theme/base/javascripts/utility/setCookie.js
+++ b/theme/base/javascripts/utility/setCookie.js
@@ -3,11 +3,15 @@ import * as Cookies from 'js-cookie';
 /**
  * Save a cookie, by default for 365 days
  *
+ * Note: httpOnly is not an option here as it's not supported by js-cookie atm.
+ *
  * @param content       the content you want to put in it
  * @param cookieName    the name for the cookie
- * @param expires       when the cookie expires (number)
+ * @param expires       when the cookie expires (number of days)
  * @param path          on which domain to save it
+ * @param secure        whether or not it's a secure cookie
+ * @param sameSite      whether or not it's samesite
  */
-export const setCookie = (content, cookieName, expires = 365, path = '/') => {
-  Cookies.set(cookieName, content, { expires: expires, path: path });
+export const setCookie = (content, cookieName, expires = 365, path = '/', secure = true, sameSite = 'strict') => {
+  Cookies.set(cookieName, content, { expires: expires, path: path, secure: secure, sameSite: sameSite });
 };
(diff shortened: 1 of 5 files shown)


theme/base/javascripts/utility/setCookie.js
```diff
@@ -3,11 +3,15 @@ import * as Cookies from 'js-cookie';
...
  * @param cookieName    the name for the cookie
- * @param expires       when the cookie expires (number)
+ * @param expires       when the cookie expires (number of days)
  * @param path          on which domain to save it
+ * @param secure        whether or not it's a secure cookie
+ * @param sameSite      whether or not it's samesite
  */
-export const setCookie = (content, cookieName, expires = 365, path = '/') => {
-  Cookies.set(cookieName, content, { expires: expires, path: path });
+export const setCookie = (content, cookieName, expires = 365, path = '/', secure = true, sameSite = 'strict') => {
```
MKodde, line 15: Great! Adding these features looks very useful. 

Consider:
1. Adding input validation, allowing only valid values to be set
2. Ensure the current behaviour is maintained. It seems the remember choice cookie will now become more restricted. Which might be a good thing, but might lead to side effects.

Badlapje: Added input validation.  

Not sure i can test the remember choice cookie atm.  So unsure of where to find possible side effects.  Figured we can give it a testrun and see if something breaks?

Badlapje: Which is my least favourite kind of testing

Badlapje: But the integration tests keep working :-)

MKodde: I think, the best way to test the feature is trying it in a 'real world' dev authentication flow. You'd have to enable the feature in EB. After a successful remember me authentication, successive authentications should no longer show the WAYF.

theme/base/javascripts/utility/setCookie.js
```diff
@@ -13,5 +13,10 @@ import * as Cookies from 'js-cookie';
  * @param sameSite      whether or not it's samesite
  */
 export const setCookie = (content, cookieName, expires = 365, path = '/', secure = true, sameSite = 'strict') => {
+  if(typeof cookieName !== 'string') return;
+  if(typeof expires !== 'number') return;
+  if(typeof path !== 'string') return;
+  if(typeof secure !== 'boolean') return;
+  if(typeof sameSite !== 'string') return;
```
MKodde, line 20: Nice! To further bullet proof this, you could add tests for possible values. Samesite: [lax|strict|none] But hey, those are marginal gains. Thanks for adding this input validation.

MKodde (approved):

PR #936, opened 2020-10-13
Introduce display_default_idp_banner_on_wayf feature flag

MKodde:
This feature flag toggles the default IdP banner on the WAYF in the preselection twig template.

By default this feature is enabled and can be set in the parameters.yml

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -166,6 +166,9 @@ parameters:
     ## Allow users to save their selected IdP and then auto-select it on returning visits.
     wayf.remember_choice: false
 
+    ## Toggle the default IdP quick link banner on the WAYF.
+    wayf.display_default_idp_banner_on_wayf: true
+
     ## Settings for detecting whether the user is stuck in a authentication loop within his session
     time_frame_for_authentication_loop_in_seconds: 60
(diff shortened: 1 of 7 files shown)


Badlapje (changes requested):
Code looks great!  It's cool how it's setup so it can easily be extended with a new flag 💯 

Can you also add an integration test for this?  In WayfGeneralBehaviour, line 99 we now test whether the eduId is being shown.  You could add a test there to ensure that it's not shown when we use `&showIdpBanner=false`.  When looking at this i noticed your branch is lagging behind the feature/ui-redesign one.

When i tried this out, it didn't work for me.  But maybe i'm doing something wrong?  I added the parameter in parameters.yml and then rebuilt the theme.

MKodde:
Thanks for your review!

1. I rebased my changes on the base branch.
2. Added an e2e test to the general behavior test. Top suggestion!
3. Using the feature toggle IRL does not require a theme rebuild, but does rely on a cache rebuild. The DI container needs to be rebuild in order to facilitate the feature flag. When debug mode is enabled in your Symfony env, this should happen automatically. Otherwise: `$ php72 ./app/console ca:cl --env=dev`

Badlapje (approved):
🚀

PR #938, opened 2020-10-15
Support cutoff point feature flag

Badlapje:
Support the cutoff point feature flag.  This means hiding all remaining Idps on the page when there would be more, unless the user is searching.


MKodde (changes requested):
The change is looking good. As far a I could tell the same behaviour is shown as in the openconext/material theme.

There however are 5 conflicting tests. Maybe this is another ghost in the system, or maybe this change requires changes to the exising tags. I'm unable to inspect this at this moment.

Badlapje:
Will check the tests after merging the other open PRs, as that'll give me a clearer picture.

MKodde (approved):

PR #939, opened 2020-10-15
Reinstate the back to SP link

MKodde:
The back to SP link was added to the base template.

Todo:
 - The `focus` style shows the border with to much margin, causing a not so nice render on the screen. @Badlapje can you help?
 - Tests for this feature should be added


Badlapje:
on it ;-)

theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/backButton.html.twig
```diff
@@ -0,0 +1,9 @@
+{% if backLink is defined and backLink == true %}
```
Badlapje, line 1: shouldn't this file be called `backLink.html.twig`?  And the class `wayf__backLink`?  I'm fine either way though.

Badlapje (commented):
Looking good 👍 

One question though: why was NunitoOFL.txt removed?  License info like that should not go missing imo.  We could add it as a comment to fonts.scss?

If we do remove it, then probably we should delete `-e base/stylesheets/fonts/*.txt` from the build:copy script in package.json.

theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/backButton.html.twig
MKodde: Very sharp, seeing this is a link and not a button! Will change that

MKodde:
> One question though: why was NunitoOFL.txt removed? License info like that should not go missing imo. We could add it as a comment to fonts.scss?

Personally I'm fine with keeping the file, but then we must update the git ignore file to prevent adding it unintentionally. Also we should think/discuss on how we should go about using these fonts and how we'd like to advertise we are using this font. Maybe your comment is a good proposal. Let's discuss this in the stand up or on the Slack channel

Badlapje (approved):
Nice! 💯

PR #940, opened 2020-10-17
Ensure previous selection is hidden after clicking on the add account button

Badlapje:
Fix for the bug as discovered by Edwin.

Did some minor housekeeping while at it.


MKodde (approved):
Looks good!

PR #941, opened 2020-10-17
Add accessibility support for aria-pressed attribute on labels in the WAYF

Badlapje:
Refactored it to be more general in consent, then added it in the WAYF.

### DIFF
--- a/theme/base/javascripts/utility/handleAriaPressed.js
+++ b/theme/base/javascripts/utility/handleAriaPressed.js
@@ -1,0 +1,22 @@
+/**
+ * Add aria-pressed support to elements with that attribute.
+ *
+ * @param elements
+ */
+export const handleAriaPressed = (elements) => {
+  for (let i = 0; i < elements.length; i++) {
+    elements[i].addEventListener('click', function(e) {
+      const element = e.target;
+      const ariaPressed = element.getAttribute('aria-pressed');
+      const newValue = {
+        false: true,
+        true: false,
+      };
+
+      // We use the newValue object because ariaPressed is a string containing either "true" or "false".
+      // Casting that to a boolean will, in both instances, give us true.
+      // By using an object we can use array notation to access the right value based on what we get back from the attribute.
+      element.setAttribute('aria-pressed', newValue[ariaPressed]);
+    });
+  }
+};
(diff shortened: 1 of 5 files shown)


theme/base/javascripts/utility/handleAriaPressed.js
```diff
@@ -0,0 +1,18 @@
...
+export const handleAriaPressed = (elements) => {
+  for (let i = 0; i < elements.length; i++) {
+    elements[i].addEventListener('click', function(e) {
+      const element = e.target;
+      const ariaPressed = element.getAttribute('aria-pressed');
+      const newValue = {
+        false: true,
+        true: false,
+      };
+      element.setAttribute('aria-pressed', newValue[ariaPressed]);
```
MKodde, line 15: The newValue solution can maybe be simplified by setting the attribute value with a negation of the `ariaPressed` constant?

Example:
```js
       const ariaPressed = element.getAttribute('aria-pressed');
       element.setAttribute('aria-pressed', !ariaPressed);
```

MKodde (commented):
Looks good! Maybe address my suggestion below?

theme/base/javascripts/utility/handleAriaPressed.js
Badlapje: Sadly not.  Either set or get attribute turns a boolean into a string (pretty sure it's the set).  Both the string "true" and the string "false" revert to the boolean true when cast to a boolean in JS (as they are not empty strings).

MKodde: Ah, okay, but the values in the new value object are of type boolean too right? I don't see the difference here. If such a construction is used, it might be good to set/link an explanation as to why this cause of action was required.

Badlapje: the values are boolean, but they are accessed using a string.  Because the value we get via setAttribute is a string, we cannot do `newValue.ariaPressed`.  Newvalue is an object, and i use the "array-notation" to access it's properties.  This allows for using the strings.  

So `newValue[ariaPressed]` translates to `newValue['true']` (for example) which means we do `newValue.true`.

Badlapje: added to the code as an explanation

MKodde (approved):
Thanks for adding the explanation!

PR #942, opened 2020-10-18
Preselect the banner IDP when clicking the banner link

Badlapje:
When clicking on the banner IDP link, the focus should shift to the banner idp.

### DIFF
--- a/theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/remainingIdps.html.twig
+++ b/theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/remainingIdps.html.twig
@@ -9,9 +9,16 @@
     {% include '@theme/Authentication/View/Proxy/Partials/WAYF/search.html.twig' %}
 
     {% if showIdPBanner %}
+        {% set requestUri %}
+            {% if '?' in app.request.requestUri %}
+                {{ app.request.requestUri|replace({'?': '#defaultIdp?'}) }}
+            {% else %}
+                {{ app.request.requestUri ~ '#defaultIdp' }}
+            {% endif %}
+        {% endset %}
         {% include '@theme/Default/Partials/informational.html.twig' with {
             class: 'remainingIdps__eduId',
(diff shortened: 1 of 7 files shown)


MKodde:
#943 might help you in these dev-endeavors.

theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/remainingIdps.html.twig
```diff
@@ -9,9 +9,16 @@
     {% include '@theme/Authentication/View/Proxy/Partials/WAYF/search.html.twig' %}
 
     {% if showIdPBanner %}
+        {% set requestUri %}
+            {% if '?' in app.request.requestUri %}
+                {{ app.request.requestUri|replace({'?': '#defaultIdp?'}) }}
+            {% else %}
+                {{ app.request.requestUri ~ '#defaultIdp' }}
+            {% endif %}
```
MKodde, line 17: Great attention to detail!

MKodde (approved):
Nice work! I tested this in my Firefox. The scrolling behavior is functional, but not the entire article comes into view. Maybe overscroll by a couple of ems?

Aside from that, go ahead!

Badlapje:
The problem is that the behaviour of the scrollntoView differs per browser (here comes the rabbit hole):
- in firefox: 
  - a first click wil scroll the item into view, but only just.
  - a second click will scroll the item to the top of the viewport
- in chrome en edgium: it will the item halfway the viewport
- in ie11: it doesn't work

Now, the above behaviour is despite me using the true argument explicitly.  According to the spec that should ensure that:

> If true, the top of the element will be aligned to the top of the visible area of the scrollable ancestor. Corresponds to scrollIntoViewOptions: {block: "start", inline: "nearest"}. This is the default value.

Waiting for broader feedback on this to ensure that i don't loose time implementing behaviour that turns out to not fit the expectations.  Putting it on my list of things to sollicit feedback about.

PR #943, opened 2020-10-19
Add the default IdP EntityID configuration parameter

MKodde:
Allow configuration of the default IdP entity ID. This id is then used to mark the default IdP in the wayf list. Or at least to easily select it when the link in the banner is clicked.

To test this feature, use this qurey param in the function testing wayf

https://engine.vm.openconext.org/functional-testing/wayf?defaultIdpEntityId=https://example.com/entityId/3

### DIFF
--- a/theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/idp/idp.html.twig
+++ b/theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/idp/idp.html.twig
@@ -6,6 +6,7 @@
     data-index="{{ loop.index }}"
     data-keywords="{{ idp['keywords']|lower }}"
     data-title="{{ idp['displayTitle']|lower }}"
+    data-is-default-idp="{{ idp['isDefaultIdp'] ? 'true' : 'false' }}"
     tabindex="0"
 >
     <picture class="idp__logo">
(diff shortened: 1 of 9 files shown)


theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/idp/idp.html.twig
```diff
@@ -6,6 +6,7 @@
     data-index="{{ loop.index }}"
     data-keywords="{{ idp['keywords']|lower }}"
     data-title="{{ idp['displayTitle']|lower }}"
+    data-is-default-idp="{{ idp['isDefaultIdp'] ? 'true' : 'false' }}"
```
Badlapje, line 9: This should be an id rather than a data-attribute.  That allows us to link to it easily from the banner-link without JS.  Something like this (not sure about the name):

`{% if id['isDefaultIdp'] %}id="defaultIdp"{% endif %}`

Badlapje (approved):
Looking great 💯 

Just one small change to easily link to the default idp without JS.

Badlapje (changes requested):
hum, this time with the right option on the list 😊

Badlapje (approved):

Badlapje:
Looks great 🚀

PR #944, opened 2020-10-19
Add support for animations

Badlapje:
Add animations in an accessible way.  

For consent:
- tooltips
- modals
- show more button
- no consent button

For wayf:
- showing / hiding the previous/remaining section (a tad better, but couldn't really make it nice)


MKodde (commented):
Code wise I don't see any red flags. I'm by no means an animation expert, but I like the defensive/accessible approach you took.

Regarding the annimation, maybe it should be a tad quicker? Also personally I do not like the text sliding in. It makes it hard to start reading the text while the animation is playing. (talking about the modal/tooltip sceanrio)

Badlapje:
The text doesn't slide in anymore & it's twice as fast.

PR #945, opened 2020-10-20
Move error page stylesheets to base theme

MKodde:
Moves the openconext error page styling to the base theme

Things to concider: 
FontAwesome is used to display the icons on the error page. We might not like that?


theme/base/stylesheets/error-page.sass
```diff
@@ -0,0 +1,21 @@
...
+
+@import "../../node_modules/font-awesome-sass/assets/stylesheets/font-awesome"
+@import "../../openconext/stylesheets/shared/utilities/reset.css.sass"
+@import "../../openconext/stylesheets/shared/utilities/mixins.css.sass"
+
+@import "base/stylesheets/error-page/base/page.css.sass"
+@import "../../openconext/stylesheets/shared/base/type.css.sass"
+
+@import "../../openconext/stylesheets/shared/components/language.css.sass"
+@import "base/stylesheets/error-page/components/language.css.sass"
```
Badlapje, line 12: This doesn't seem to be used on the page?

Badlapje: Also the styles in it clash with those pulled in by the language.css.sass file from openconext on the line above.  There probably should be either one or none there i think?

Badlapje (approved):
Looks like a pragmatic solution :shipit: .  I'm not too crazy about the font-awesome icons and the vast amount of css added.  But we can weed the css once the page is redesigned.

theme/base/stylesheets/error-page.sass
MKodde: Wel spotted, will address this issue

MKodde: Looking into this. I think it best to remove all references to the openconext theme in the base theme. But this will result in adding openconext resets/utilities in the base theme. Let's discuss the best course of action.

PR #946, opened 2020-10-20
Re/move obsolete JS

Badlapje:
(Re)moved all the JS from the openconext theme.


MKodde (approved):
No JS related 404's in my console on both skeune and openconext themes, obsolete JS was removed. 

:100:

PR #947, opened 2020-10-20
Style idp images

Badlapje:
I've added some testimages for now to easily test the functionality.  These **must** be removed once testing is done.

To check during the standup tomorrow:
- how do we approach all institutions to renew their logo?
- how do we approach all institutions to add alt text for their logo?
- logo preferences (in order):
  - svg with paths/text/ ... and a title element containing the alt text.
  - svg with an image element, using a data-url as a href, and a title element containing the alt text.
  - optimized png: the smaller the better.

For those who wonder, width should be 75 px, but is adjust automatically, as is height.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "beberlei/assert": "^2.6",
+        "composer/package-versions-deprecated": "^1.11",
         "doctrine/doctrine-bundle": "^1.11",
         "doctrine/doctrine-migrations-bundle": "^1.3",
         "doctrine/orm": "~2.6",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a924cd537878a64579e06ece921d60eb",
+    "content-hash": "567af9a64ed2f33fa3d9bcb48ecdc2ac",
     "packages": [
         {
             "name": "beberlei/assert",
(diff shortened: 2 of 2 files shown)


MKodde:
Given the temporary nature of the logo images, It's not that important that the image copy task does not copy the files. I had to manually copy them to the web folder to make them work. Except for the SVG one, that worked without it being in the /web/images folder.

And what caught my eye was that some of the translations are no longer translated (wayf_search_placeholder, wayf_eduId). 

Changes to the code look good.

theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/idp/idp.html.twig
```diff
@@ -10,10 +10,16 @@
...
     <div class="idp__logo">
-        {% if 'svg' in idp['logo'] %}
-            {{ source('@' ~ idp['logo']) }}
+        {% if idp.logo is not null %}
+            {% set logoUrl = idp.logo %}
+        {% else %}
+            {% set logoUrl = '/images/placeholder.png' %}
+        {% endif %}
+
+        {% if 'svg' in logoUrl %}
```
MKodde, line 19: This check is not really bulletproof. whenever 'svg' is part of the uri, the image will be handled as if it is a SVG. This does not have to be the case. 

1. To make this a bit better, only test for the filename extension to be 'svg' and lowercase that portion of the string.
2. To make it even better. Try to see if the file type (mime) is of type 'image/svg+xml' (do we need to allow SVG's with JS embedded? in that case also add the application/svg+xml as well. Hightly doubt that should be done..

This check can maybe be done in the image VO: `src/OpenConext/EngineBlock/Metadata/Logo.php` by adding an 'isSvg' method.

MKodde (changes requested):
Thanks for adding support for the placeholder image! Looking at this PR again, I noticed the svg detection. See two suggestions for approval on how to make that more robust.

PR #948, opened 2020-10-20
Ensure all cypress commands are theme agnostic

Badlapje:
Ensure all custom cypress commands can be used by other themes as well


MKodde (approved):
Nice touch!

PR #950, opened 2020-10-21
Support minimal consent & fix a few translation bugs

Badlapje:
Wayf:
- Ensure that the organization noun is used everywhere.
- EduId needs to start with a small letter the banner.

Consent:
- add minimal consent variant

### DIFF
--- a/theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/ctas.html.twig
+++ b/theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/ctas.html.twig
@@ -4,10 +4,19 @@
         <input type="hidden" name="ID"
                value="{{ responseId }}">
         <input type="hidden" name="consent" value="yes"/>
+
+        {% if minimalConsent %}
+            {% set text_ok = 'consent_buttons_ok_minimal'|trans({'%arg1%': spName}) %}
+            {% set text_nok = 'consent_buttons_nok_minimal'|trans %}
+        {% else %}
+            {% set text_ok = 'consent_ok'|trans %}
+            {% set text_nok = 'consent_buttons_nok'|trans %}
+        {% endif %}
+
         {% include '@theme/Default/Partials/button.html.twig' with
             {
                 buttonClass: 'cta_consent_ok',
-                text: 'consent_ok'|trans
+                text: text_ok
             }
         %}
     </form>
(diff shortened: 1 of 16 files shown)


theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/ctas.html.twig
```diff
@@ -4,10 +4,19 @@
...
                value="{{ responseId }}">
         <input type="hidden" name="consent" value="yes"/>
+
+        {% if minimalConsent %}
+            {% set text_ok = 'consent_buttons_ok_minimal'|trans({'%arg1%': spName}) %}
+            {% set text_nok = 'consent_buttons_nok_minimal'|trans %}
+        {% else %}
+            {% set text_ok = 'consent_ok'|trans %}
+            {% set text_nok = 'consent_buttons_nok'|trans %}
+        {% endif %}
```
MKodde, line 14: :superhero:

theme/skeune/translations/messages.en.php
```diff
@@ -30,11 +30,13 @@
     'wayf_noaccess_email'       => 'Your emailaddress',
     'wayf_noaccess_motivation'  => 'Motivation',
     'wayf_noaccess_success'     => 'Your request for access has been sent.',
-    'wayf_eduId'                => 'If your organisation is not listed, <a href="%eduIdLink%" class="wayf__eduIdLink">eduID is available as an alternative.</a>',
+    'wayf_eduId'                => 'If your %organisation_noun% is not listed, <a href="%eduIdLink%" class="wayf__eduIdLink">eduID is available as an alternative.</a>',
```
MKodde, line 33: For our own purposes, this is a good setup. But other federations that might have an 'eduId' alternative wil have to update this translation. 

We can:

1. Leave it as is for now, think thats best for now. But discuss this with the PR
2. Remove the SURFism (eduID) and introduce a translatable (in translations override files) like we did with the organisation noun.
3. Remove surfisms by renamin eduId to something like defaultIdP.

For now, I'd say go with 1, but discuss this with the PM's

MKodde (approved):
Nice work again. See my concern in the comment below. Should not halt you from merging this tho. If we need to address this, it should be in a PR of its own.

theme/skeune/translations/messages.en.php
Badlapje: I went for 3.  It's a small change and a good remark.  It's easy enough to overwrite that way.

PR #951, opened 2020-10-25
Ensure focus styles match the styleboard

Badlapje:
After conferring with Edwin, changed the focus styles to these


Badlapje:
Dunno why the fixup isn't working, but other than that it's good to go.

PR #952, opened 2020-10-26
Style the IdP images

MKodde:
After messing up the original branch while upstream was rebased. This is a substitute PR for #947 

Also, after some discussion the SVG support for IdP images was dropped from this PR. It will take quite some time for the IdP team to convert all logos to SVG and only then would this change start making sense.


Badlapje (approved):
:shipit:

PR #955, opened 2020-10-29
Add keyboard & mouse tests for consent

Badlapje:
Refactored naming of spec files to be consistent.
Added keyboard tests for consent.
Added mouse tests for consent.
Seperated general tests for consent.
Fixed a bug which existed since refactoring the button styling for the nok-modal. Also fixed all cypress tests to use the .should('be.visible') test instead of the contains test (which also returns true if the element in question is hidden).


MKodde:
I've cherry-picked your changes to a new PR. That made more sense than rebasing this PR. Closing this in favour of #970

PR #956, opened 2020-10-29
Center items in the wayf vertically

Badlapje:


MKodde (commented):
Looked good in my functional review. See two formatting related items below.

PR #957, opened 2020-10-29
Change text of add account button

Badlapje:


MKodde (approved):
I see the PHP72 build sometimes breaks on the Cypress test suite. Glitch? Or is something wrong?

Badlapje:
> I see the PHP72 build sometimes breaks on the Cypress test suite. Glitch? Or is something wrong?

Seems like something cypress related we don't understand fully.  PHP version shouldn't have any effect on the Cypress tests.

PR #958, opened 2020-10-29
Change default isMinimalConsent behavior

MKodde:
By default when no idp/sp configuration is available the consent style
is set to be minimal.


Badlapje (approved):
Looking good 👍

PR #960, opened 2020-11-02
Select and Submit topmost IdP on enter

MKodde:
Select the first visible IdP from the list when the user presses the enter/return button.

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/WayfController.php
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/WayfController.php
@@ -22,6 +22,7 @@ use OpenConext\EngineBlockFunctionalTestingBundle\Helper\TestEntitySeeder;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Twig_Environment;
 
 /**
@@ -56,7 +57,7 @@ class WayfController extends Controller
         return new Response($this->twig->render(
             '@theme/Authentication/View/Proxy/wayf.html.twig',
             [
-                'action' => 'https://engine.vm.openconext.org/',
+                'action' => $this->generateUrl('functional_testing_handle_wayf'),
                 'greenHeader' => $currentLocale,
                 'helpLink' => '/authentication/idp/help-discover?lang='.$currentLocale,
                 'backLink' => $backLink,
@@ -71,5 +72,18 @@ class WayfController extends Controller
                     '<div id="request-access"></div></div></div>',
             ]
         ));
+    }
+
+    public function handleWayfAction(Request $request)
+    {
+        if ($request->request->has('idp')) {
+            return $this->redirectToRoute(
+                'open_conext_engine_block_authentication_homepage',
+                [
+                    'idp' => $request->request->get('idp')
+                ]
+            );
+        }
+        throw new AccessDeniedException('No IdP parameter found');
     }
 }
--- a/theme/base/javascripts/wayf/handleEnter.js
+++ b/theme/base/javascripts/wayf/handleEnter.js
@@ -3,6 +3,7 @@ import {fireClickEvent} from '../utility/fireClickEvent';
 import {handleDeleteDisable} from './handleDeleteDisable';
 import {submitForm} from './submitForm';
 import {handleIdpBanner} from './handleIdpBanner';
+import {selectFirstIdPAndSubmitForm} from "./selectFirstIdPAndSubmitForm";
 
 /**
  * Behaviour expected to happen after a user presses the enter button.
@@ -33,6 +34,8 @@ export const handleEnter = (e, previouslySelectedIdps) => {
       case 'idp__form':
       case 'idp__logo':
         submitForm(e, previouslySelectedIdps); break;
+      case 'search__field':
+        selectFirstIdPAndSubmitForm(previouslySelectedIdps); break;
       case 'wayf__defaultIdpLink':
         handleIdpBanner(e); break;
     }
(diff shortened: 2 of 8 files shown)


src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/WayfController.php
```diff
@@ -72,4 +72,17 @@ public function wayfAction(Request $request)
...
+    {
+        if ($request->request->has('idp')) {
+            return $this->redirectToRoute(
+                'open_conext_engine_block_authentication_homepage',
+                [
+                    'idp' => $request->request->get('idp')
+                ]
+            );
+        }
+        die('No IdP parameter found');
```
mroest, line 86: Are you sure you want die() instead of throwing an exception?

theme/base/javascripts/wayf/handleEnter.js
```diff
@@ -11,6 +12,7 @@ export const handleEnter = (e, previouslySelectedIdps) => {
   const classList = e.target.classList;
 
   classList.forEach(className => {
+    console.log(className);
```
mroest, line 15: Sure you want this to be here?

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Selects the first visible IdP from the wayf idpList and submits that IdP
+ *
+ * @param previouslySelectedIdps
+ */
+export const selectFirstIdPAndSubmitForm = (previouslySelectedIdps) => {
+  let idps = document.querySelectorAll('.wayf__idpList > li:first-child article');
+  // convert nodelist to array
+  idps = Array.prototype.slice.call(idps);
```
mroest, line 11: Why convert it. You can do a forEach() on the return value of querySelectorAll().
`Array.prototype.forEach.call(document.querySelectorAll())` or just loop it with an old-fashion style for loop.

Or if this is being transpiled you can do:

`for (let checkbox of document.querySelectorAll()) {}`

Badlapje: Actually: why use querySelectorAll?  This should be a querySelector as it should never return more than one item.  That should mean you can do:
```
const firstIdp = document.querySelector('.wayf__idpList > li:first-child > article');
// do something with firstIdp
```  

In case we do need to convert a nodelist to an array, there is already a helper function for that.

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
```diff
@@ -0,0 +1,27 @@
...
+  let selectedIdP = null;
+  // Select the first visible IdP from the list
+  idps.every((article) => {
+    // Select an item from the filtered list (weight should be > 0) or pick the
```
Badlapje, line 15: why are we checking weight here?  The search behaviour is well tested, no test it again as far as i can see.  Seems to me all we need to do is (going further on the above comment):
```
if (firstIdp) {
  selectAndSubmit(firstIdp, previouslySelectedIdps);
}
```

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
MKodde: I like the ability to break out of the sequence. Thats why I opted for the 'array' only `every` construction.

src/OpenConext/EngineBlockFunctionalTestingBundle/Controllers/WayfController.php
MKodde: Good point! Given this is test/debug code, I simply let the code die. But why not use the correct means..

theme/base/javascripts/wayf/handleEnter.js
MKodde: Thanks, removed that!

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
MKodde: When weight is `0`, the element is hidden from view (not a search result). Thats why the weight must be > 0 in order to make it a selectable option.

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
MKodde: Ah, and I need to select more than just the first item, as selecting only the visible IdP items was not something I was able to achieve using a simple css selector. Maybe you can help here?

mroest (approved):

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
Badlapje: that should be possible yes

Badlapje: `.wayf__remainingIdps .wayf__idpList > li:first-child > .wayf__idp`  should work.  I've just quickly tested it and that's fine.  

The weight check below shouldn't be necessary i believe, but you can do it using CSS like this: `.wayf__remainingIdps .wayf__idpList > li:first-child > .wayf__idp:not([data-weight="0"])`
I specifically set the value to 0 for the not clause, as when you hit enter without searching the items don't have a data-weight yet and so this selector works equally well when there has been no search.  

I confirmed it returns 0 items when the search returns "no items found"

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
Badlapje: Filtered out all possible hits with data-weight=0 in the css selector above.

theme/base/javascripts/wayf/selectFirstIdPAndSubmitForm.js
MKodde: Going to incorporate this into the PR!

Badlapje (approved):
:shipit:

PR #962, opened 2020-11-04
Repair the previously selected feature in test

MKodde:
The test WAYF did not function as intened. It did not show the previously selected IdP's as the cookie could not be used by the Twig Function:

1. The TWIG function compared the EnittyID of the IDP in a case sensitive manner (good thing) but the EntityId was lower cased in the Twig template (no bueno).
2. The cookie for previous selection was no longer indexed on entityID, making the matching logic in the Twig function impossible. This was also fixed by re-indexing the values once they where loaded in the function.


Badlapje (approved):
nice work!  Means my outstanding PR will need some serious rework to weed out everything you've already done in separate PRs (good thing really)

MKodde:
> nice work! Means my outstanding PR will need some serious rework to weed out everything you've already done in separate PRs (good thing really)

Sorry for that. :innocent:

PR #963, opened 2020-11-04
Increase whitespace between consent title and text

MKodde:
The whitespace needed to be a bit more roomy. By setting a margin-right on the title, this should not interfere with other viewports as they render title and text beneath each other.

### DIFF
--- a/theme/base/stylesheets/pages/consent.scss
+++ b/theme/base/stylesheets/pages/consent.scss
@@ -172,11 +172,13 @@
 
                 > .attribute__name {
                     @include grid-position(1, 2, 1, 2);
+                    margin-right: 0.5rem;
                     text-align: right;
 
                     @include screen('mobile') {
-                        text-align: left;
                         display: inline-block;
+                        text-align: left;
+                        margin-right: 0;
                     }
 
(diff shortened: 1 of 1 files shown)


theme/base/stylesheets/pages/consent.scss
```diff
@@ -172,6 +172,7 @@
 
                 > .attribute__name {
                     @include grid-position(1, 2, 1, 2);
+                    margin-right: 0.5rem;
```
Badlapje, line 175: I'd reset the margin-right to 0 for mobile (line 179) to be sure it doesn't add some weird layout breakage for some IDP attributes.  It's unlikely, but experience learns that it's better to prevent it alltogether.

Badlapje (changes requested):
Small change to prevent any possible issues.  Other than that looking good!

theme/base/stylesheets/pages/consent.scss
MKodde: Good suggestion, it did look good on smaller screens, but better safe then sorry!

Badlapje (approved):
:shipit:

PR #964, opened 2020-11-04
Set other consent icons at privacy and sec section

MKodde:
The new icons where exported from figma and are a round grey badge with kind of the same logo in them. Additionally a drop shadow is part of the SVG. 

The drop shadow looks best when the SVG is rendered a bit larger. I considered removing the SVG drop shadow and let CSS render a box shadow. But this was an idiot idea. As the box part of the box layout caused a square shadow around the round logo..

Screenshot_2020-11-04 OpenConext - This information will be shared with DisplayName


theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/SVG/shield_key.svg
```diff
@@ -1,5 +1,23 @@
...
+<svg width="50" height="50" focusable="false" viewBox="0 0 50 50" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Badlapje, line 1: The svg should need an empty title element to indicate the alt text.  Same for the other one.  Then on the outside svg element add a aria-labelledby attribute containing the id of the title tag.  

`<title id="some_id_for_the_svg"></title>`

Badlapje (changes requested):
A small accessibility tweak.  Other than that it looks good.

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/SVG/shield_key.svg
MKodde: I tried this, but to no avail. The HTML check in the A11y spec complains about `no-missing-referencesElement references missing id "fifo-title"`. 

Why is it important to have this aria-labelledby solution in place? Wouldn't setting a simple empty label/title on the svg not have the same effect?

Badlapje: A label wouldn't, but an empty title by itself should suffice.  Some screenreaders however aren't build sturdy enough and require some more nuts and bolts.  If a11y spec keeps complaining about it though, then best to do it without the id/aria-labelledby.  I guess the automatic inspection has caught up with bad implementation from screenreaders now.

MKodde: Will give your other suggestion a try. 

But maybe we should not comply to a poor inspection when the solution you proposed is the actual best practice.. It's up to you to decide. For now I'm going to set the empty title.

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/SVG/file_info.svg
```diff
@@ -1,4 +1,4 @@
-<svg width="50" height="50" focusable="false" viewBox="0 0 50 50" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg title="" width="50" height="50" focusable="false" viewBox="0 0 50 50" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Badlapje, line 1: this shouldn't be an attribute but a title element.

so <title></title> as the first child element within the svg-element

Badlapje (changes requested):
Other than the title element looking good :-)

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/SVG/file_info.svg
MKodde: Driemaal is scheepsrecht

PR #966, opened 2020-11-04
Fix undefined index error

tvdijen:
Oct 30 19:14:34 sv2010010 EBLOG-FIN[21532]: [2020-10-30 19:14:34] app.ERROR: Undefined index: idp [/apps/installation/OpenConext-engineblock/engine.broker.fin.rijksdienst.nl/OpenConext-engineblock-6.3.6/library/EngineBlock/Corto/Module/Service/ContinueToIdp.php:67]

This should never happen in the first place, but it could happen if a session is lost.. The undefined index error can be prevented

### DIFF
--- a/library/EngineBlock/Corto/Module/Service/ContinueToIdp.php
+++ b/library/EngineBlock/Corto/Module/Service/ContinueToIdp.php
@@ -64,15 +64,19 @@ class EngineBlock_Corto_Module_Service_ContinueToIdp implements EngineBlock_Cort
      */
     public function serve($serviceName, Request $httpRequest)
     {
-        $selectedIdp = urldecode($_REQUEST['idp']);
-        if (!$selectedIdp) {
+        $selectedIdp = null;
+        if (array_key_exists('idp', $_REQUEST)) {
+            $selectedIdp = urldecode($_REQUEST['idp']);
+        }
+
+        if ($selectedIdp !== null) {
             throw new EngineBlock_Corto_Module_Services_Exception(
                 'No IdP selected after WAYF'
             );
         }
 
         // Retrieve the request from the session.
-        $id      = $_POST['ID'];
+        $id = $_POST['ID'];
         if (!$id) {
             throw new EngineBlock_Exception(
                 'Missing ID for AuthnRequest after WAYF',
--- a/library/EngineBlock/Corto/Module/Service/ContinueToIdp.php
+++ b/library/EngineBlock/Corto/Module/Service/ContinueToIdp.php
@@ -69,7 +69,7 @@ class EngineBlock_Corto_Module_Service_ContinueToIdp implements EngineBlock_Cort
             $selectedIdp = urldecode($_REQUEST['idp']);
         }
 
-        if ($selectedIdp !== null) {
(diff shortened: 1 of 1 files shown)


thijskh:
Thanks! The code change looks good to me. The build fails (seems unrelated, but I'd prefer that we have a passing build when merging things to master)

library/EngineBlock/Corto/Module/Service/ContinueToIdp.php
```diff
@@ -64,15 +64,19 @@ public function __construct(
...
     public function serve($serviceName, Request $httpRequest)
     {
-        $selectedIdp = urldecode($_REQUEST['idp']);
-        if (!$selectedIdp) {
+        $selectedIdp = null;
+        if (array_key_exists('idp', $_REQUEST)) {
+            $selectedIdp = urldecode($_REQUEST['idp']);
+        }
+
+        if (!isset($selectedIdp)) {
```
MKodde, line 72: I have one small nitpick. And thats regarding the use of `isset`. I think using the old construction of `if (!$selectedIdP) {` makes more sense semantically.  But feel free to leave as is!

Thanks @tvdijen, I love these defense in depth kind of contributions!

MKodde (commented):
@thijskh we are running into the Composer 2 incompatibility here. 

What do you prefer?

1. Downgrade to composer 1.
2. Provide composer 2 compatibility.

I recon 1 is the most pragmatic an least time consuming option at this point..

library/EngineBlock/Corto/Module/Service/ContinueToIdp.php
tvdijen: But that's a loose comparision and we shouldn't want that...  `!== null` be better then?

MKodde: Agreed :+1:  `!== null` or even using the `is_null` function sounds like a plan. Again, your proposal is fine with me. Just venting my personal preferences here.

tvdijen: I took care of it..

thijskh:
1 seems good

thijskh:
@tvdijen Build should be fixed in master, can you rebase so we’re sure the build passes?

library/EngineBlock/Corto/Module/Service/ContinueToIdp.php
```diff
@@ -64,15 +64,19 @@ public function __construct(
...
     public function serve($serviceName, Request $httpRequest)
     {
-        $selectedIdp = urldecode($_REQUEST['idp']);
-        if (!$selectedIdp) {
+        $selectedIdp = null;
+        if (array_key_exists('idp', $_REQUEST)) {
+            $selectedIdp = urldecode($_REQUEST['idp']);
+        }
+
+        if ($selectedIdp !== null) {
```
thijskh, line 72: Probably should be `=== null`?

thijskh (changes requested):

thijskh (approved):

PR #969, opened 2020-11-10
Leave the previously selected IdPs in the 'select another IdP' list

MKodde:
In the previous situation the previously selected IdPs where only shown in the shortlist. This is not very intuitive when clicking the 'select a different IdP' link from the short list. As the IdPs you used before are no longer available for selection.


Badlapje (approved):
Brilliantly simple code-change :-)

MKodde:
Merging this despite the dailing cypress tests, will fix those on my next PR

PR #973, opened 2020-11-17
Show search icon on page load

MKodde:
And hide the reset button visually. When typing, show the reset button (remove the class).

The background has also been removed from the poirot element. This prevents setting extra logic to color with the parent elements focus style. E.g. when the search bar is focussed, you want the poirot to be focussed too

### DIFF
--- a/theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/search.html.twig
+++ b/theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/search.html.twig
@@ -10,6 +10,6 @@
         placeholder="{{ 'wayf_search_placeholder'|trans }}"
         type="search"
     />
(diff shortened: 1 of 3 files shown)


theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/search.html.twig
```diff
@@ -10,6 +10,6 @@
         placeholder="{{ 'wayf_search_placeholder'|trans }}"
         type="search"
     />
+    <button type="reset" class="search__reset visually-hidden"><span class="visually-hidden">{{ 'reset'|trans }}</span></button>
```
MKodde, line 13: To prevent the images from overlapping each other, the reset button is visually hidden. This class is removed once it is triggered for display (`toggleSearchAndResetButton` function)

Badlapje (approved):
💯  nice work!

PR #974, opened 2020-11-17
Only show default IdP when available

MKodde:
Only show default IdP when available and when the feature is enabled. Testing this feature showed an error in the way this feature was created. The di container checked for the existence of a config setting, and not for the existence of a the wayf parameter. This was also fixed in this commit


Badlapje (approved):
Smooth 👍   🚀

PR #970, opened 2020-11-11
Repair and upgrade cypress tests

MKodde:
The aim of this PR is to make the cypress tests more robust. And to address test issues that did previously not trigger an error in the CI tests. 

Replaces: https://github.com/OpenConext/OpenConext-engineblock/pull/955

PR #975, opened 2020-11-18
Validate required fields

MKodde:
Adding rudimentary validation to the request access form

~~Tests are not provided to prevent issues with #970~~ 

~~When that PR is merged, tests can be added to this PR.~~ Added them in https://github.com/OpenConext/OpenConext-engineblock/pull/975/commits/ccec1c4019e03861c8d6cc0181847d24e325a2ce

### DIFF
--- a/theme/base/javascripts/wayf/handleEnter.js
+++ b/theme/base/javascripts/wayf/handleEnter.js
@@ -4,7 +4,6 @@ import {handleDeleteDisable} from './handleDeleteDisable';
 import {submitForm} from './submitForm';
 import {handleIdpBanner} from './handleIdpBanner';
 import {selectFirstIdPAndSubmitForm} from "./selectFirstIdPAndSubmitForm";
-import {toggleFormFieldsAndButton} from './noAccess/toggleFormFieldsAndButton';
 
 /**
  * Behaviour expected to happen after a user presses the enter button.
@@ -39,8 +38,6 @@ export const handleEnter = (e, previouslySelectedIdps) => {
         selectFirstIdPAndSubmitForm(previouslySelectedIdps); break;
       case 'wayf__defaultIdpLink':
         handleIdpBanner(e); break;
-      case 'cta__showForm':
-        toggleFormFieldsAndButton(); break;
     }
   });
 };
--- a/theme/base/javascripts/wayf/noAccess/validation.js
+++ b/theme/base/javascripts/wayf/noAccess/validation.js
@@ -1,0 +1,38 @@
+import {nodeListToArray} from "../../utility/nodeListToArray";
+
+/**
+ * Verify the name and email fields are not empty. Due to the way the validation was set up we can
+ * not utilize the 'required' attribute on the form input fields. So a JS validation is required
+ * to prevent empty form submits.
+ *
+ * @param formData
+ * @returns {boolean}
+ */
+export const valid = (formData) => {
+  // Start with hiding previous validation messages, they will reappear if the error persists.
+  hideValidationMessages();
+  const nameValid = notAnEmptyField(formData, 'name');
+  const emailValid = notAnEmptyField(formData, 'email');
+  // Next, test the name and email fields for validity, set error message if invalid.
+  return nameValid && emailValid;
+};
+
+function notAnEmptyField(formData, elementName) {
+  if (!formData.has(elementName) || formData.get(elementName).trim() === '') {
+    showValidationMessage(elementName);
+    return false;
+  }
+  return true;
+}
+
+function hideValidationMessages(){
+  const errorMessages = nodeListToArray(document.getElementsByClassName('form__error'));
+  errorMessages.forEach(errorMessage => {
+    errorMessage.classList.add('hidden');
+  });
+}
+
+function showValidationMessage(elementName){
+  const errorMessage = document.querySelector(`p[data-labelfor="${elementName}"]`);
+  errorMessage.classList.remove('hidden');
+}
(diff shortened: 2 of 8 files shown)


MKodde:
@Badlapje When you've got a lost moment, would you be so kind to look into this strange behavior? 

The failing Cypress tests seem to break on the malfunctioning `Request access` button. This button no longer reacts to `enter` presses. I debugged this to some extent and think this has to do with the construction below. 

https://github.com/OpenConext/OpenConext-engineblock/blob/21b687c59f5cf3f240ce1ef0ffc2fb8609d67098/theme/base/javascripts/wayf/noAccess/attachClickHandlerToRequestButton.js#L17-L32

Do you recal how/why this construction was created? And do you have a clue on how to fix this? I'm hitting a wal here.

Badlapje:
> 
> 
> @Badlapje When you've got a lost moment, would you be so kind to look into this strange behavior?
> 
> The failing Cypress tests seem to break on the malfunctioning `Request access` button. This button no longer reacts to `enter` presses. I debugged this to some extent and think this has to do with the construction below.
> 
> https://github.com/OpenConext/OpenConext-engineblock/blob/21b687c59f5cf3f240ce1ef0ffc2fb8609d67098/theme/base/javascripts/wayf/noAccess/attachClickHandlerToRequestButton.js#L17-L32
> 
> Do you recal how/why this construction was created? And do you have a clue on how to fix this? I'm stuck.

It was created because else the clickhandler is added more than once.  With that construction it should only be added once.

I'll take a look in an hour or so to see if i can fix it.

theme/base/javascripts/wayf/noAccess/validation.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param formData
+ * @returns {boolean}
+ */
+export const valid = (formData) => {
+  // Start with hiding previous validation messages, they will reappear if the error persists.
+  hideValidationMessages();
+  // Next, test the name and email fields for validity, set error message if invalid.
+  let valid = true;
+  if (!formData.has('name') || formData.get('name').trim() === '') {
```
Badlapje, line 16: this could be extracted into a function called something like `notAnEmptyField(formData, formName)`.

Would improve readability & make code a bit DRYer as line 20 is the exact same, save for the name of the field.

Badlapje: could even make it into a method called `validateFieldNotEmpty(formData, formName)` and have it return a boolean.  

`valid = validateFieldNotEmpty(...);`

theme/base/javascripts/wayf/noAccess/validation.js
```diff
@@ -0,0 +1,39 @@
...
+
+function hideValidationMessages(){
+  const errorMessages = nodeListToArray(document.getElementsByClassName('form__error'));
+  errorMessages.forEach(message => {
+    message.style.display = 'none';
+  });
+}
+
+function showValidationMessage(elementName){
+  const errorMessage = document.querySelector(`input[name="${elementName}"]`).parentElement.getElementsByClassName('form__error')[0];
```
Badlapje, line 35: you could simplify this query by adding a data-forField (name up to you) to the html of the form__error element.  That'd allow you to select it directly rather than going the long way round.

Probably not that important, but would be a bit more performant.

theme/base/javascripts/wayf/noAccess/validation.js
```diff
@@ -0,0 +1,39 @@
...
+  if (errorMessage.style.display !== undefined && errorMessage.style.display !== 'block') {
+    errorMessage.style.display = 'block';
```
Badlapje, line 37: You could show and hide it by adding / removing the hidden attribute.  Not that important, but i try to avoid inline css like this whenever possible (same for the hideValidationMessage function).

Badlapje (changes requested):
I'd make some smaller changes.  Otherwise nice work!  

If you disagree, feel free to merge ;-)  Nothing big in what I'd change.

theme/base/javascripts/wayf/noAccess/validation.js
MKodde: Great suggestion! Going to incorporate this right away.

theme/base/javascripts/wayf/noAccess/validation.js
MKodde: Good point, fixed that up

theme/base/javascripts/wayf/noAccess/validation.js
MKodde: Thanks for the suggestion, reworked this to make this work with the hidden class.

theme/base/javascripts/wayf/handleEnter.js
```diff
@@ -39,8 +38,6 @@ export const handleEnter = (e, previouslySelectedIdps) => {
         selectFirstIdPAndSubmitForm(previouslySelectedIdps); break;
       case 'wayf__defaultIdpLink':
         handleIdpBanner(e); break;
-      case 'cta__showForm':
-        toggleFormFieldsAndButton(); break;
```
MKodde, line 43: :wastebasket: FTW!

Badlapje (approved):
Looking great, nice work!

MKodde:
@Badlapje I suspect removing the handler did cause a behavioral change. As demonstrated by the failing WAYF keyboard tests. Could you verify that, and maybe fix it :grin:

MKodde:
Lets move ahead with this!

PR #977, opened 2020-11-23
Only show default idp when connected

MKodde:
Only show the defautl IdP when it is connected. This makes more sense from an user perspective. 
And we now show the actual IdP title instead of always rendering eduID as the default IdP name.


Badlapje (approved):
Nice work! :shipit:

PR #978, opened 2020-11-23
Do not show delete button on the noAccess screen

Badlapje:


MKodde:
Code wise there is nothing wrong with this i guess. Would it be possible to add a test for this change?

MKodde (approved):

Badlapje:
Didn't add a test for this change as it's impossible to test in the functional testing endpoint without spending too much time on modifying that endpoint for this case.

PR #979, opened 2020-11-23
Ensure logo height never grows too big

Badlapje:
You can test this by using the following code to replace line 14 in idp.html.twig & and using the logo that triggered the bug report conext_logo
 (or something similar):

```
           {% if loop.index == 1 %}
                {% set logoUrl = '/images/temp/conext_logo.png' %}
            {% else %}
                {% set logoUrl = idp.logo %}
            {% endif %}
```

### DIFF
--- a/theme/base/stylesheets/pages/wayf.scss
+++ b/theme/base/stylesheets/pages/wayf.scss
@@ -92,6 +92,7 @@
                         > svg {
                             height: auto;
                             margin: auto 0;
+                            max-height: 100%;
                             width: 100%;
                         }
                     }


MKodde:
Thanks for the great reproduction report!

I was able to toggle the change in the inspector. And it now seems the logo is compressed. Changing the correct aspect ratio. 

Is that the way to go? Or should we scale down the entire image. Or even crop it. Maybe a good thing to discuss with PM's.

Badlapje:
There's really no alternative.  The image should scale 99% of the time.  The 1% where it doesn't because of something weird in the image, it won't break out of the container div, but that's as good as it gets really.  I can look into it more, but i don't know of any technique that could circumvent this.

MKodde:
I agree! Let's merge this PR!

PR #980, opened 2020-11-24
Use organization name for SP in disclaimer text

MKodde:
The following fallback sequence was implemented.

Show:
1. Organization Display Name
2. Organization Name when 1 is not available
3. 'Unknown' when 1 and 2 are not available

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
@@ -364,6 +364,32 @@ class ServiceProvider extends AbstractRole
         return $spName;
     }
 
+    public function getOrganizationName(string $preferredLocale = ''): string
+    {
+        $orgName = '';
+        if ($preferredLocale === 'nl') {
+            $orgName = $this->organizationNl->displayName;
+            if (empty($orgName)) {
+                $orgName = $this->organizationNl->name;
+            }
+        } elseif ($preferredLocale === 'en') {
+            $orgName = $this->organizationEn->displayName;
+            if (empty($orgName)) {
+                $orgName = $this->organizationEn->name;
+            }
+        } elseif ($preferredLocale === 'pt') {
+            $orgName = $this->organizationPt->displayName;
+            if (empty($orgName)) {
+                $orgName = $this->organizationPt->name;
+            }
+        }
+
+        if (empty($orgName)) {
+            $orgName = '';
+        }
+        return $orgName;
+    }
+
     /**
      * @return bool
      */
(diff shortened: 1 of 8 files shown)


src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
```diff
@@ -364,6 +364,23 @@ public function getDisplayName($preferredLocale = '')
...
+        } elseif ($preferredLocale === 'en') {
+            $orgName = $this->organizationEn->displayName;
+        } elseif ($preferredLocale === 'pt') {
+            $orgName = $this->organizationPt->displayName;
+        }
+
+        if (empty($orgName)) {
+            $orgName = $this->getDisplayName($preferredLocale);
+        }
+        return $orgName;
```
Badlapje, line 381: is the "unknown" fallback not missing here?

src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
```diff
@@ -364,6 +364,23 @@ public function getDisplayName($preferredLocale = '')
         return $spName;
     }
 
+    public function getOrganizationName(string $preferredLocale = ''): string
+    {
+        $orgName = '';
+        if ($preferredLocale === 'nl') {
```
Badlapje, line 370: You could get rid of the first if by making dutch the default locale and overwriting orgName if the preferred locale is en or pt?

I also think using a switch would be better for extendability / readability.  Even better would be to use dynamic method names with something like this (going from memory here, not sure if the () after $locale on the second line is necessary)

```
$locale = "organization${ucFirst($preferredLocale)}";
$orgName = $this->$locale()->displayName;
```

Badlapje (changes requested):
I seem to be missing the "unknown" fallback?

src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
MKodde: I'm catching that in the template. That way we can prevent translations ending up in the code. Alternatively I could return the translation key here. But think this is the most user friendly solution for now.

src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
MKodde: I'm aware of the possibilities PHP provides in making this method more condensed. And I'm tempted to rewrite the method. 

However, for now I'd like to keep this method as it is, given it is very similar to the 'getDisplayName' method. Who is structured similarly.

Badlapje: sounds good 👍

src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
Badlapje: 💯

Badlapje (approved):

PR #981, opened 2020-11-24
Isolate current consent content

Badlapje:
Making this a micro commit to avoid merge conflicts down the road.

To be clear: 
- moved the h2 in the consent__content div
- ensured that all h2's one level down from the consent container will have the same styling
- added a consent__content-body div
- moved styling from consent__content to consent__content-body

### DIFF
--- a/theme/base/stylesheets/pages/consent.scss
+++ b/theme/base/stylesheets/pages/consent.scss
@@ -24,7 +24,7 @@
         }
     }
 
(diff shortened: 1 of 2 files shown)


MKodde (approved):
Thanks for the re-organization. I did not test drive this change, but assume you did. Any (if any) regressions can be caught in our upcoming Consent effort.

:ship: it

PR #983, opened 2020-11-24
Consent tooltip changes

MKodde:
Update the imagery:
* The closing cross was rid of its circle
* The questionmark was exchanged for an exclamation mark

Update the tooltip value presentation:
* Align the text on the left
* Widen the tooltip text area to 3/4th of the available space, effectively aligning it directly beneath the attribute value


MKodde:
@Badlapje would you be so kind to look into making the hit surface for the i button bigger? I can increase the padding but that makes the icon to go out of alignment. While at it, maybe you can check out why the icon is mis-aligned in mobile mode. :angel:

Badlapje (approved):
Looking good 💯  :shipit:

PR #999, opened 2020-12-03
Localize the sort

Badlapje:
Add a language/locale parameter so sort uses a localized comparison

### DIFF
--- a/theme/base/javascripts/wayf/utility/sortIdpMethods.js
+++ b/theme/base/javascripts/wayf/utility/sortIdpMethods.js
@@ -12,8 +12,9 @@ import {getData} from '../../utility/getData';
 export const sortByTitle = (firstElement, secondElement) => {
   const titleOne = getTitle(firstElement.children[0]);
   const titleTwo = getTitle(secondElement.children[0]);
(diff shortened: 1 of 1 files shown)


theme/base/javascripts/wayf/utility/sortIdpMethods.js
```diff
@@ -12,8 +12,9 @@ import {getData} from '../../utility/getData';
 export const sortByTitle = (firstElement, secondElement) => {
   const titleOne = getTitle(firstElement.children[0]);
   const titleTwo = getTitle(secondElement.children[0]);
+  const lang = document.querySelector('html').getAttribute('lang') || 'nl';
```
mroest, line 15: Usually the locale is nl_NL or en_GB. Are you sure the default is nl?

mroest (approved):

theme/base/javascripts/wayf/utility/sortIdpMethods.js
mroest: Never mind. I'v checked the params and the locales are nl,en etc...

PR #1022, opened 2020-12-11
Use english name as backup of locale-specific name

Badlapje:
The algorithm for specifying a name is now:
If OrgDisplayName:TAAL is set: OrgDisplayname:TAAL;
Elseif OrgName:TAAL is set: OrgName:TAAL
Elseif OrgDisplayName:en is set: OrgDisplayname:en
Elseif OrgName:en is set: OrgName:en
Else "Onbekend"

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
@@ -367,15 +367,11 @@ class ServiceProvider extends AbstractRole
     public function getOrganizationName(string $preferredLocale = ''): string
     {
         $orgName = '';
+
         if ($preferredLocale === 'nl') {
             $orgName = $this->organizationNl->displayName;
             if (empty($orgName)) {
                 $orgName = $this->organizationNl->name;
-            }
-        } elseif ($preferredLocale === 'en') {
-            $orgName = $this->organizationEn->displayName;
-            if (empty($orgName)) {
-                $orgName = $this->organizationEn->name;
             }
         } elseif ($preferredLocale === 'pt') {
             $orgName = $this->organizationPt->displayName;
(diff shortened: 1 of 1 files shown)


src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
```diff
@@ -367,26 +367,30 @@ public function getDisplayName($preferredLocale = '')
...
+            if (empty($orgName)) {
+                $orgName = $this->organizationEn->name;
+            }
+        }
+
         if (empty($orgName)) {
             $orgName = '';
         }
+
         return $orgName;
```
mroest, line 394: Als ik deze functie/code zie dan heb ik toch wat mixed feelings. Er wordt 3 keer op dezelfde manier een variable opgehaald. Dat zou toch wel wat 'netter' kunnen en dit hele ding aanzienlijk versimpelen.


`
function _getOrgname($preferredLocale, $default = 'en') {
$preferred = 'organisation' . ucfirst($preferredLocale);
$fallback = 'organisation' . ucfirst($default);
return !empty($this->$varName->name) ? $this->$varName->name : $this->$fallback->name;
}
`
En dat kan dan nog vast iets netter :)

mroest (approved):

src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
Badlapje: I agree completely & made [a similar remark when this was changed to the previous algorithm](https://github.com/OpenConext/OpenConext-engineblock/pull/980#discussion_r529379926).  However, given that the rest of the code is organized the same way we chose consistency of code over DRY/ness.  

Given that i remained in the same code style for this PR.

PR #1029, opened 2020-12-15
Ensure links don't overflow on mobile

Badlapje:
On the index page, links don't break nicely on mobile.  This pr resolves that in a generic way


MKodde:
Watch out for rabbit holes here. And please consult with product owners if we should address these points.

home-page-feedback

Other then that this is a perfectly elegant change and improvement.

MKodde (approved):
Ship that sucker!

PR #1030, opened 2020-12-15
Ensure informational is big enough for icon

Badlapje:
On tablets the design sometimes uses elements of the mobile design.  For the informational that means the icon isn't shown entirely as the padding is the mobile one and the text is single-line.  

This pr solves that by adding a min-height to the informational.


MKodde:
I was not able to see the previous buggy behaviour in my Firefox browser. Tried different viewports for mobile and tablets. With and without the line-height. But I assume this issue might arise when the default IdP text is longer/shorter this will become problematic. Nice to see a min-height is set to make sure the icon is always shown correctly.

MKodde (approved):

PR #1031, opened 2020-12-15
Align idpRow label to the left on mobile

Badlapje:


MKodde (changes requested):
@Badlapje can you provide some steps for me to reproduce this? 

What I did find out was that pressing the default idp link in the banner caused a JS error. Possibly because the default IdP is not listed?

The changes seems fine to me, but I'd like to verify it in my own dev-vm

Badlapje:
It sounds like you're looking in the wrong place.  This change is for the consent screen, the default Idp is for the WAYF.

Steps to reproduce:
- open consent page
- go to mobile view (phone)
- look at the alignment of the "something incorrect" label.  Atm, it looks like this:
image
After this PR it should look like this:
image

MKodde:
Ah, I must-have had something sticking in my eye.

MKodde:
Thanks for the steps to reproduce! I fully agree with this change.

While keeb-testing. I ran into a small inconsistance that I was unable to screenshot, but reproduced it in Photoshop. See screenshot below. The hover state on a hidden (x) button might have to be removed. But I think this is for another PR..

hidden-close-button-gets-hover-state

MKodde (approved):

PR #1032, opened 2020-12-15
Implement optimizations found in lighthouse/webhint audits

Badlapje:
DONE:
- preload critical fonts
- use the charset meta-tag and make sure it's first
- add noreferrer/noopener to all _blank links
- minify css for better performance

TODO, added to [this ticket ](https://www.pivotaltracker.com/n/projects/1425900/stories/174968005)in pivotal:  
- correct wrong content-types
- add missing content-types
- check cache control headers
- ensure text-based resources are compressed
- add missing x-content-type headers

### DIFF
--- a/theme/base/templates/layouts/scripts/default.html.twig
+++ b/theme/base/templates/layouts/scripts/default.html.twig
@@ -2,7 +2,7 @@
 <html lang="{{ locale()|escape('html_attr') }}">
 <head>
     <title>{% block title %}{{ defaultTitle }}{% endblock %}</title>
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+    <meta charset="utf-8">
     <meta name="robots" content="noindex, nofollow"/>
     <meta content="IE=edge" http-equiv="x-ua-compatible">
     <meta content="initial-scale=1.0,width=device-width" name="viewport">
--- a/theme/base/templates/layouts/scripts/default.html.twig
+++ b/theme/base/templates/layouts/scripts/default.html.twig
@@ -10,6 +10,10 @@
     <meta content="translucent-black" name="apple-mobile-web-app-status-bar-style">
     <link href="/favicon.ico" rel="shortcut icon">
     {% block stylesheets %}{% spaceless %}
+        <link rel="preload" href="/fonts/sourcesanspro-semibold-webfont.woff2" as="font" type="font/woff2" crossorigin>
+        <link rel="preload" href="/fonts/sourcesanspro-bold-webfont.woff2" as="font" type="font/woff2" crossorigin>
+        <link rel="preload" href="/fonts/nunito-semibold.woff2" as="font" type="font/woff2" crossorigin>
+        <link rel="preload" href="/fonts/sourcesanspro-regular-webfont.woff2" as="font" type="font/woff2" crossorigin>
         <link href="/stylesheets/application.css?v={{ assetsVersion }}" rel="stylesheet" type="text/css"/>
     {% endspaceless %}{% endblock %}
 
--- a/theme/base/templates/layouts/scripts/default.html.twig
+++ b/theme/base/templates/layouts/scripts/default.html.twig
@@ -1,13 +1,13 @@
 <!DOCTYPE html>
 <html lang="{{ locale()|escape('html_attr') }}">
 <head>
-    <title>{% block title %}{{ defaultTitle }}{% endblock %}</title>
     <meta charset="utf-8">
     <meta name="robots" content="noindex, nofollow"/>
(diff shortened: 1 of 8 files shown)


theme/base/templates/layouts/scripts/default.html.twig
```diff
@@ -1,15 +1,19 @@
...
 <head>
-    <title>{% block title %}{{ defaultTitle }}{% endblock %}</title>
-    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+    <meta charset="utf-8">
     <meta name="robots" content="noindex, nofollow"/>
     <meta content="IE=edge" http-equiv="x-ua-compatible">
     <meta content="initial-scale=1.0,width=device-width" name="viewport">
     <meta content="yes" name="apple-mobile-web-app-capable">
     <meta content="translucent-black" name="apple-mobile-web-app-status-bar-style">
+    <title>{% block title %}{{ defaultTitle }}{% endblock %}</title>
```
MKodde, line 10: Funny to see that the move of the title element makes a difference from a optimization standpoint.

theme/base/templates/layouts/scripts/default.html.twig
```diff
@@ -1,15 +1,19 @@
...
     <link href="/favicon.ico" rel="shortcut icon">
     {% block stylesheets %}{% spaceless %}
+        <link rel="preload" href="/fonts/sourcesanspro-semibold-webfont.woff2" as="font" type="font/woff2" crossorigin>
+        <link rel="preload" href="/fonts/sourcesanspro-bold-webfont.woff2" as="font" type="font/woff2" crossorigin>
+        <link rel="preload" href="/fonts/nunito-semibold.woff2" as="font" type="font/woff2" crossorigin>
+        <link rel="preload" href="/fonts/sourcesanspro-regular-webfont.woff2" as="font" type="font/woff2" crossorigin>
         <link href="/stylesheets/application.css?v={{ assetsVersion }}" rel="stylesheet" type="text/css"/>
```
MKodde, line 17: Why is the application stylesheet on also preloaded? I'm a noob at this discipline, but it seems to make sense to preload all resources?

theme/base/templates/modules/Authentication/View/Index/index.html.twig
```diff
@@ -176,7 +176,7 @@
 
                         {% if helpLink is defined %}
                             <li>
-                                <a href="{{ helpLink }}" target="_blank">{{ 'help'|trans|capitalize }}</a>
+                                <a href="{{ helpLink }}" target="_blank" rel="noreferrer noopener">{{ 'help'|trans|capitalize }}</a>
```
MKodde, line 179: Great defence in depth addition. Going to try and incorporate this in my daily toolchain!

Have you even seen the proof of concept where a guy played some sort of pong game between two browser windows, based on the window.opener feature? As always I'm unable to find the link in my browser history..

theme/postcss.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  plugins: [
+    require('cssnano')({
+      preset: 'default',
+    }),
+  ],
+};
```
MKodde, line 7: This was required to get a better described css output?

MKodde (approved):

MKodde:
I trust you added the remaining findings to another ticket? Maybe mention a link to that ticket in the comments, or PR description.

theme/base/templates/layouts/scripts/default.html.twig
Badlapje: It was more about the meta-charset tag to be first, probably so it knows what charset it should expect which means skipping some if/elses i presume

theme/base/templates/layouts/scripts/default.html.twig
Badlapje: The stylesheet doesn't need to be preloaded, as it's in the head.  It'll be loaded before the content does.  However, browsers optimize on their own, which means that they only load fonts (in a stylesheet or in a link) after all content is loaded & rendered.

By using a link with preload we tell them that for those 4 fonts, it's critical to load them before the content rather than after.

theme/base/templates/modules/Authentication/View/Index/index.html.twig
Badlapje: i haven't but that sounds very interesting!

theme/postcss.config.js
Badlapje: yes, at least locally the cssnano wasn't applied without it.  Looking through the docs as to why it wasn't applied i noticed they mentioned that you need this file to make it work.  Worked like a charm as soon as i put it there :-)

PR #1033, opened 2020-12-16
Increase room for tooltipless attribute values

Badlapje:
The idea here is to allow enough room for at least a 40-character attribute value, both on all screen sizes.  The reason being that we want to ensure that the identifier fits 95% of the time.  Because identifiers tend to be persistent nameids, which are a sh1-hash, that means 40 characters.

### DIFF
--- a/theme/base/stylesheets/pages/consent/attribute.scss
+++ b/theme/base/stylesheets/pages/consent/attribute.scss
@@ -7,6 +7,14 @@
     border-bottom: 1px solid $softBorderGray;
     line-height: 1.63;
 
+    &.consent__attribute--noTooltip {
+        @include grid-template-columns(33% 1% 65% 1%);
+
+        @include screen('mobile') {
+            @include grid-template-columns(100%);
+        }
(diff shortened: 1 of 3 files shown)


MKodde:
All changes seem to make sense. I however was unable to see the NameId correctly displayed. All i got was 'Array' as the nameId value released by openconext.

See screenshot below. I think the actual value needs to be extracted from the NameId value object in the Twig Template.

Screenshot_2020-12-17 OpenConext - Do you give consent to share your information

Badlapje:
> All changes seem to make sense. I however was unable to see the NameId correctly displayed. All i got was 'Array' as the nameId value released by openconext.
> 
> See screenshot below. I think the actual value needs to be extracted from the NameId value object in the Twig Template.

This seems to be something in the test controller as it doesn't happen on test2 and the one above is perfectly fine.  Added to the leftover ticket with a link here.

MKodde (approved):
Thanks for adding the task to review the test controller

PR #1034, opened 2020-12-16
Allow html in the warning text

Badlapje:


theme/base/templates/modules/Default/Partials/warning.html.twig
```diff
@@ -2,5 +2,5 @@
     {% if notificationTitle is defined %}
         <h3>{{ notificationTitle }}</h3>
     {% endif %}
-    {{ text }}
+    {{ text|raw }}
```
MKodde, line 5: Html for consent settings is not specified in a translation file. But hails from Manage. Where a SP/IdP combination can be set with a personalized notification text.

As this is data coming from outside EngineBlock, I would like to see a more strict allowance of styling attirbutes. Using RAW filtering for example also allows the inclusion of style or script tags. Leading to a very remote possibility to create xss and other types of exploits.

Maybe using the `filter` with an agreed upon number of html elements is a bit safer here. I think of something like: `filter:a,br,p,strong,italic,ul,li,` and so forth. But not allowing style/script/canvas/iframe and other monstrocities.

Maybe I'm a bit to strict in not trusting Manage. But better safe than sorry in my opinion. 

We should discuss with PM's on what styling is considered safe, and what not.

MKodde (changes requested):

theme/base/templates/modules/Default/Partials/warning.html.twig
MKodde: Also see some of the history of this feature:
`https://www.pivotaltracker.com/n/projects/1425900/stories/158145547`
`https://github.com/OpenConext/OpenConext-engineblock/pull/544`

Badlapje: Interesting.  Not sure yet how to filter that out.  Would be easier to block some things i reckon, but will find out :-)

I can discuss with Thijs/Arnout tomorrow

Badlapje: I have implemented a filter with all elements that i consider 'safe' and 'not wrecking the styling'.  Quite a big list, but no table for example, nor a h1 or h2.  Works surprisingly easy :-)

MKodde (commented):
Excellent work! This will prevent a finding in the next security audit for sure. 

Try to get a whitelist that the PM's can agree upon. Maybe make it a configurable setting in the translation or `parameters.yml` file if that is at all possible. Maybe you can use the ConsentSettings value object to piggy back that setting into the application. I'd love to help with this, but monday is my next working day.

And next remove the `|raw` filter to further fortify the change.

theme/base/templates/modules/Default/Partials/warning.html.twig
Badlapje: Adjusted the list this morning after conferring with PM

Badlapje:
> Try to get a whitelist that the PM's can agree upon. 

it is ;-)

> Maybe make it a configurable setting in the translation or `parameters.yml` file if that is at all possible. 

Good idea :-)  I used the translation file.

> Maybe you can use the ConsentSettings value object to piggy back that setting into the application. I'd love to help with this, but monday is my next working day.

that'll be for a PR on monday then 😛 

> And next remove the `|raw` filter to further fortify the change.

That'd mean the html is no longer rendered, including the allowed tags.  You get the tags as a string, which is not the idea.  Basically that'd mean the feature is no longer available.

But by stripping the tags, the security hole is plugged, no?

theme/base/templates/modules/Default/Partials/warning.html.twig
```diff
@@ -2,5 +2,5 @@
     {% if notificationTitle is defined %}
         <h3>{{ notificationTitle }}</h3>
     {% endif %}
-    {{ text }}
+    {{ text|striptags('consent_warning_allowed_html'|trans)|replace({'<a ': '<a target="_blank "'})|raw }}
```
mroest, line 5: Deze eenvoudige replace is niet helemaal "waterdicht". Als iemand al het target attribute gebruikt... Is er een afspraak wat in de a tag gebruikt mag worden?

Volgens mij staat er een tailing spatie in de target

mroest (approved):

theme/base/templates/modules/Default/Partials/warning.html.twig
Badlapje: goeie van die spatie.

Akkoord dat ie niet waterdicht is, maar het is niet de bedoeling dat mensen al een target tag plaatsen.  Gewoon een a met href attribuut.  Heb er wel nog net een rel attribuut bijgeplaatst om de security/performance mooi te houden.

PR #1035, opened 2020-12-16
Fix the regression of the error pages in material

Badlapje:


MKodde:
It's impossible for me to judge if a regression was fixed as I can not read what went wrong.

I do see another small regression (i think) 

Alignment of the language switcher labels seems to be out of center. See screenshot

Screen Shot 2020-12-17 at 11 39 14

Other then that, I see no harm in your changes. So feel free to merge this PR. But next time maybe leave some more info for the reviewer. Enabling me to judge if the regression is actually fixed.

Tip, you should be able to run the screenshot tool using Cypress. Showing before/after screenshots for the error pages. Some before screenshots should be tracked in VCS..

PR #1038, opened 2020-12-16
Sort unconnected idps after connected idps on load

Badlapje:
- The sort works once searching has commenced, but not on load.  Fixed that in this PR.
- Also added a way to test this & the sort with more realistic data.
- The search navigation by arrow didn't work with the new sorting so fixed that.
- Noticed the submit button was hidden when the reset button was visible, so ensured it stays accessible to aural UI.

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Helper/TestEntitySeeder.php
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Helper/TestEntitySeeder.php
@@ -75,6 +75,74 @@ class TestEntitySeeder
     }
 
     /**
+     * Build a random collection of (unconnected) IdPs
+     *
+     * This is not an array of IdentityProvider value objects, but a derivative that can be used for showing IdPs on
+     * the WAYF.
+     *
+     * @param int $numberOfIdps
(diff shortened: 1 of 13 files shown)


src/OpenConext/EngineBlockFunctionalTestingBundle/Helper/TestEntitySeeder.php
```diff
@@ -74,6 +74,74 @@ public static function buildIdps($numberOfIdps, $numberOfUnconnectedIdps, $local
...
+            'Hotelschool The Hague',
+            'IHE Delft Institute for Water Education',
+            'KNMI',
+            'Koninklijke Nederlandse Akademie van Wetenschappen (KNAW)',
+            'Leids Universitair Medisch Centrum',
+            'Maastricht University',
+            'Netherlands eScience Center',
+            'SURF bv',
+            'Thomas More Hogeschool',
+            'VSNU',
```
mroest, line 118: Mijn eerste gevoel geeft aan dat deze lijst aan verandering onderhevig is. Weet je zeker dat dit een "hardcoded" lijst kan/mag zijn?

mroest (approved):

PR #1041, opened 2020-12-19
Remove stray space from consent screen disclaimer

Badlapje:
As reported by Thijs, there is a stray space after "heeft deze informatie nodig om te kunnen werken" in the first item of the disclaimerlist.  It was caused by formatting of the code.  This PR removes that.

image


MKodde:
@Badlapje the Consent behat test seems to suffer from this change. I'm going to investigate my dear Watson!

sherlock

MKodde:
Report: Behat tests are running on the old `openconext` theme. That is/was fine, but feature and text wise the themes are diverging more and more. 

I'm going to update the Behat tests for them to run against the new theme. This will involve quite a bit of tinkering. I will isolate that change to a PR of my own. For now, please be very wary of what test fails and reason if your code change has anything to do with that. If this is not the case. For now ignore the acceptance tests and :shipit:

MKodde (approved):

PR #1043, opened 2020-12-22
Ensure JS in the theme is overrideable

Badlapje:
See [the pivotal story](https://www.pivotaltracker.com/story/show/130523403)

As we do not want to change our JS bundler, i opted for the following approach:
- use a handler file to make callbacks overrideable
- use a selector file to make the UI selectors overrideable
- copy all JS files to the scaffold when creating a new theme

That ensures that a new theme can easily override the base theme JS.  It has as a downside that it duplicates code.  I don't know of / could not find a way to prevent the duplicate code without changing the JS bundler (i found references that webpack might be able to do it, but not sure).

This way is easily maintainable however and has the advantage that it's easy for new theme creators.  The downside seems worth it to me.


MKodde (commented):
I find this hard to review. I love some of the changes you made. For example, moving all selectors to a dedicated file is brilliant. The decoupling enables for easier refactoring/updating in the future. You know where to update a reference/selector without second thought.

Some of the other changes are abracadabra to me. Maybe ask a more advanced JS developer to co-review this ticket.

MKodde:
Rebased this branch on the parent, ensuring recent QA and other improvements are included in this refactor PR.

Badlapje:
> Some of the other changes are abracadabra to me. Maybe ask a more advanced JS developer to co-review this ticket.

Wessel will be reviewing this ticket on monday (he doesn't work today and it's too big of a PR to have been squeezed in yesterday)

PR #1046, opened 2020-12-24
Ensure build installs npm

Badlapje:
~The current makeRelease does not install npm packages.  While there is a build script, this script assumes `npm install` has already run.  Fixed that in this version.~

The script uses `npm ci` to ensure a cleanly installed `node_modules` is composed. In this ticket we also investigated other possible causes for the failing switch between themes. See individual commit messages for more details.

### DIFF
--- a/bin/makeRelease.sh
+++ b/bin/makeRelease.sh
@@ -79,7 +79,7 @@ fi
 # can be removed when we stop doing that
 echo "Build assets"
 cd ${PROJECT_DIR}/theme &&
-npm ci --unsafe-perm &&
+npm ci --unsafe-perm && npm install &&
 npm run release
 
 if [ $? -eq 0 ]; then
--- a/bin/makeRelease.sh
+++ b/bin/makeRelease.sh
@@ -79,7 +79,7 @@ fi
 # can be removed when we stop doing that
 echo "Build assets"
 cd ${PROJECT_DIR}/theme &&
-npm ci --unsafe-perm && npm install &&
+npm ci --unsafe-perm &&
 npm run release
 
 if [ $? -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


bin/makeRelease.sh
```diff
@@ -79,7 +79,7 @@ fi
 # can be removed when we stop doing that
 echo "Build assets"
 cd ${PROJECT_DIR}/theme &&
-npm ci --unsafe-perm &&
+npm ci --unsafe-perm && npm install &&
```
MKodde, line 82: I think adding the npm install is not required, npm ci is installing the node modules in-a empty folder.

theme/scripts/build.js
```diff
@@ -17,12 +17,7 @@ try {
     console.log('Reading contents of parameters.yml.\n');
     const fileContents = fs.readFileSync(config, 'utf8');
     const parameters = yaml.safeLoadAll(fileContents);
-
-    let theme = process.env.EB_THEME || parameters[0].parameters['theme.name'] || 'openconext';
-
-    if (process.env.EB_THEME) {
-        theme = process.env.EB_THEME;
-    }
+    const theme = process.env.EB_THEME || parameters[0].parameters['theme.name'] || 'openconext';
```
MKodde, line 20: I like this optimization. Should not have big impact, but makes things cleaner!

bin/makeRelease.sh
Badlapje: undid this

theme/scripts/build.js
MKodde: Maybe this is the time to move to the Skeune theme for the default theme that is to be installed?

MKodde (approved):
Personal thought, nitpick, nag,..

Instead of adding an undo commit, I personally would have popped the previous commit from the stack, removed the change and force pushed it back as if it never happened. I like your more honest solution too! Just showing another way to get it done.

Badlapje:
rebased this so the latest changes are in it, merging if it stays green

PR #1047, opened 2020-12-26
Optimize consent screenreader support

Badlapje:


theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/Attributes/defaultAttribute.html.twig
```diff
@@ -1,18 +1,20 @@
 {% set addTooltip = attributeMotivations[attributeIdentifier] is defined %}
+{% set id %}
+    tooltip{{ loop.index }}{{ orgId }}
+{% endset %}
```
MKodde, line 4: I've never seen this set/endset construction consciously before, but makes sense and looks very intuitive in use. Twig FTW!

On every successive use of the id `var` we use the `trim` filter to remove stray whitespaces from the identifier. Moving the trim filter option to the declaration provides a small optimization in my opinion. 

You be the judge.

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/Attributes/engineBlockAttribute.html.twig
```diff
@@ -1,16 +1,17 @@
 {% set addTooltip = nameIdIsPersistent %}
+{% set id %}
+    tooltip{{ loop.index }}{{ orgId }}
```
MKodde, line 3: See my previous `trim` suggestion. It might also apply here.

theme/skeune/translations/messages.en.php
```diff
@@ -50,7 +50,7 @@
     'consent_ok'    =>  'Yes, I agree',
     'consent_identifier_explanation'    => 'The identifier for this service is generated by %suite_name% en differs amongst each service you use through %suite_name%. The service can therefore recognise you as the same user when you return, but services cannot recognise you amongst each other as the same user.',
     'consent_provided_by'   => 'provided by',
-    'consent_tooltip_screenreader'  => 'Why do we need this info?',
+    'consent_tooltip_screenreader'  => 'Why do we need your %attr_name%?',
```
MKodde, line 53: Are these textual changes a PM/PO request? If this is the case, please add a reference to that Pivotal ticket in the PR description. If this is not yet part of a story, please add it as a task to one of the open leftover tickets. Or even open a new ticket. I'm not sure if you are keeping a list of 'to be translated to Portuguese' translation keys? I am not, but if you are also not keeping such a list, please create a task to do this in one of the open left-over tickets. I'd be happy to create this list and notify our resident Portuguese translator to translate the new texts.

Gots-to keep things nice and organized.

theme/base/javascripts/consent/addInteractiveAriaHandlers.js
```diff
@@ -1,22 +1,26 @@
...
+ * @param elementSelectors
+ */
+export const addInteractiveAriaHandlers = (elementSelectors
+) => {
+  const elements = nodeListToArray(document.querySelectorAll(elementSelectors));
 
   elements.forEach(element => {
-    const input = element.parentElement.querySelector(`input.${elementType}`);
+    const forValue = element.getAttribute('for');
+    const alert = document.querySelector(`[data-for="${forValue}"]`);
```
MKodde: Got to love this variable name. Brings back memories from way back when I used the browser built in `alert` messages to notify CMS users of their mistakes.

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/Attributes/tooltip.html.twig
```diff
@@ -12,5 +12,12 @@
...
     </div>
-    <div class="tooltip__value" role="alert" aria-hidden="true">{{ tooltipValue }}</div>
+    <div
+        aria-hidden="true"
+        class="tooltip__value"
+        data-for="{{ id|trim }}"
+        role="alert"
+    >
+        {{ tooltipValue }}
+    </div>
```
MKodde, line 22: Should we be concerned about the additional white-spaces that are added with this construction. Sometimes we use the `spaceless` Twig feature to prevent excessive white-spaces in the generated HTML. As this value is read by a screenreader, I thougth that might be relevant here? Or do they just ignore packer white-spaces?

MKodde (approved):
Lookin' good! See some suggestions and possible improvements to maybe dot some left over i's, but this change set is approved for merge IMHO.

theme/base/javascripts/consent/addInteractiveAriaHandlers.js
Badlapje: hehe, the good old "by closing this alert you'll open a new one" days 😄

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/Attributes/tooltip.html.twig
Badlapje: Good Q: in my tests so far they ignore it.  I've yet to try it out with JAWS (as i don't have a license) and VoiceOver (need to get my hands on a mac), but in previous tests with those for other projects they behaved the same.

I've always been a tad surprised Twig doesn't filter those extra whitespaces out though.  I'll add a spaceless tag to be on the safe side.

theme/skeune/translations/messages.en.php
Badlapje: No, when testing I simply noticed that all those tooltips show up as the same, which for SR users who navigate by lists (list of all tooltips / all headers / all links, ...) is rather opaque.

As for a list of to be translated to Portuguese keys: i don't have one, but given the translations are mostly separated, it shouldn't be too much work to make?  Clarified the task in the left over ticket for translations to incorporate that.

MKodde: For the sake of transparancy: maybe create a list of the newly added translation keys and the translation values for EN and NL and show them to either Thijs or Arnout. Let them see if they agree with the SR variations of the already existing translations. And consider adding a screenreader `marking` on these translation keys.

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/Attributes/defaultAttribute.html.twig
Badlapje: good point, fixed it 👍

theme/base/templates/modules/Authentication/View/Proxy/Partials/Consent/Attributes/engineBlockAttribute.html.twig
Badlapje: fixed

theme/skeune/translations/messages.en.php
Badlapje: agreed, the screenreader marking is there for the consent specific SR translation, will add for wayf.

PR #1050, opened 2021-01-04
Expand consent linguistics

MKodde:
Default consent is now also aliassed userconsent
Minimal consent was aliassed informational consent
The no consent option remained unchanged

This change was applied at several levels:
 - ConsentSettings added support for the two new aliasses (which will
   become the default in the future.
 - Translation keys where updated from minimal to informational.
 - Tests have been updated.

This change is requested in https://www.pivotaltracker.com/story/show/176185130


Badlapje (approved):
Nice work!  💯

PR #1053, opened 2021-01-04
Change translation as requested

Badlapje:


theme/skeune/translations/messages.en.php
```diff
@@ -27,7 +27,7 @@
     'wayf_remaining_idps_title_screenreader' => 'Add an account from the list below',
     'wayf_select_account_screenreader'       => 'Select an account from the list below',
     'wayf_search_placeholder'   => 'Search...',
-    'wayf_search_screenreader'          => 'Search identity providers',
+    'wayf_search_screenreader'          => 'Search identity %organisationNounPlural%',
```
MKodde, line 30: I feel you should remove the `identity` portion from the translation. I don't think this translation is intended: `Search identity organisations`. The Identity Provider term is not used inna end user context. They should not be confronted with protocol lingo. To them, they are authenticating to their organization/institute (however the implementer of the EngineBlock translated their orgNoun)

Same applies for the EN and PT translations.

theme/skeune/translations/messages.nl.php
```diff
@@ -27,7 +27,7 @@
     'wayf_remaining_idps_title_screenreader' => 'Voeg een account toe uit de onderstaande lijst',
     'wayf_select_account_screenreader'       => 'Selecteer een account uit de onderstaande lijst',
     'wayf_search_placeholder'   => 'Zoeken...',
-    'wayf_search_screenreader'          => 'Zoek een identity provider',
+    'wayf_search_screenreader'          => 'Zoek een identity %organisationNoun%',
```
MKodde, line 30: It might be a Dutch grammer quirk, but why is the `organizationNoun` used here, and the `organisationNounPlural` in English and Portuguese translations?

`Zoek een organisatie` seems to make sense, so this is probably good :+1:

MKodde (changes requested):
I think should do some additional trimming and weeding. But overall this is starting to look very good from a screen-reader point of view!

theme/skeune/translations/messages.nl.php
Badlapje: Dutch quirk with the identity modifier, but without it (as you suggested above), we can make it the singular noun for all 👍

theme/skeune/translations/messages.en.php
Badlapje: That's a critical on your spot check 💯

MKodde (approved):
Approved

PR #1055, opened 2021-01-06
Style debug page

Badlapje:
Ensure styling for the debug page from the old theme is still there in the base theme.

Added old theme as components in a dedicated folder to avoid duplication.  

One point of note is the problem with fontawesome.  We imported it via npm, including the file declaration like so:
`@import "../../../node_modules/font-awesome-sass/assets/stylesheets/font-awesome"`

However, the font files themselves were imported from our font folder like so (so without the ?v=4.7.0 version number at the end) :
```
@font-face {
    font-display: swap;
    font-family: 'FontAwesome';
    font-style: normal;
    font-weight: 900;
    src: url('/fonts/fontawesome-webfont.woff2') format('woff2'),
         url('/fonts/fontawesome-webfont.woff') format('woff'),
         url('/fonts/fontawesome-webfont.ttf') format('truetype');
}
```

This caused 404 errors in the console.  Any file using our font imports still worked though, as the browser did load the fonts and could see them.  But files that didn't (like the debug page) had no font-awesome fonts loaded and hence it didn't work.

I've reworked this in a common method for all, using everything from the npm package except the path declaration & replacing that with our own.


theme/base/templates/layouts/scripts/default.html.twig
```diff
@@ -16,20 +16,14 @@
...
     {% endspaceless %}{% endblock %}
-
-    <!-- Firefox CSS -->
-    <style>
-        @-moz-document url-prefix() {
-            .comp-links .help {
-                line-height: 1.7rem !important;
-            }
-        }
-    </style>
```
MKodde, line 27: Yeah! get rid of that old stuff

theme/base/stylesheets/pages/debug.scss
```diff
@@ -0,0 +1,13 @@
+@import "../../../node_modules/font-awesome-sass/assets/stylesheets/font-awesome";
```
MKodde, line 1: My installation has difficulties loading font awesome, see screenshot below

MKodde (changes requested):
Nice solution!

I noticed the foldername: `oldNotConverted` does not adhere to the folder naming convention: `old-not-converted` is the correct way.

Other than that, and one other feedback point. Looks like it never broke!

Badlapje:
> I noticed the foldername: `oldNotConverted` does not adhere to the folder naming convention: `old-not-converted` is the correct way.

*waves his fixup wand*

This spelling was always there 😇 (nothing like a bit of git magic to make you feel like a wizard)

MKodde:
Please don't mention the `src` folder structure. Which is infested with CamelCased folder names. That however is PSR-4 spicing and therefor allowed :rabbit2:

theme/base/stylesheets/pages/debug.scss
Badlapje: 't should be fixed :-)  Meanwhile also converted error-page.sass to error-page.scss

MKodde (approved):
You got it! Thanks for your patience.

Badlapje:
> You got it! Thanks for your patience.

And yours!  What a nailbiter in the end 🚵‍♂️

PR #1057, opened 2021-01-06
Style cookie page

Badlapje:
The cookie page was not redesigned, yet inherited the css for the new design.  That did not look great.

This PR fixes that and ensures that "fixing" other pages that were not redesigned is a lot easier.


MKodde (approved):
Very nice! Once the debug page changes are merged this is also good to go!

PR #1060, opened 2021-01-07
Egalize not converted pages

Badlapje:
The error page and the index/front page were also "not converted" but used their own way of handling that.  Egalized that to the same way as the debug/cookie page.

This is a part of [the "leftover" ticket](https://www.pivotaltracker.com/story/show/174968005)


MKodde (approved):
Thanks for expanding the readme!

Nothing but goodness here!

Moving the old styles to the components is clever, as we use it as a component. There are stylesheets that might not purely belong in that folder. But 'tis but a small complaint. Please leave as is.

PR #1056, opened 2021-01-06
Tighten cookie security

MKodde:
Several cookie changes:

1. The `http_only` flag was set on the `lang` cookie
2. The rememberchoice and previousSelection cookies where set to be `secure`

:deciduous_tree: To reduce JavaScript build output polution, the babel config was updated to no longer spit very verbose debug output.

See: https://www.pivotaltracker.com/story/show/176287495
See: https://github.com/OpenConext/OpenConext-deploy/pull/320

PR #1062, opened 2021-01-11
Remove HttpOnly flag from `lang` cookie

MKodde:
Other non-PHP components use the lang cookie in their Javascript,
setting the flag on this global cookie would mess up their operations.

Thanks @tveijen voor pointing this out!

See: https://github.com/OpenConext/OpenConext-deploy/pull/320#issuecomment-756109327
See: https://github.com/OpenConext/OpenConext-engineblock/pull/1056


Badlapje (approved):
:shipit:

PR #1066, opened 2021-01-25
Ensure hitting reset shows default idp

Badlapje:
Prior to this change, when the user pressed the reset button whilst searching; the default idp remained hidden.

This change ensures the default idp is shown in the above scenario.  It also adds the visibility of the default idp to the test, so if it fails it'll get detected.

[Issue brought up as part of the feedback round](https://wiki.surfnet.nl/display/coininfra/Bevindingen+nav+tests+door+supportteam)

### DIFF
--- a/theme/cypress/integration/skeune/wayf/wayf.general.spec.js
+++ b/theme/cypress/integration/skeune/wayf/wayf.general.spec.js
@@ -41,22 +41,9 @@ context('WAYF behaviour not tied to mouse / keyboard navigation', () => {
       cy.get('.wayf__idp--noAccess')
         .should('not.exist');
     });
-
-    it('Should show 5 disconnected IdPs', () => {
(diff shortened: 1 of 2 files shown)


theme/cypress/integration/skeune/wayf/wayf.general.spec.js
```diff
@@ -41,22 +41,9 @@ context('WAYF behaviour not tied to mouse / keyboard navigation', () => {
       cy.get('.wayf__idp--noAccess')
         .should('not.exist');
     });
-
-    it('Should show 5 disconnected IdPs', () => {
```
pablothedude, line 45: Has the removal of these test anything to do with this change and why are these not needed any more?

Badlapje: they were removed because i noticed - whilst working on this - that these tests are double.  They are present again right above the ones i deleted.

pablothedude (approved):
Thanks for the heads up!

PR #1072, opened 2021-01-28
Couple cypress tests to same selectors

Badlapje:
Couple cypress tests to same selectors

Prior to this change cypress tests were entirely decoupled from the selectors used in the rest of the application.  This resulted in tests breaking when one of those selectors was changed.

This change imports the selectors from the same file as the one used by the application JS, so that both have a single source of truth.  In instances where special test selectors are needed, a test-selector file is used which composes it's selectors maximally by incorporating the "normal" selectors.

[Issue brought up as part of fixing the IE11 bug for the latest feedback round](https://wiki.surfnet.nl/display/coininfra/Bevindingen+nav+tests+door+supportteam)

### DIFF
--- a/theme/base/javascripts/selectors.js
+++ b/theme/base/javascripts/selectors.js
@@ -10,6 +10,13 @@ export const metadataCertificateLinkSelector = 'dl.metadata-certificates-list a'
 export const indexPageSelector = '#engine-main-page';
 
 /***
+ * ERROR PAGE SELECTORS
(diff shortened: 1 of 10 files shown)


Badlapje:
Note, this should only be reviewed after the IE11 bugfix branch is merged.  Or you should only look at the files for the cypress tests when reviewing ;-)

pablothedude (approved):

PR #1074, opened 2021-02-01
Ensure selected idp list works right

Badlapje:
Prior to this change deleting a selected idp was not possible.  While it was visually removed, returning to the wave later showed it was still there.

This change ensures the behaviour is as expected:
- hitting delete button truly deletes it
- returning to the wave without a hard refresh will show the correct list of already selected idps, even if returning via the backbutton


pablothedude (approved):

pablothedude:
I approved the PR' but please fix the tests before merging.

PR #1075, opened 2021-02-01
Adjust styling of PEP page

Badlapje:
Prior to this change the text in the error message was bolded.

This change changes the bold error message to a normal weight error message.  It also adds, in the readme, a link to the page containing functional testing endpoints for all different error pages, so they can be tested.
Last but not least an accessibility improvement was included: the logo now has an empty alt text.

### DIFF
--- a/theme/base/templates/modules/Authentication/View/Feedback/authorization-policy-violation.html.twig
+++ b/theme/base/templates/modules/Authentication/View/Feedback/authorization-policy-violation.html.twig
@@ -6,7 +6,6 @@
 {% block pageHeading %}{{ pageTitle }}{% endblock %}
 
 {% block errorMessage %}
-
     <h2>{{ 'error_authorization_policy_violation_info'|trans|raw }}</h2>
 
     {% if logo is not null %}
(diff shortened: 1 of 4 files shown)


theme/base/templates/modules/Authentication/View/Feedback/authorization-policy-violation.html.twig
```diff
@@ -6,14 +6,14 @@
...
 {% block errorMessage %}
-
     <h2>{{ 'error_authorization_policy_violation_info'|trans|raw }}</h2>
 
     {% if logo is not null %}
         <img src="{{ logo.url }}?v={{ assetsVersion }}"
              class="pull-right"
              width="{{ logo.width }}"
              height="{{ logo.height }}"
+             alt=""
```
pablothedude, line 16: Is there no sane default as value for alt for example "Identity provider logo"?

pablothedude (approved):

theme/base/templates/modules/Authentication/View/Feedback/authorization-policy-violation.html.twig
Badlapje: The logo is visually appealing but doesn't add anything to the content of the page.  Given that it isn't content, the alt should be an empty string.  That signals to a screen-reader that they don't need to mention it to their users.

PR #1077, opened 2021-02-03
Ensure IE11 search works

Badlapje:
Prior to this change the search in IE11 did not function properly

This change fixes the search.  Turns out querySelectorAll behaviour in IE11 is weird.  So additional checks and balances were added to detect when/if IE acts up.  The element being searched/sorted was also changed from the idp-element to the li.  The change also affected arrow behaviour / cookie behaviour so that was adjusted to fit the changes as well.

### DIFF
--- a/theme/base/javascripts/wayf/searchBehaviour.js
+++ b/theme/base/javascripts/wayf/searchBehaviour.js
@@ -4,6 +4,7 @@ import {searchAndSortIdps} from './search/searchAndSortIdps';
 import {toggleDefaultIdPLinkVisibility} from "./search/toggleDefaultIdPLinkVisibility";
 import {toggleSearchAndResetButton} from "./search/toggleSearchAndResetButton";
 import {remainingIdpLiSelector, searchFieldSelector, searchResetSelector} from '../selectors';
+import {focusAndSmoothScroll} from '../utility/focusAndSmoothScroll';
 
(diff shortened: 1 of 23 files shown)


theme/base/javascripts/wayf/searchBehaviour.js
```diff
@@ -4,21 +4,35 @@ import {searchAndSortIdps} from './search/searchAndSortIdps';
...
 import {remainingIdpLiSelector, searchFieldSelector, searchResetSelector} from '../selectors';
+import {focusAndSmoothScroll} from '../utility/focusAndSmoothScroll';
 
 export const searchBehaviour = () => {
   const idpList = document.querySelectorAll(remainingIdpLiSelector);
   const searchBar = document.querySelector(searchFieldSelector);
   const resetButton = document.querySelector(searchResetSelector);
   const idpArray = nodeListToArray(idpList);
 
+  // const zeArray = [];
```
pablothedude, line 15: Forget to remove this while debugging, please remove this dead code.

pablothedude (approved):

theme/base/javascripts/wayf/searchBehaviour.js
Badlapje: good spot! 💯

PR #1079, opened 2021-02-05
Add logging to ValidateMfaAuthnContextClassRef command.

thijskh:
This so we can know from the logs that a login actually happened with
this context. The log_logins logger also includes the
AuthnContextClassRef but this is not complete as it will not be set
to the right value for legacy ADFS which has it in an attribute.

### DIFF
--- a/library/EngineBlock/Corto/Filter/Command/ValidateMfaAuthnContextClassRef.php
+++ b/library/EngineBlock/Corto/Filter/Command/ValidateMfaAuthnContextClassRef.php
@@ -17,6 +17,7 @@
  */
 
 use OpenConext\EngineBlock\Assert\Assertion;
+use OpenConext\EngineBlock\Logger\Message\AdditionalInfo;
 use OpenConext\EngineBlock\Metadata\TransparentMfaEntity;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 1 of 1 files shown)


library/EngineBlock/Corto/Filter/Command/ValidateMfaAuthnContextClassRef.php
```diff
@@ -45,9 +46,16 @@ public function __construct(LoggerInterface $logger)
     public function execute()
     {
         $originalSP = $this->_server->findOriginalServiceProvider($this->_request, $this->logger)->entityId;
+        $IdP = $this->_identityProvider->entityId;
         $mfaEntity = $this->_identityProvider->getCoins()->mfaEntities()->findByEntityId($originalSP);
+
+        $additionalInfo = AdditionalInfo::create()->setIdp($IdP)->setSp($originalSP);
```
MKodde, line 52: Nice use of the AdditionalInfo log helper!

library/EngineBlock/Corto/Filter/Command/ValidateMfaAuthnContextClassRef.php
```diff
@@ -45,9 +46,16 @@ public function __construct(LoggerInterface $logger)
     public function execute()
     {
         $originalSP = $this->_server->findOriginalServiceProvider($this->_request, $this->logger)->entityId;
+        $IdP = $this->_identityProvider->entityId;
```
MKodde, line 49: Nitpick: all variable names should be in lower camel case. So please use `$idp`. When this code is migrated to the Symfony app, the code quality inspection would fail on this naming issue. So no big deal, we would fish it out then.

MKodde (approved):

library/EngineBlock/Corto/Filter/Command/ValidateMfaAuthnContextClassRef.php
```diff
@@ -57,12 +65,16 @@ public function execute()
...
+                    $IdP,
                     $originalSP,
                     $mfaEntity->level()
                 )
             );
         }
+
+        $message = sprintf('Validated presence of AuthnContextClassRef "%s" in assertion',
+            $mfaEntity->level());
+        $this->logger->notice($message, ['additional_info' => $additionalInfo->toArray()]);
```
MKodde, line 77: QA tooling is nagging about this indentation. Should be like this

```
sprintf(
   'foo bar',
   $mfaEnt->level()
);
```

thijskh:
Rebased for fixed tests and squash

MKodde (approved):
Thanks for the improvements!

PR #1080, opened 2021-02-08
Update security checker to use local checker

MKodde:
The composer depencency is no longer operable, and was removed. This
change uses the new way of running the checker, which is to download the
local checker binary an running that.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -125,7 +125,7 @@ jobs:
               run: |
                   cd docker && docker-compose exec -T php-fpm.vm.openconext.org bash -c '
                       echo -e "\nSensioLabs Security Check\n" && \
-                      ./vendor/bin/security-checker security:check && \
+                      ./bin/securityChecker.sh && \
                       cd theme && \
                       echo -e "\nNPM Audit\n" && \
                       npm run audit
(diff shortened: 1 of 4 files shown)


MKodde:
@thijskh this fixes what keeps breaking your PR. The previously used security checker bin is no longer supported.

docker/php-fpm/Dockerfile-php74
```diff
@@ -12,6 +12,7 @@ RUN apt-get update && apt-get install -y \
     git \
     nodejs \
     zip \
+    wget \
```
thijskh, line 15: Since `curl` is already available, use that for the download?

https://github.com/OpenConext/OpenConext-engineblock/blob/de7177276632d9a445e700817c4cc4ceaded4759/docker/php-fpm/Dockerfile-php72#L10

thijskh (approved):

docker/php-fpm/Dockerfile-php74
MKodde: Might be my `curl` noobieness, but when I tried to use curl to download the file, I was unable to specify the target name of the file. And was tempted to use the wget solution. I wil give it another quick try tho.

tvdijen: The lowercase `-o` should allow you to specify a destination

MKodde: `>` worked too, my problem turned out to be that the rest of the script did not wait for curl to finish. Did not know curl was doing things async

thijskh (approved):

PR #1081, opened 2021-02-09
Keep WAYF operable when syncing cookies

Badlapje:
Prior to this change, it was possible that the WAYF was rendered inoperable when syncing cookies.  This happened when an idp was in the cookie, but not available when connecting to a service.  The resulting null error in the JS rendered the entire wayf inoperable.

This change ensures the cookie-syncing doesn't freeze the WAYF.  The problem should be solved with the extra null checks.  But as the sync should never freeze the wayf, it is now wrapped in a try/catch block to be on the safe side.

Issue discovered as part of acceptation testing and [reported in the wiki](https://wiki.surfnet.nl/display/coininfra/Bevindingen+nav+tests+door+supportteam).  It can be tested on functional by adding some of the regular "fake idps" to your selected accounts (preferably only one), then testing with [the random idp link](https://engine.vm.openconext.org/functional-testing/wayf?randomIdps=20)

### DIFF
--- a/theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
+++ b/theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
@@ -4,61 +4,74 @@ import {sortIdpList} from './utility/sortIdpList';
 import {getData} from '../utility/getData';
 
 export const matchPreviouslySelectedWithCookie = () => {
-  const cookieName = JSON.parse(document.getElementById(configurationId).innerHTML).previousSelectionCookieName;
-  const selectedIdps = document.querySelectorAll(selectedIdpsSelector);
-  const selectedIdpList = document.querySelector(selectedIdpsListSelector);
-  let cookie = Cookies.get(cookieName);
-  let listMustBeSorted = false;
+  try {
+    const cookieName = JSON.parse(document.getElementById(configurationId).innerHTML).previousSelectionCookieName;
+    const selectedIdps = document.querySelectorAll(selectedIdpsSelector);
+    const selectedIdpList = document.querySelector(selectedIdpsListSelector);
+    let cookie = Cookies.get(cookieName);
+    let listMustBeSorted = false;
 
-  if (cookie) {
-    if (selectedIdps) {
-      for (const idp of selectedIdps) {
-        const id = getData(idp, 'entityid');
+    if (cookie) {
+      if (selectedIdps) {
+        for (const idp of selectedIdps) {
+          const id = getData(idp, 'entityid');
 
-        // check if idp is in cookie, if not remove it from the list
-        if (cookie && cookie.indexOf(id) === -1) {
-          idp.parentElement.remove();
-          listMustBeSorted = true;
+          // check if idp is in cookie, if not remove it from the list
+          if (cookie && cookie.indexOf(id) === -1) {
+            idp.parentElement.remove();
+            listMustBeSorted = true;
+          }
         }
+      }
+
+      cookie = JSON.parse(cookie);
+      // check if each idp in the cookie is in the list, if not add it
+      cookie.forEach(idp => {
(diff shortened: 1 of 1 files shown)


theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
```diff
@@ -4,61 +4,74 @@ import {sortIdpList} from './utility/sortIdpList';
...
+    const selectedIdpList = document.querySelector(selectedIdpsListSelector);
+    let cookie = Cookies.get(cookieName);
+    let listMustBeSorted = false;
 
-  if (cookie) {
-    if (selectedIdps) {
-      for (const idp of selectedIdps) {
-        const id = getData(idp, 'entityid');
+    if (cookie) {
+      if (selectedIdps) {
```
MKodde, line 15: Boy scout opportunity:

The double `if` block can be easily made into one by utilizing an `and` comparison operator.

I know this was in place before your change

theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
```diff
@@ -4,61 +4,74 @@ import {sortIdpList} from './utility/sortIdpList';
...
+        }
+      });
+
+      if (listMustBeSorted) {
+        const selectedIdpLis = document.querySelectorAll(selectedIdpsLiSelector);
+        sortIdpList(selectedIdpLis, 'previous');
+      }
     }
+  } catch (e) {
+    console.log(e);
```
MKodde, line 75: Is this what we want? We usually not log anything to console but in this case (exception occurred) it might be useful.

MKodde (approved):
Looking good from a code review view. Have not tested this functionally, can you provide steps to reproduce the faulty behavior so I can do this?

theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
Badlapje: it can, but then lines 26-62, which are outside of the if (selectedIdps) block but inside the if (cookie) block would not get executed if there's no selectedIdps.  Which would ruin the functionality.

I could split it into
```
if (cookie && selectedIdps) { // line 15-23 }
if (cookie) { // line 26-62 }
```
but i'm not sure that's actually an improvement?

theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
Badlapje: Absent a logging system of our own, i'd use this indeed.  It's not optimal, but it should - in theory - never trigger.  Just a failsafe to get some feedback in case it is triggered on an unforeseen edgecase.

Badlapje:
> Looking good from a code review view. Have not tested this functionally, can you provide steps to reproduce the faulty behavior so I can do this?

- check out the code before this fix & build
- load the functional testing wayf
- ensure you have only 1 previously selected account (either by deleting the others, or by adding one, depending on your current situation)
- use [the random IDP testing link](https://engine.vm.openconext.org/functional-testing/wayf?randomIdps=20) until you don't see the previously selected account section (just refresh a couple of times, sooner or later the entityId in the cookie won't be present on the page).
- the WAYF should now not work and the console should have an error

To verify the fix, just repeat the steps above, but with the code of this branch.

theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
MKodde: Ah, now that you mention this.. yeah, maybe leave as is. I did not inspect the code well enough as you just proved!

MKodde:
Thanks for describing a means to reproduce the error. Issue fixed, verified in my dev VM.

PR #1083, opened 2021-02-10
Ensure IdPs are deletable after sync

Badlapje:
Prior to this change, idps added to selected accounts after cookie sync were not deletable because they still had the "remaining" class on them.

This change removes that class ensuring the newly added idp is also deletable.

Steps to reproduce:
- have one previously selected account
- choose a new account to  log in with
- hit the backbutton
- try to delete the newly added account => shouldn't work before this fix, should work after

### DIFF
--- a/theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
+++ b/theme/base/javascripts/wayf/matchPreviouslySelectedWithCookie.js
@@ -43,6 +43,7 @@ export const matchPreviouslySelectedWithCookie = () => {
           }
 
           if (!!clone) {
+            clone.classList.remove('remaining');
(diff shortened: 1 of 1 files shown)


MKodde (commented):
I tried to reproduce the behavior but was unable. Was this in a specific browser? What I did:

1. Choose IdP 2
2. Go back to WAYF
3. Hit F5 to get the previously selected screen to pop up
4. Choose the 'use anothre account' button
5. Select a different IdP
6. Back on the wayf afterwards, I'm able to remove the IdPs no problem.

Tested in FF on `master` and this branch

MKodde (approved):
I was unable to reproduce the problem! Good to go in my book!

PR #1084, opened 2021-02-10
Avoid double encoding logo url for attribute source.

thijskh:
Before, the resulting HTML looked like this:

```html
<img class="logo-small" src="https&#x3A;&#x2F;&#x2F;static.surfconext.nl&#x2F;media&#x2F;aa&#x2F;voot.png?v=6.4.6" alt="">
```

As a side note, this url comes from our own translations files.


Badlapje (approved):
nice work!

PR #1085, opened 2021-02-10
Ensure images after search are ok in IE11

Badlapje:
Prior to this change, when searching all logos became the same placeholder logo in IE11 only.

This change ensures the image url is the correct one, so the logo is the correct one.


MKodde (approved):
Nifty!

PR #1086, opened 2021-02-10
Prevent logo flicker in FF when possible

Badlapje:
Prior to this change, the logo flickers in FF whenever there's an interaction with the search bar.

This change limits the flicker to actual redraws of the idp list by preventing the search handler from being triggered if the previous search term matches the current one.

### DIFF
--- a/theme/base/javascripts/wayf/searchBehaviour.js
+++ b/theme/base/javascripts/wayf/searchBehaviour.js
@@ -11,21 +11,33 @@ export const searchBehaviour = () => {
   const searchBar = document.querySelector(searchFieldSelector);
   const resetButton = document.querySelector(searchResetSelector);
   const idpArray = nodeListToArray(idpList);
+  let previousSearchTerm = '';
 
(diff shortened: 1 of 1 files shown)


MKodde:
Is it possible to only redraw the list if the search results have actually changed? When typing IdP, the weight of the search results might change, but that does not have effect on the actual search results as to how they are rendered. This might prevent more flickering.

Badlapje:
> Is it possible to only redraw the list if the search results have actually changed? When typing IdP, the weight of the search results might change, but that does not have effect on the actual search results as to how they are rendered. This might prevent more flickering.

I think it is, but that'd mean it's a performance hit for all browsers.  We'd have to look over each item in the list to see if it's still in the same order as it was before the new search.  And that for every letter typed.  For big IdP lists that might become something of a slowdown.  Not sure that's worth it.

MKodde:
Nah, let's leave it as is for now.

Shipit

MKodde (approved):

PR #1087, opened 2021-02-11
Prevent idprow-logo from overlapping with text

Badlapje:
Prior to this change, the logo in the idp-row could overlap with the text.

This change ensures that's no longer possible.  After the change this looks as follows now: 
image

### DIFF
--- a/theme/base/stylesheets/pages/consent/idpRow.scss
+++ b/theme/base/stylesheets/pages/consent/idpRow.scss
@@ -1,6 +1,6 @@
 .idpRow {
     @include display-grid;
-    @include grid-template-columns(10% 1% 50% 1% 37% 1%);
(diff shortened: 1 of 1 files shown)


MKodde (approved):
A very long horizontal image could potentially still cause this problem if I interpret this change correctly. Will probably never be an issue.. Maybe some overflow hidden trick might be usefull in that case?

Badlapje:
> 
> 
> A very long horizontal image could potentially still cause this problem if I interpret this change correctly. Will probably never be an issue.. Maybe some overflow hidden trick might be usefull in that case?

Tested it with a logo that was twice the available width.  The `max-width: 100%` rule should prevent it, but it never hurts to add extra security so added `overflow: hidden` 💯

PR #1088, opened 2021-03-01
Use PDP as source for LoA enforcement

thijskh:
In addition to the existing per SP and per-IdP/SP ways to configure stepup LoA enforcement in EB, this PR adds interpretation of responses from the PDP. We can rather easily combine the existing PDP client with the existing SFO client.

OpenConext-PDP has the notion of Stepup policies which will result in a Permit response with a LoA obligaion returned if they match. for the decision are the existing inputs of the PDP (SP, IdP, user's attributes) and now also the user's IP address. An example rule: if the eduPersonAffiliation == employee, then Permit with LoA obligation = http://test.surfconext.nl/loa2. 

The question that EB asks of the PDP remains the same with the addition of the client IP address as an extra attribute. Responses can still be Permit or Deny, but a Permit can have additional obligations. We search these for a LoA and if found, require this LoA via the existing stepup callout.

The basic functionality works and includes unit tests and some documentation.

One question was how to deal with typehints. I’ve now more or less tried to use the same style as the surrounding code.

The build fails because of a security vulnerability in a dev dependency, which is out of scope for this PR.

Open work:

- [x] Feature must be tested with behat
- [x] A PDP obligation now trumps any SP or IdP-SP configured LoA, even if higher. We may need a more sophisticated resolution mechanism.
- [x] Only the first loa obligation is recognised. PdP can return multiple if multiple rules match. Devise a way to deal with it (or just forbid it for now)


src/OpenConext/EngineBlock/Stepup/StepupDecision.php
```diff
@@ -67,21 +76,28 @@ public function __construct(IdentityProvider $idp, ServiceProvider $sp, LoaRepos
...
     }
 
     /**
      * @return Loa|null
      */
     public function getStepupLoa()
     {
+        if ($this->pdpLoa) {
+            return $this->pdpLoa;
+        }
```
MKodde: Maybe document in code why the pdpLoa takes precedence over the spLoa and idpLoa? That did not become immediately clear to to me when reviewing this code.

docs/stepup_callout.md
```diff
@@ -1,17 +1,26 @@
...
+It's possible to require second factor authentication in EngineBlock using OpenConext StepUp in second factor only (SFO) mode. When configured correctly, EngineBlock will utilize StepUp Gateway's SFO to do a second factor callout with a given minimum Level of Assurance (LoA).
 
+Currently the following configrations are supported:
+* On a per-sp basis: configure the required LoA in Manage which will be enforced for all logins to this SP by any IdP.
+* On a idp-sp basis: configure the required loa in Manage for an IdP + a specific SP. All users from this IdP will for logging into this SP require a token with this LoA.
+* Based on PDP rules. The policy enforcement point in Engineblock can interpret decisions from the PDP that specify a minimum LoA.
+
+Engineblock disallows configuration at both the SP and the IdP-SP combination at the same time. A PDP decision for a LoA currently trumps any SP or IdP-SP setting.
```
MKodde: Ah, disregard my previous comment, you did document which Sp/IdP/PdP loa obligation takes precedence over which. I still think this would be a good in-line documentation opportunity though.

src/OpenConext/EngineBlockBundle/Pdp/PolicyDecision.php
```diff
@@ -46,9 +46,9 @@ final class PolicyDecision
     private $statusMessage;
 
     /**
-     * @var string|null
+     * @var array
```
MKodde, line 49: It may be usefull to typehint as such: `@var Obligations[]`, same goes for the return typehint on line 105. That could be similarly hinted.

This aids the IDE in autocomplete tasks

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Stepup.feature
```diff
@@ -24,7 +24,7 @@ Feature:
       And I pass through EngineBlock
     Then the url should match "/functional-testing/SSO-SP/acs"
 
-    Scenario: Stepup authentication should be supported if set trough IdP configuration mapping
+    Scenario: Stepup authentication should be supported if set through IdP configuration mapping
```
MKodde, line 27: :+1: for fixing the typo!

MKodde (commented):
Nice work! I have some suggestions in the comments below.

src/OpenConext/EngineBlockBundle/Pdp/Dto/Response/Obligation.php
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace OpenConext\EngineBlockBundle\Pdp\Dto\Response;
+
+final class Obligation
```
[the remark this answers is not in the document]
thijskh: This was done so it's the same as the other classes that constituate a Response. So we should change all or none. For now I'll just go with what was in use.

src/OpenConext/EngineBlock/Stepup/StepupDecision.php
thijskh: Well, in fact it doesn't. This is in a later commit in this PR improved by a more sophisticated LoA selection mechanism.

docs/stepup_callout.md
thijskh: This doc is also changed in a subsequent commit. I could squash them all together?

src/OpenConext/EngineBlockBundle/Pdp/PolicyDecision.php
```diff
@@ -91,6 +98,23 @@ public static function fromResponse(Response $response)
         return $policyDecision;
     }
 
+    /**
+     * Checks obgligations for any stepup LoA requirements, returns first one found.
+     *
+     * @param array|null $obligations
+     * @return string|null
+     */
+    private static function findStepupObligation($obligations)
```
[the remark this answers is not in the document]
thijskh: It did not and it happens all the time in the current code...

src/OpenConext/EngineBlockBundle/Pdp/Dto/Response.php
```diff
@@ -146,6 +152,26 @@ public static function fromData(array $jsonData)
             }
         }
 
+        if (isset($responseData['Obligations'])) {
+            foreach ($responseData['Obligations'] as $obligationData) {
+                $obligation = new Obligation;
```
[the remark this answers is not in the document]
thijskh: It now only has public properties. This again mirrors how the rest of the Dto is currently built. See: AttributeAssignment and AssociatedAdvice in the same dir.

src/OpenConext/EngineBlockBundle/Pdp/PolicyDecision.php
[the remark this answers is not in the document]
thijskh: But: this return statement has already been changed in a later commit.

thijskh:
@MKodde Thanks for the review. I've decided to add the type hints to everything newly added and to everything under Stepup/ and PdP/, so these modules are now consistently typed. For the same reasons of consistency I've decided not to change the finality and immutability of the Dto classes. I think I've addressed all review comments now.

MKodde (approved):
Thanks for the fixes, LGTM!

PR #1090, opened 2021-03-10
Fix errors flagged by nu html checker

Badlapje:
Prior to this change, there were 3 errors & one warning on the page: three wrong ids in the svgs & one section without a title (warning).

This change ensures that those errors & the warning are fixed.


MKodde (approved):
I see nothing wrong with this. And if it improves a11y, more power to you!

PR #1091, opened 2021-03-10
Handle accessibility for hideable attributes

Badlapje:
Prior to this change, there were several issues with the hidden attributes:
- their tooltips were not removed from the tab order when hidden
- showing all hidden attributes did not result in a focus change, meaning SR users were left in the dark

This change ensures that the correct attribute get's focus when showing the hidden attributes.  It also ensures the hidden attribute tooltips are no longer in the taborder.

### DIFF
--- a/theme/base/stylesheets/pages/consent/attributeValue.scss
+++ b/theme/base/stylesheets/pages/consent/attributeValue.scss
@@ -20,8 +20,9 @@
     }
 }
 
-.attribute__valueWrapper {
+.attribute__valueWrapper:not(.joske) {
     @include grid-position(1, 2, 3, 6);
+    padding: 0;
 
     @include screen('mobile') {
         @include grid-position(2, 3, 1, 2);
@@ -29,7 +30,7 @@
 
     > .attribute__value.attribute__value--list {
         margin: 0;
-        padding: 0;
+        padding: 1rem 0;
 
         &.animated {
             max-height: initial;
@@ -43,6 +44,8 @@
 
         li:not(.consent__attributeNested--noTooltip) {
             &:first-of-type {
+                margin-top: -4px;
+
                 .attribute__value {
                     display: inline-block;
                     padding-right: 0;
@@ -51,12 +54,19 @@
 
                 .ie11__label {
                     display: inline-block;
+                    margin-top: -4px;
+                    padding-top: 12px;
                     width: 12%;
+
+                    label {
+                        background-position: left 24px top 1.1rem
+                    }
                 }
             }
 
             .attribute__value {
                 padding-right: 1rem;
+                vertical-align: text-bottom;
             }
         }
     }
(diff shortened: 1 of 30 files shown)


theme/base/stylesheets/pages/consent/attributeValue.scss
```diff
@@ -20,16 +20,17 @@
     }
 }
 
-.attribute__valueWrapper {
+.attribute__valueWrapper:not(.joske) {
```
MKodde, line 23: `joske` seems to be a ghost in the system? Grepping the change-set yield no other results. I later found a reference to Joske in the cypress tests, but no need to reference that in a production bound stylesheet?

theme/skeune/templates/modules/Default/Partials/a11yButtonExpandable.html.twig
```diff
@@ -0,0 +1,2 @@
+<span class="visually-hidden expandable">{{ 'button_expandable_screenreader'|trans }}</span>
```
MKodde, line 1: I'm no expert, but looking at the translation for this key it makes little sense to me as to what is happening here.

 theme/skeune/translations/messages.en.php:
`'button_expandable_screenreader' => ', button, expandable',`

The translation seems to be a list of classes (which might also better not be translatable). but they are rendered in the visually hidden `span` element. 

Leaves me puzzled.

MKodde (commented):
Given you have functionally tested this yourself and I'm only able to test SR stuff with a very limited reader (Orca) I've only code reviewed this. 

Two things aside, this looks real good. Nice work!

theme/base/stylesheets/pages/consent/attributeValue.scss
Badlapje: joske is a reference to [Joske Vermeulen](https://www.youtube.com/watch?v=mR6Uson36lE).  The reason a joske class is added here is because the specificity of the selector needed to be high enough.  Adding a :not selector does the trick, but then you need to put something between the brackets which won't exist.  

The perfect opportunity for some mild humor.

theme/skeune/templates/modules/Default/Partials/a11yButtonExpandable.html.twig
Badlapje: Wel, normally when one uses a button element the screenreader will read the text and then also state "button".  Because we use a checkbox with role=button, but click on the label, not all SRs announce that it's a button.  So to solve that we add SR-text only that adds the button description so the result is the same as with a normal button.  Adding that role to the label is a no-go because that's invalid html (because html has some weirdly overlooked use-cases for elements).

Because we use this "button" to toggle the display of content, it's a disclosure widget aria-wise.  That means we need to put `aria-expanded` (and `aria-pressed` cause it's a button) attributes on a checkbox.  Aria-expanded is normally announced by the screenreader as "expandable" in the false state.  But since we don't use a normal button, that once again is not announced in all screenreaders.

Badlapje: If they are ever kind enough to add a :pressed selector for buttons, akin to the :checked selector for a checkbox we can make it a lot easier to have a disclosure widget that functions when CSS is off and which is perfectly accessible.

Badlapje:
Merged this already given @MKodde is a tad sick and unlikely to reply shortly.  If the above comments yield any additional feedback i'll address it in a new branch

PR #1092, opened 2021-03-11
Handle a11y for pressing cancel button in consent screen

Badlapje:
Prior to this change, pressing the cancel button was implemented for SRs & keyboard users with an aria alert and changes to the aria-hidden state.

This change ensures that clicking the cancel button is implemented for SRs & keyboard users as a disclosure widget.  That means the focus is moved to the title when the nok section is shown and to the button when it the user moves back.  It also means the aria-expanded attribute is set from false to true (or vice versa) depending on whether or not the nok section is shown.


MKodde (approved):
Aside from the fixup commit that is not yet squashed, this is good to go.

PR #1093, opened 2021-03-19
Remove assertsversion tag from externally hosted images.

thijskh:
The IdP/Attribute source logo are hosted on a static vhost and do not
change as a result of a new EB software version. They may change for
other reasons but the EB assetversion has no relation to it. Remove the
assetsversion from the request so we do not needlessly invalidate these
logos when upgrading EB.


thijskh:
The build fails (only) on an unrelated library security alert.

thijskh:
Build fixed.

MKodde (approved):
Thanks for refining the image invalidation changes I blunt-axed in previously!

FYI, I rebased your commit onto `master` as Koen's PR had already been merged back.

PR #1094, opened 2021-03-23
Resolve insecure packages to secure version

Badlapje:
Prior to this change, there were 7 security warnings, of which 1 was high and needed manual review.

This change implements a resolution for the high security threat (manual review) and audit fix + npm upgrade for the others.


MKodde (approved):
t'is good!

PR #1096, opened 2021-03-29
Reduce missing translation warnings

MKodde:
Several changes where made to the WAYF generation logic to reduce the amount of missing translation warnings.

1. Only load the IdP name for the current locale. This gets rid of the missing PT translation on our platform where PT is not an active language. This should solve similar issues for OpenConext users that use only one or two of the supported languages.
2. Translated the `required` translation id
3. Lowered the log level for this missing IdP name log entry.

See: https://www.pivotaltracker.com/story/show/176817180

### DIFF
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -109,7 +109,7 @@ HTML
     'cancel'                => 'Cancel',
     'send'                  => 'Send',
     'close'                 => 'Close',
-
+    'required'              => 'Required',
     'send_confirm'          => 'Your request has been sent',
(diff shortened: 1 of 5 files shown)


Badlapje (changes requested):
Cypress tests are failing, so that needs fixing.

Why was the required translation added?  I don't see it being used in the changes, was it generating spam in the logs?  

Other than that nice work!

MKodde:
Ah, the mighty JS test suite once again saves us from a regression! Will fix that right away.

> Why was the required translation added? I don't see it being used in the changes, was it generating spam in the logs?

Yes, it logged a warning on every WAYF build.

PR #1099, opened 2021-03-30
WAYF a11y improvements

Badlapje:
Improve screenreader experience for WAYF

Prior to this change, the screenreader experience was not optimal or AA compliant.

This change:
- changes the descriptive text of the searchfield for a SR
- adds the idpBanner as a description for a list
- ensures only the reset button is not focusable if it is invisible
- ensures the No Results section is announced to a SR
- ensures search results are announced to a SR
- changes the title text for the remaining section which only SR users can access
- handles the SR/focus of the previous accounts well:
    - add notification of what element was deleted
    - ensure focus remains ok
    - change text of delete button so that it also says which account will be deleted
- handles no access SR/focus correctly:
    - ensures the title is read along with the paragraph below
    - sets focus to the title on change
    - removes focusability of the idp
    - hides the delete disable button of the idp from SR
    - removes the login form from the idp
    - switches focus correctly when opening the idp form
    - handles validation for SR
    - switches focus correctly when submitting the form succesfully
    
As per the boyscout rule it also fixes the following:
- ensures the noaccess section is always correctly hidden when hitting cancel / submitting the form (this was not always the case in rare edge cases)
- updates the caniuse db    

Issue found when learning for WAS.

### DIFF
--- a/theme/base/javascripts/handlers.js
+++ b/theme/base/javascripts/handlers.js
@@ -7,6 +7,7 @@ import {switchConsentSection} from './consent/switchConsentSection';
 import {addClickHandlerOnce} from './utility/addClickHandlerOnce';
 import {
   backButtonSelector,
+  firstRequestFieldId,
   nokButtonSelector,
   tooltipsAndModalLabels,
 } from './selectors';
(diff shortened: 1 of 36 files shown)


MKodde (approved):
My main attention during review went to a good old wayf clickthrough. Found no issues here. The code changes also look good. No red flags. 

Note that I'm a non a11y/js/css expert. /disclaimer

PR #1101, opened 2021-03-31
Resolve Cypress to a maintained version

MKodde:
For now set the version to 5.6.0. We might want to upgrade to a more
modern version of cypress soon, but that would also entail
upgrading/replacing the cypress-plugin-snapshots and cypress-axe. Both
are incompatible with the latest Cypress.


MKodde (commented):
Thanks for your additions @Badlapje seeing you did as much work as I did, I hereby declare this PR approved

PR #1102, opened 2021-03-31
Fix missing titles in functional tests

Badlapje:
Prior to this change, the titles / names / displaytitle were missing in functional tests following the changes for the PT translation.

This change ensures they are restored.


MKodde:
I'm sorry for breaking this in the first place. I never ran the JS e2e tests due to them not running at that point. Thanks for finishing my work!

MKodde (approved):

PR #1103, opened 2021-04-01
Feature: configurable metadata

tvdijen:


MKodde:
Hi @tvdijen an interesting addition to the metadata. Your change-set is looking great! The tests need some tweaking. Some of the metadata test are somewhat treacherous. If you want to rubber duck, or need some input from my part, feel free to contact me on Slack.

MKodde:
The issue with the translator mock expectations can be resolved by changing the index to:

```
        $translator = $this->createMock(TranslatorInterface::class);
        $translator->expects($this->at(0))
            ->method('trans')
            ->with('suite_name')
            ->willReturn('test-suite');
        $translator->expects($this->at(1))
            ->method('trans')
            ->with('metadata_organization_name')
            ->willReturn('configuredOrganizationName');
        $translator->expects($this->at(2))
            ->method('trans')
            ->with('metadata_organization_displayname')
            ->willReturn('configuredOrganizationDisplayName');
        $translator->expects($this->at(3))
            ->method('trans')
            ->with('metadata_organization_url')
            ->willReturn('configuredOrganizationUrl');
```

tvdijen:
> But using %suiteName%, %supportUrl% and such does not, they are not interpolated.

That's strange because %suiteName% is being used all over the place in the translations-file...

thijskh:
Yes, this works when the string ends up being output in Twig, which is usually the case, but not for metadata.

tvdijen:
But the metadata is build based on a Twig template? What am I missing here? :/

thijskh:
Maybe they should be passed through the trans filter for it to work?

tvdijen:
Thanks, I hadn't had time to look into this!

thijskh (approved):
Tested, works fine

tvdijen:
We have noticed that the same issue with %placeholders% also applies to the md:Organization elements in the metadata. Those need to be passed through the translator too, or you will end up with invalid metadata (Shibboleth for example chokes on `<md:OrganizationURL>%supportURL%</md:Organization>` because it's not a valid URL.

PR #1105, opened 2021-04-06
Allow AuthnRequest based stepup requests

MKodde:
When the AuthnContextClassRefs contain valid LoA(s), then they are to be included in the step up decision process.

https://www.pivotaltracker.com/story/show/176960392

### DIFF
--- a/src/OpenConext/EngineBlock/Stepup/StepupDecision.php
+++ b/src/OpenConext/EngineBlock/Stepup/StepupDecision.php
@@ -79,8 +79,8 @@ class StepupDecision
         foreach ($pdpLoas as $loaId) {
             $this->pdpLoas[] = $loaRepository->getByIdentifier($loaId);
         }
-        foreach ($authnRequestLoas as $loaId) {
-            $this->authnRequestLoas[] = $loaRepository->getByIdentifier($loaId);
+        foreach ($authnRequestLoas as $loa) {
+            $this->authnRequestLoas[] = $loa;
         }
     }
 
--- a/src/OpenConext/EngineBlock/Stepup/StepupDecision.php
+++ b/src/OpenConext/EngineBlock/Stepup/StepupDecision.php
@@ -48,11 +48,6 @@ class StepupDecision
     private $spNoToken;
 
     /**
-     * @param IdentityProvider $idp
-     * @param ServiceProvider $sp
-     * @param Loa[] $authnRequestLoas
-     * @param Loa[] $pdpLoas
-     * @param LoaRepository $loaRepository
      * @throws InvalidStepupConfigurationException
      */
     public function __construct(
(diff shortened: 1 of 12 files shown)


src/OpenConext/EngineBlock/Stepup/StepupDecision.php
```diff
@@ -86,6 +86,10 @@ public function __construct(
 
     public function shouldUseStepup() : bool
     {
+        // If the highest level is 1, no step up callout is requred.
+        if ($this->getStepupLoa() && $this->getStepupLoa()->getLevel() === 1) {
+            return false;
+        }
         return ($this->spLoa || $this->idpLoa || count($this->authnRequestLoas) > 0 || count($this->pdpLoas) > 0);
```
thijskh, line 93: If we've already done getStepupLoa(), can we then make test in the `return` statement much simpler?

src/OpenConext/EngineBlock/Stepup/StepupDecision.php
```diff
@@ -72,11 +79,18 @@ public function __construct(
...
     }
 
     public function shouldUseStepup() : bool
     {
-        return ($this->spLoa || $this->idpLoa || count($this->pdpLoas) > 0);
+        // If the highest level is 1, no step up callout is requred.
+        if ($this->getStepupLoa() && $this->getStepupLoa()->getLevel() === 1) {
+            return false;
+        }
+        return ($this->spLoa || $this->idpLoa || count($this->authnRequestLoas) > 0 || count($this->pdpLoas) > 0);
```
Badlapje, line 93: this return statement seems to be identical to the if statement in `$this->allowNoToken`.  Can both not be abstracted into a method that can be called?

Badlapje: would also have the advantage of making the method name descriptive so it's easier to follow what's going on.

Badlapje (changes requested):
I haven't tested it functionally as i was unsure how to do so.  That said: nice work!

One small remark, other than that it's good to go i think.

src/OpenConext/EngineBlock/Stepup/StepupDecision.php
thijskh: Or remove it.

src/OpenConext/EngineBlock/Stepup/StepupDecision.php
MKodde: Good call!

src/OpenConext/EngineBlock/Stepup/StepupDecision.php
MKodde: Yes, I think you are right. A `null` from the `getStepupLoa` method equals the return statements logical comparison. We can simply return `true` at that point.

Badlapje (approved):
Nice work! :shipit:

PR #1108, opened 2021-04-08
Include Twig linting in CI pipeline

MKodde:
The test-integration was updated to include a Twig Lint action for the `theme` folder.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -92,10 +92,12 @@ jobs:
                   '
               env:
                   SYMFONY_ENV: ci
-            - name: Run frontend linting tests
+            - name: Run linting tests
               if: always()
               run: |
                   cd docker && docker-compose exec -T php-fpm.vm.openconext.org bash -c '
+                      echo -e "\nTwig lint\n"  && \
+                      php72 app/console lint:twig theme/ && \
                       cd theme && \
                       echo -e "\nNPM lint\n" && \
                       npm run lint
--- a/build.xml
+++ b/build.xml
@@ -60,6 +60,12 @@
         <echo message="OK"/>
     </target>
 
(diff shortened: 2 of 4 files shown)


.github/workflows/test-integration.yml
```diff
@@ -92,10 +92,12 @@ jobs:
                   '
               env:
                   SYMFONY_ENV: ci
-            - name: Run frontend linting tests
+            - name: Run linting tests
               if: always()
               run: |
                   cd docker && docker-compose exec -T php-fpm.vm.openconext.org bash -c '
+                      echo -e "\nTwig lint\n" && \
+                      php72 app/console lint:twig theme/
```
Badlapje, line 100: Maybe also add this in the package.json and/or the documentation?

Badlapje (changes requested):

.github/workflows/test-integration.yml
MKodde: Wilco, sir!

build.xml
```diff
@@ -60,6 +60,12 @@
         <echo message="OK"/>
     </target>
 
+    <target name="php-twig-lint" description="Perform syntax check of Twig templates">
+        <exec executable="app/console" failonerror="true">
+            <arg line="lint:twig theme/" />
+        </exec>
+    </target>
```
MKodde, line 67: @Badlapje ended up adding this to the build.xml instead of the too much unrelated packages.json npm scripts. Are you with me on this one? Or can you argue otherwise?

Badlapje: I like adding it here, but I'm not sure a front-ender who's theming will think to look here or know / understand how to get it working.  So that'd require including it in the theming documentation with some instructions on how to run it.  The advantage a package.json has is that it's self-documenting & known territory for frontenders.

Badlapje (approved):
❤️ this PR!

PR #1113, opened 2021-04-15
Use "account" instead of organisation

Badlapje:
Prior to this change, there was a forgotten instance of
%organisationName% in the english and dutch translation.

This change uses the "account" vernacular instad of the organisationName
 variable.

This PR replaces https://github
.com/OpenConext/OpenConext-engineblock/pull/1111 as made by @arnoutt who
noticed the typos.


arnout:
The description tags the wrong Arnout, it should be @arnoutt

Badlapje:
> The description tags the wrong Arnout, it should be @arnoutt

woops, rectified

MKodde (approved):
I see no harm in this.

LGTM!

PR #1114, opened 2021-04-15
Add a 'back to SP' button on the Stepup failed error pages

MKodde:
When required data is available, a back to SP button is rendered. Clicking this button sends a AuthnFailed SAML response back to the issuing SP.

https://www.pivotaltracker.com/story/show/176960606


Badlapje (approved):
Nice work!

PR #1115, opened 2021-04-19
Adjust php version in the twig linting script

Badlapje:
We used php72 in the twig linting script, which is not needed in the CI.
  Just plain old "php" wil do.


MKodde (approved):
Thanks for the catch!

PR #1116, opened 2021-04-19
Replace the hamburger and search icon

Badlapje:
Prior to this change, these were a seperate image meaning increased load
times.

This change replaces the icons with a font-awesome icon.

Issue on Pivotal: https://www.pivotaltracker.com/story/show/159457056

**Edit** Changed the Pivotal link


MKodde (changes requested):
Nice work, I haven't test-driven this yet. But already have some suggestions:

1. You can remove the hamburger and search svg's now am I right?
2. Do we need to verify the EN/NL/PT translations? Consider adding a google trans translation for the PT text.
3. Do the visual regression snapshots need updating?

Going to test this in browser now.

After some functional testing:
1. Do you plan to replace the spinner icon? Or would that be too great a regression?
2. I don't think the user.svg icon is used anywhere. Maybe remove it too?

Badlapje:
>     1. You can remove the hamburger and search svg's now am I right?

good one, done!
 
>     2. Do we need to verify the EN/NL/PT translations? Consider adding a google trans translation for the PT text.

I just asked arnout about them.  Added a portugese translation.

>     3. Do the visual regression snapshots need updating?

Good one!  Totally forgot about those.  😊 

>     1. Do you plan to replace the spinner icon? Or would that be too great a regression?

I recommended the fa-spinner to arnout , but as he correctly pointed out: the spinner was already replaced in the redesign.  So we left it in place.
 
>     2. I don't think the user.svg icon is used anywhere. Maybe remove it too?

Done.

theme/openconext/templates/modules/Authentication/View/Proxy/wayf.html.twig
```diff
@@ -18,7 +18,7 @@
         <h2 class="search-institutions">{{ 'search'|trans }}</h2>
         <input placeholder="{{ 'search'|trans }}" type="search" tabindex="0" class="mod-search-input active">
 
-        <button class="search">Search</button>
+        <button class="search"><span class="sr-only">Search</span><i class="fa fa-search" aria-hidden="true"></i></button>
```
MKodde, line 21: Totally missed this one in my first review round. Shouldn't the 'Search' text be a translation?

theme/openconext/translations/messages.pt.php
```diff
@@ -10,6 +10,7 @@
     // General
     'search'                    => 'Procure por uma %organisationNoun%...',
     'log_in_to'                 => 'Seleccione uma %organisationNoun% para se autenticar no serviço:',
+    'hamburger_screenreader'     => 'pular para o rodapé',
```
MKodde, line 13: Toalha de banho bem feita!

MKodde (commented):
Is it me, or did the screenshots not end up in your commit stage? Other then that! Things are looking up! I found a possible translation thingy.

Badlapje:
@MKodde for the screenshots: we don't screenshot the pages where these changes reside.  So nothing needs replacing there.  Added it to my todo to make that second visual test PR.

theme/openconext/translations/messages.pt.php
Badlapje: 😆

theme/openconext/templates/modules/Authentication/View/Proxy/wayf.html.twig
Badlapje: well spotted!

MKodde (approved):
Grand!

PR #1128, opened 2021-04-20
Add a global site notice

Badlapje:
Prior to this release there was no way to set a global site notice.

This release will add the ability to set a site notice via the parameters.yml file.  

More info can be found at the pivotal story: https://www.pivotaltracker.com/story/show/176815369

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -170,6 +170,10 @@ parameters:
     wayf.display_default_idp_banner_on_wayf: true
     wayf.default_idp_entity_id: https://default-idp.vm.openconext.org
 
+    ## Toggle display & content of global site notice
+    global.site_notice.show: false
+    global.site_notice.content: '<p>There is nothing wrong with your television set. <strong>Do not attempt to adjust the picture.</strong> We are controlling transmission. If we wish to make it louder, we will bring up the volume. If we wish to make it softer, we will tune it to a whisper. We will control the horizontal. We will control the vertical. We can roll the image, make it flutter. We can change the focus to a soft blur, or sharpen it to crystal clarity. <strong></p><p>For the next hour, sit quietly and we will control all that you see and hear.</strong> We repeat: There is nothing wrong with your television set. You are about to participate in a great adventure. You are about to experience the awe and mystery which reaches from the inner mind to... <a href="https://en.wikiquote.org/wiki/The_Outer_Limits_(1963_TV_series)">The Outer Limits</a>.</p>'
+
     ## Settings for detecting whether the user is stuck in a authentication loop within his session
     time_frame_for_authentication_loop_in_seconds: 60
     maximum_authentication_procedures_allowed: 5
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -173,7 +173,6 @@ parameters:
     ## Toggle display & content of global site notice
     global.site_notice.show: false
     global.site_notice.allowed.tags: '<a><u><i><br><wbr><strong><em><blink><marquee><p><ul><ol><dl><li><dd><dt><div><span><blockquote><hr><h2></h2><h3><h4><h5><h6>'
-    global.site_notice.content: '<p>There is nothing wrong with your television set. <strong>Do not attempt to adjust the picture.</strong> We are controlling transmission. If we wish to make it louder, we will bring up the volume. If we wish to make it softer, we will tune it to a whisper. We will control the horizontal. We will control the vertical. We can roll the image, make it flutter. We can change the focus to a soft blur, or sharpen it to crystal clarity. <strong></p><p>For the next hour, sit quietly and we will control all that you see and hear.</strong> We repeat: There is nothing wrong with your television set. You are about to participate in a great adventure. You are about to experience the awe and mystery which reaches from the inner mind to... <a href="https://en.wikiquote.org/wiki/The_Outer_Limits_(1963_TV_series)">The Outer Limits</a>.</p>'
 
     ## Settings for detecting whether the user is stuck in a authentication loop within his session
     time_frame_for_authentication_loop_in_seconds: 60
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -172,6 +172,7 @@ parameters:
 
     ## Toggle display & content of global site notice
     global.site_notice.show: false
+    global.site_notice.allowed.tags: '<a><u><i><br><wbr><strong><em><blink><marquee><p><ul><ol><dl><li><dd><dt><div><span><blockquote><hr><h2></h2><h3><h4><h5><h6>'
     global.site_notice.content: '<p>There is nothing wrong with your television set. <strong>Do not attempt to adjust the picture.</strong> We are controlling transmission. If we wish to make it louder, we will bring up the volume. If we wish to make it softer, we will tune it to a whisper. We will control the horizontal. We will control the vertical. We can roll the image, make it flutter. We can change the focus to a soft blur, or sharpen it to crystal clarity. <strong></p><p>For the next hour, sit quietly and we will control all that you see and hear.</strong> We repeat: There is nothing wrong with your television set. You are about to participate in a great adventure. You are about to experience the awe and mystery which reaches from the inner mind to... <a href="https://en.wikiquote.org/wiki/The_Outer_Limits_(1963_TV_series)">The Outer Limits</a>.</p>'
 
     ## Settings for detecting whether the user is stuck in a authentication loop within his session
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -43,8 +43,6 @@ return $overrides + [
     'note'                  => 'Note',
     'note_no_script'        => 'Since your browser does not support JavaScript, you must press the button below to proceed.',
     'unknown_organization_name' => 'Unknown',
-    'warning_allowed_html'  =>  '<a><u><i><br><wbr><strong><em><blink><marquee><p><ul><ol><dl><li><dd><dt><div><span><blockquote><hr><h2><h3><h4><h5><h6>',
-
 
         // Feedback
     'requestId'             => 'UR ID',
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -43,6 +43,7 @@ return $overrides + [
     'note'                  => 'Note',
     'note_no_script'        => 'Since your browser does not support JavaScript, you must press the button below to proceed.',
     'unknown_organization_name' => 'Unknown',
+    'site_notice'           => '<p>There is nothing wrong with your television set. <strong>Do not attempt to adjust the picture.</strong> We are controlling transmission. If we wish to make it louder, we will bring up the volume. If we wish to make it softer, we will tune it to a whisper. We will control the horizontal. We will control the vertical. We can roll the image, make it flutter. We can change the focus to a soft blur, or sharpen it to crystal clarity. <strong></p><p>For the next hour, sit quietly and we will control all that you see and hear.</strong> We repeat: There is nothing wrong with your television set. You are about to participate in a great adventure. You are about to experience the awe and mystery which reaches from the inner mind to... <a href="https://en.wikiquote.org/wiki/The_Outer_Limits_(1963_TV_series)">The Outer Limits</a>.</p>',
 
         // Feedback
     'requestId'             => 'UR ID',
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -43,8 +43,10 @@ return $overrides + [
     'note'                  => 'Note',
     'note_no_script'        => 'Since your browser does not support JavaScript, you must press the button below to proceed.',
     'unknown_organization_name' => 'Unknown',
+    'warning_allowed_html'  =>  '<a><u><i><br><wbr><strong><em><blink><marquee><p><ul><ol><dl><li><dd><dt><div><span><blockquote><hr><h2><h3><h4><h5><h6>',
 
-    // Feedback
+
+        // Feedback
     'requestId'             => 'UR ID',
     'identityProvider'      => 'IdP',
     'serviceProvider'       => 'SP',
--- a/library/EngineBlock/Corto/Module/Service/ProvideConsent.php
+++ b/library/EngineBlock/Corto/Module/Service/ProvideConsent.php
@@ -204,9 +204,6 @@ class EngineBlock_Corto_Module_Service_ProvideConsent
             $nameId->setFormat($serviceProvider->nameIdFormat);
         }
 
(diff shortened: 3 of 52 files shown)


languages/messages.en.php
```diff
@@ -43,8 +43,10 @@
     'note'                  => 'Note',
     'note_no_script'        => 'Since your browser does not support JavaScript, you must press the button below to proceed.',
     'unknown_organization_name' => 'Unknown',
+    'warning_allowed_html'  =>  '<a><u><i><br><wbr><strong><em><blink><marquee><p><ul><ol><dl><li><dd><dt><div><span><blockquote><hr><h2><h3><h4><h5><h6>',
```
MKodde, line 46: The parameters.yml might be a more suitable place to configure this wouldn't you say? I can't think why English readers should have a limited set of elements compared to the Dutch or Portuguese. Even consider hardcoding this and not making it configurable.

theme/base/templates/modules/Authentication/View/Proxy/Partials/Shared/site-notice.html.twig
```diff
@@ -0,0 +1,7 @@
+{% if showGlobalSiteNotice %}
```
MKodde, line 6: At this point you could use the Twig thingy to inform the template if the notice should be shown or not

theme/README.md
```diff
@@ -182,6 +182,8 @@ There are a number of feature flags which need to be supported by a theme in ord
...
 - cutoffPointForShowingUnfilteredIdps: unlike the previous flags which were booleans, this one is an integer.  When this flag is present & the user is not searching (so no value in the search field): there should only be idps shown when there are no more than the cutoff point.  So as an example: if there are 100 idps that would be shown normally, and the cutoff point is 50: the user should see no idps at all (so only the empty search field is shown).  You can test this by going to `/functional-testing/wayf?cutoffPointForShowingUnfilteredIdps=50`.
 - showConsentExplanation: whether to show a special explanation on the consent page.  There is currently no test endpoint for this.
+- showGlobalSiteNotice: whether to show a global site notice or not.  You can test this by going to `/functional-testing/wayf?showGlobalSiteNotice=1` or `/functional-testing/consent?showGlobalSiteNotice=1`
+- globalSiteNotice: if you want to add a siteNotice you can add the text for it here.  You can test this by going to `/functional-testing/wayf?showGlobalSiteNotice=1&globalSiteNotice=<text>` or `/functional-testing/consent?showGlobalSiteNotice=1&globalSiteNotice=<text>`
```
MKodde, line 186: My man! Thanks for updating this right away

theme/openconext/templates/modules/Authentication/View/Proxy/consent.html.twig
```diff
@@ -151,13 +153,13 @@
                 </table>
             </section>
 
-            {% if showConsentExplanation and consentSettings.consentExplanationIn(app.request.locale, spEntityId) %}
+{#            {% if showConsentExplanation and consentSettings.consentExplanationIn(app.request.locale, spEntityId) %}#}
```
MKodde, line 156: You removed this later I presume, If not please resolve this

theme/README.md
```diff
@@ -167,7 +167,7 @@ There are a lot of error pages.  To test all different kinds, you can use the ur
 - redirect page: `templates > modules > authentication > view > proxy > redirect.html.twig`.
 - spinner page: `templates > modules > authentication > view > proxy > form.html.twig`.  To test it disable the onload handler on the body-tag and go to your profile (or load the page without JS).
 - index.html.twig: `templates > modules > authentication > view > index > index.html.twig`.  You can use `https://engine.vm.openconext.org/` to develop the page.
-- cookie removal page: `templates > modules > authentication > view > identityprovider > remove-cookies.html.twig`.  You can use `https://engine.vm.openconext.org/authentication/idp/remove-cookies` to develop the page.
+- cookie removal page: `templates > modules > authentication > view > identityprovider > remove-cookies.html.twig`.  You can use `https://engine.vm.openconext.org/authentication/idp/remove-cookies` to develop the page.  The page is only accessible if you set the `wayf.remember_choice` parameter in `parameters.yml` to true.
```
MKodde, line 170: Thanks for amending this to the readme!

src/OpenConext/EngineBlockBundle/Controller/IndexController.php
```diff
@@ -44,6 +45,10 @@ public function __construct(Environment $twig, array $keyPairs)
      */
     public function indexAction()
     {
+        $settings = EngineBlock_ApplicationSingleton::getInstance()->getDiContainer();
```
MKodde, line 48: We don't want to litter our 'modern' Symfony application with stuff from the legacy 'library'. You can use the adapter solution for this. I think there are a couple of examples of this in the application. If you'd like we can have a look around together.

MKodde (changes requested):
You did a great job. I have some suggestions that hopefully makes this PR a little less involved.

app/config/parameters.yml.dist
```diff
@@ -170,6 +170,10 @@ parameters:
     wayf.display_default_idp_banner_on_wayf: true
     wayf.default_idp_entity_id: https://default-idp.vm.openconext.org
 
+    ## Toggle display & content of global site notice
+    global.site_notice.show: false
+    global.site_notice.content: '<p>There is nothing wrong with your television set. <strong>Do not attempt to adjust the picture.</strong> We are controlling transmission. If we wish to make it louder, we will bring up the volume. If we wish to make it softer, we will tune it to a whisper. We will control the horizontal. We will control the vertical. We can roll the image, make it flutter. We can change the focus to a soft blur, or sharpen it to crystal clarity. <strong></p><p>For the next hour, sit quietly and we will control all that you see and hear.</strong> We repeat: There is nothing wrong with your television set. You are about to participate in a great adventure. You are about to experience the awe and mystery which reaches from the inner mind to... <a href="https://en.wikiquote.org/wiki/The_Outer_Limits_(1963_TV_series)">The Outer Limits</a>.</p>'
```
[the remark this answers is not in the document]
Badlapje: @thijskh do you agree with the default site_notice content being a translatable string?  Makes sense to me, but that does mean there's two places where it needs to be configured for anyone wanting to show it?

languages/messages.en.php
Badlapje: Good one, adjusting!

library/EngineBlock/Corto/Module/Service/ProvideConsent.php
```diff
@@ -221,6 +224,8 @@ public function serve($serviceName, Request $httpRequest)
                 'nameIdIsPersistent' => $isPersistent,
                 'profileUrl' => $profileUrl,
                 'showConsentExplanation' => $identityProvider->getConsentSettings()->hasConsentExplanation($serviceProviderMetadata->entityId),
+                'showGlobalSiteNotice' => $showGlobalSiteNotice,
+                'globalSiteNotice' => $globalSiteNotice,
```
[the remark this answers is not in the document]
Badlapje: I created a twig helper modelled on the feedback.php helper.  That means the check can be contained in the site-notice.html.twig, which is perfect (all logic in one place).

src/OpenConext/EngineBlockBundle/Controller/IndexController.php
Badlapje: the twig helper actually resolved this too :-)

theme/openconext/templates/modules/Authentication/View/Proxy/consent.html.twig
Badlapje: resolved

theme/base/templates/modules/Authentication/View/Proxy/Partials/Shared/site-notice.html.twig
Badlapje: i ❤️ the twig extension ... so much cleaner that way 💯

app/config/parameters.yml.dist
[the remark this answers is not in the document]
Badlapje: > Given this is a feature toggle, may I suggest moving this config to around line 200 and renaming it in to something along the lines: `feature_show_global_site_notice: false`

I actually considered this.  However, it seems that those feature toggles are mainly concerning technical stuff (encrypting assertions, using the metadata api, ... ).  This toggle is more akin to the feature toggles we have that are higher up in parameters.yml such as "wayf.display_default_idp_banner_on_wayf" or "wayf.remember_choice", where we display parts of the frontend based on a toggle.  Given that fit, I've included it there.  Is that a reasoning you can follow @MKodde ?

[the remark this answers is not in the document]
Badlapje: fixed translations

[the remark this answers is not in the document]
MKodde: Thanks for your motivation, I can certainly live with that!

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/GlobalSiteNotice.php
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private $request;
+
+    public function __construct(
+        EngineBlock_ApplicationSingleton $application,
+        RequestStack $requestStack,
+        TranslatorInterface $translator
+    ) {
+        $this->application = $application;
+        $this->diContainer = $application->getDiContainer();
```
MKodde, line 56: I see down below you use the diContainer to load the `shouldDisplayGlobalSiteNotice` and `getAllowedHtmlForNotice` data from the container. This however is discouraged, you can directly pass those parameters to this service in the service definition. That way we are not dependent on the library (old EB) code.

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/GlobalSiteNotice.php
```diff
@@ -0,0 +1,112 @@
...
+    public function getAllowedHtmlForNotice(): string
+    {
+        return (string) $this->diContainer->getAllowedHtmlForNotice();
+    }
+
+    private function isTest(): bool
+    {
+        $uri = $this->request->getRequestUri();
+        return preg_match('/functional-testing/', $uri);
+    }
```
MKodde, line 111: I do not like this setup much. Business code should not be test aware in my opinion. Alternatively you could create a test stand in for the GlobalSiteNotice extension.

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/GlobalSiteNotice.php
```diff
@@ -0,0 +1,112 @@
...
+        if ($this->isTest()) {
+            return (bool) $this->request->get('showGlobalSiteNotice', false);
+        }
+
+        return (bool) $this->diContainer->shouldDisplayGlobalSiteNotice();
+    }
+
+    public function getGlobalSiteNotice(): string
+    {
+        if ($this->isTest()) {
```
MKodde, line 81: When the `isTest` portion is moved to a test stand-in you can also move the dummy text to that class, making this much cleaner.

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/GlobalSiteNotice.php
```diff
@@ -0,0 +1,112 @@
...
+            new TwigFunction('shouldDisplayGlobalSiteNotice', [$this, 'shouldDisplayGlobalSiteNotice']),
+            new TwigFunction('getGlobalSiteNotice', [$this, 'getGlobalSiteNotice']),
+            new TwigFunction('getAllowedHtmlForNotice', [$this, 'getAllowedHtmlForNotice']),
+        ];
+    }
+
+    public function shouldDisplayGlobalSiteNotice() : bool
+    {
+        if ($this->isTest()) {
+            return (bool) $this->request->get('showGlobalSiteNotice', false);
```
MKodde, line 73: Getting rid of the 'isTest' feature of this service will also result in the loss of the requestStack dependency.

MKodde (commented):
You are getting even closer now. Some things stood out to me:

1. The skeune warning notices did not look very skeune yet ;) See screenshots below.
2. The Engine homepage did not seem to show the site notice, nor did the auth debug landing page and the error/feedback page. Maybe that is intentional?
3. See some optimization suggestions for the awesome global site notice Twig Extension.

Screenshot_2021-04-28 OpenConext - Select an account to login to DisplayName
Screenshot_2021-04-28 OpenConext - Review your information that will be shared

src/OpenConext/EngineBlockBundle/Resources/config/services.yml
```diff
@@ -293,6 +293,15 @@ services:
         tags:
              - { name: 'twig.extension' }
 
+    engineblock.twig.extension.globalsitenotice:
+        class: OpenConext\EngineBlockBundle\Twig\Extensions\Extension\GlobalSiteNotice
+        arguments:
+            - "@engineblock.compat.application"
+            - "@request_stack"
+            - "@translator"
```
MKodde, line 301: Instead of injecting the application service, you can inject the parameters you are loading from the application singletons di-container later-on here directly. 

Parameters can be accessed suchlike "%locale%"

src/OpenConext/EngineBlockBundle/Twig/Extensions/Extension/GlobalSiteNotice.php
Badlapje: hah, learn something new every day 👍   Even better this way ^__^

Badlapje (commented):
>     1. The skeune warning notices did not look very skeune yet ;) See screenshots below.

Is it possible you had old CSS?  It looks like this for me:
image
image

>     2. The Engine homepage did not seem to show the site notice, nor did the auth debug landing page and the error/feedback page. Maybe that is intentional?

For the error/feedback page: intentional.  As for the auth debug page: I assumed it's something akin to the error page.  Not a good page to show the site-notice.  Not sure in what context it's shown to normal users, but i figured it's the same type as an error page so didn't show it there.

For the homepage: should be visible (but in material style as that's not been redesigned yet):
image

>     3. See some optimization suggestions for the awesome global site notice Twig Extension.

❤️ thx for the help!

Not sure why github made this comment a review, but here it is.

MKodde (approved):
I'm happy with the change now! My screenshots musthavebeen old css in  browsercache. 

The debug authentication page is a test feature built into the EngineBlock. Here you can send a dummy AuthnRequest through EngineBlock. The response of the IdP will be shown on the debug page. Here you can see which attributes where released among other things. Given the debug nature of this page, it's probably OK not to show the site notice there. On the other hand, why not? I'd say, ask the PM's what they want in this regard. While at it maybe ask them if the dummy texts are acceptable. I like them, but a more conservative lorem ipsum text might be more fitting? Again something for the PO's to decide upon.

PR #1129, opened 2021-04-26
Change the mail configuration to smtp in stead of Sendmail

quartje:
The result is the same, but it allows disabling the function "proc_open" in the PHP FPM configuration

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -25,7 +25,7 @@ open_conext_engine_block:
         eb.block_user_on_violation: "%feature_block_user_on_violation%"
 
 swiftmailer:
-    transport: sendmail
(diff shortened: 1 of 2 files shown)


app/config/config.yml
```diff
@@ -25,7 +25,7 @@ open_conext_engine_block:
         eb.block_user_on_violation: "%feature_block_user_on_violation%"
 
 swiftmailer:
-    transport: sendmail
+    transport: "smtp"
```
MKodde, line 28: Good to go in my opinion! I suppose you considered making this configurable and decided against that. Perfectly fine with me!

MKodde (approved):
Looking good. I like the additional flexibility this grants!

PR #1131, opened 2021-04-28
Transparent SAML error Response for MFA entities

MKodde:
When a SP request a AuthnContext and transparent_authn_context is enabled, we pass on to the IDP. There is no way for us to know the meaning or value for this AuthnContext. If a user can't use the requested AuthnContext, the SP should be informed so it can decide what to do.

When an SP is configured for transparant_authn_context, an NoAuthContext saml error from the IDP should be copied to the SP, so it can act accordingly.

https://www.pivotaltracker.com/story/show/173602383

### DIFF
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -415,6 +415,18 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
                     )
                 );
 
+                // If this is a transparent MFA authentication, we want to forward the error response
+                $originalRequest = $this->_server->findRequestFromRequestId($sspResponse->getInResponseTo());
+                $originalSpIssuer = $originalRequest->getIssuer()->getValue();
+                $mfaHelper = EngineBlock_ApplicationSingleton::getInstance()->getDiContainer()->getMfaHelper();
+                $isMfaTransparentSp = $mfaHelper->isTransparent($originalSpIssuer, $idpEntityId->getValue());
+                if ($isMfaTransparentSp) {
+                    // The AuthnFailed response is passed transparently back to the issuer
+                    $response = new EngineBlock_Saml2_ResponseAnnotationDecorator($sspResponse);
+                    $response->setIsTransparentErrorResponse(true);
+                    return $response;
+                }
+
                 $statusCodeDescription = $status['Code'];
                 if (isset($status['SubCode'])) {
                     $statusCodeDescription .= '/' . $status['SubCode'];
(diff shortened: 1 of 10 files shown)


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -415,6 +415,18 @@ public function receiveResponse($serviceEntityId, $expectedDestination)
                     )
                 );
 
+                // If this is a transparent MFA authentication, we want to forward the error response
+                $originalRequest = $this->_server->findRequestFromRequestId($sspResponse->getInResponseTo());
```
Badlapje, line 419: shouldn't this variable be named `$spResponse`?  If not, what does the extra s stand for?

tests/unit/OpenConext/EngineBlock/Service/MfaHelperTest.php
```diff
@@ -0,0 +1,115 @@
...
+        self::assertFalse($isTransparent);
+    }
+
+    private function createIdP($spEntityId, $isSpInMfaEntities, $isTransparent)
+    {
+        $idp = m::mock(IdentityProvider::class);
+        $mfaEntities = m::mock(MfaEntityCollection::class);
+        $mfaEntity = null;
+        if ($isSpInMfaEntities) {
+            if ($isTransparent) {
```
Badlapje, line 105: can be turned into a ternary operator, but i'm fine with it as is as well.

Badlapje (approved):
1 thing i'm not clear on and one nitpick (up to you whether it makes sense).  Code-wise this looks good to me!

library/EngineBlock/Corto/Module/Bindings.php
MKodde: ssp == simple saml php. It denotates we got this from the SSP library.

library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -90,6 +90,13 @@ public function serve($serviceName, Request $httpRequest)
         $application = EngineBlock_ApplicationSingleton::getInstance();
         $log = $application->getLogInstance();
 
+        if ($receivedResponse->isTransparentErrorResponse()) {
+            $log->info('Response contains NoPassive status code: responding with NoPassive status to SP');
```
MKodde, line 94: Ouch, forgot to update the log entry :blush:

library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -868,7 +878,11 @@ public function getReceivedRequestFromResponse(EngineBlock_Saml2_ResponseAnnotat
                 EngineBlock_Exception::CODE_NOTICE
             );
         }
+        return $this->findRequestFromRequestId($requestId);
+    }
 
+    public function findRequestFromRequestId(string $requestId): ?EngineBlock_Saml2_AuthnRequestAnnotationDecorator
+    {
```
MKodde, line 885: I dumbed down the original method to allow for retrieving the Request from a string. This allows getting the Request from the SSP library received 'inResponseTo' string.

phavekes (approved):

PR #1135, opened 2021-05-18
Ensure fallbacks in the error pages function correctly

Badlapje:
Prior to this change:
- the fallback of the unknown-service-provider had
a missing case: that for no idp name.  There was also a word missing
from the Dutch translation.
- there was no fallback for the received an invalid signed response error
- there was a missing fallback for no-idps-error
- the name of the fallbacks for missing required fields was incorrect
- the openconext theme had several missing fallbacks

This change adds the cases & fixes the translation.

Pivotal story at: https://www.pivotaltracker.com/story/show/176714319
Relevant comments at:
- https://www.pivotaltracker.com/story/show/176714319/comments/224272628
- https://www.pivotaltracker.com/story/show/176714319/comments/224272649

### DIFF
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -244,6 +244,7 @@ Your %organisationNoun% has denied you access to this service. You will have to 
     'error_received_invalid_response'       => 'Error - Invalid %idpName% SAML response',
     'error_received_invalid_response_no_idp_name'       => 'Error - Invalid %organisationNoun% SAML response',
     'error_received_invalid_signed_response'=> 'Error - Invalid signature on %idpName% response',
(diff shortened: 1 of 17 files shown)


MKodde (approved):
Looks solid! Have you considered adding some new visual regression tests for these scenarios? But that might be a slippery slope at this point?

Badlapje:
> Looks solid! Have you considered adding some new visual regression tests for these scenarios? But that might be a slippery slope at this point?

I have, but seems like overkill to me.  Visual tests work best if we use them to check the major design points.  Not the intricacies of the messages itself.  If we test for that, adding cypress tests would be better imo.

PR #1137, opened 2021-05-19
Add a consentHashService

Badlapje:
Prior to this commit, there was no service specifically for hashing attribute arrays for consent.

This commit adds such a service with both the old hashing algorithm and a stable hashing algorithm.

Pivotal ticket: https://www.pivotaltracker.com/story/show/176513931


library/EngineBlock/Corto/Model/Consent.php
```diff
@@ -37,6 +38,7 @@ class EngineBlock_Corto_Model_Consent
     private $_response;
     /**
      * @var array
+     * All attributes as an associative array.
```
MKodde, line 41: I'm all for this additional documentation. According to the phpdoc man pages, you should put them on the same line as the @var annotation. 

https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.var.pkg.html

library/EngineBlock/Corto/Model/Consent.php
```diff
@@ -52,21 +54,28 @@ class EngineBlock_Corto_Model_Consent
...
      */
     public function __construct(
         $tableName,
         $mustStoreValues,
         EngineBlock_Saml2_ResponseAnnotationDecorator $response,
         array $responseAttributes,
         EngineBlock_Database_ConnectionFactory $databaseConnectionFactory,
-        $amPriorToConsentEnabled
+        $amPriorToConsentEnabled,
+        $hashService
```
MKodde, line 78: DI FTW!

:1st_place_medal:

library/EngineBlock/Corto/Model/Consent.php
```diff
@@ -75,27 +84,33 @@ public function __construct(
         $this->_responseAttributes = $responseAttributes;
         $this->_databaseConnectionFactory = $databaseConnectionFactory;
         $this->_amPriorToConsentEnabled = $amPriorToConsentEnabled;
+        $this->_hashService = $hashService;
     }
 
-    public function explicitConsentWasGivenFor(ServiceProvider $serviceProvider) {
+    public function explicitConsentWasGivenFor(ServiceProvider $serviceProvider): bool
```
MKodde, line 90: Unless you are testing every method you are return type hinting here, I advise to be hesitant in changing all return types. As the application might rely on an unstable return type. Say bool|null. I personally tend to only update the methods I've actually changed or called in my code. 

If this all works, more power to us. Otherwise you might have to tweak them a little

library/EngineBlock/Corto/Model/Consent.php
```diff
@@ -179,16 +190,27 @@ private function _storeConsent(ServiceProvider $serviceProvider, $consentType)
...
-            $attributesHash = $this->_getAttributesHash($this->_responseAttributes);
+        if ($hasUnstableConsentHash) {
+            return true;
+        }
+
+        $stableConsentHash = $this->_getStableAttributesHash($this->_responseAttributes);
+        return $this->retrieveConsentHashFromDb($dbh, $serviceProvider, $consentType, $stableConsentHash);
+    }
 
+    private function retrieveConsentHashFromDb(PDO $dbh, ServiceProvider $serviceProvider, $consentType, $attributesHash): bool
```
MKodde, line 211: Nice touch moving this to a private method! Hideing this additional responsibility is a good first step! You could consider creating a Symfony Respository (accessible via the hashing service) containing these queries. Offloading as much of this business logic away from the old library is always a great win!

src/OpenConext/EngineBlock/Service/Consent/ConsentHashService.php
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        return sha1($hashBase);
+    }
+
+    /**
+     * Lowercases all array keys and values.
+     * Performs the lowercasing on a copy (which is returned), to avoid side-effects.
+     * @param $original array
+     * @return array
```
MKodde, line 59: No need for phpdoc annotations here. You can simply type hint the parameters and the return type. This also goes for the other methods.

src/OpenConext/EngineBlock/Service/Consent/ConsentHashService.php
```diff
@@ -0,0 +1,141 @@
...
+            $hashBase = serialize($sortedAttributes);
+        } else {
+            $noEmptyAttributes = $this->removeEmptyAttributes($lowerCasedAttributes);
+            $names = array_keys($noEmptyAttributes);
+            $sortedAttributes = $this->sortArray($names);
+            $hashBase = implode('|', $sortedAttributes);
+        }
+
+        return sha1($hashBase);
+    }
```
MKodde, line 53: This method could do with some cleaning up. 
1. Try to get rid of the else condition
2. Consider creating two private methods tasked with the creation of the bashbase
3. Awesome work!

src/OpenConext/EngineBlock/Service/Consent/ConsentHashService.php
```diff
@@ -0,0 +1,141 @@
...
+
+namespace OpenConext\EngineBlock\Service\Consent;
+
+final class ConsentHashService
+{
+    public function getUnstableAttributesHash($attributes, $mustStoreValues): string
+    {
+        $hashBase = null;
+        if ($mustStoreValues) {
+            ksort($attributes);
```
MKodde, line 27: A small optimization is to add impor statements for all functions used in the class. Consider doing that here

tests/unit/OpenConext/EngineBlock/Service/Consent/ConsentHashServiceTest.php
```diff
@@ -0,0 +1,446 @@
...
+            'urn:mace:dir:attribute-def:displayName' => ['John Doe'],
+            'urn:mace:dir:attribute-def:cn' => ['John Doe'],
+            'urn:mace:dir:attribute-def:sn' => ['Doe'],
+            'urn:mace:dir:attribute-def:eduPersonPrincipalName' => ['j.doe@example.com'],
+            'urn:mace:dir:attribute-def:givenName' => ['John'],
+            'urn:mace:dir:attribute-def:mail' => ['j.doe@example.com'],
+            'urn:mace:terena.org:attribute-def:schacHomeOrganization' => ['example.com'],
+            'urn:mace:dir:attribute-def:isMemberOf' => [
+                'urn:collab:org:vm.openconext.org',
+                'urn:collab:org:vm.openconext.ORG',
```
MKodde, line 68: I'm not certain attribute values are to be lower cased. As they might really be something different based on casing. Maybe you already checked this with PM's If thats the case ignore this. Otherwise please check this. 

I love the rest of the test by the way! Have you ran code coverage? To see if the entire hashing service is covered?

MKodde (changes requested):
A first review looks very promising! See some improvement suggestions below.

library/EngineBlock/Corto/Model/Consent.php
Badlapje: I changed _hasStoredConsent and _storeConsent, which both return bools always.  The other methods call those two and simply return what they return, meaning they also return bools.  So in spirit i followed the same rule you do 😄

src/OpenConext/EngineBlock/Service/Consent/ConsentHashService.php
Badlapje: done 👍

tests/unit/OpenConext/EngineBlock/Service/Consent/ConsentHashServiceTest.php
Badlapje: Lowercasing was an explicit requirement from the PM, so that's covered 😄 

I hadn't run code coverage.  Good idea!

library/EngineBlock/Corto/Model/Consent.php
Badlapje: Repository created.  As it now exists I've also moved the other DB logic there as well.

src/OpenConext/EngineBlock/Service/Consent/ConsentHashService.php
```diff
@@ -0,0 +1,176 @@
...
+     * Lowercases all array keys and values.
+     * Performs the lowercasing on a copy (which is returned), to avoid side-effects.
+     */
+    private function caseNormalizeStringArray(array $original): array
+    {
+        return unserialize(strtolower(serialize($original)));
+    }
+
+    /**
+     * Recursively sorts an array via the ksort function.  Performs the sort on a copy to avoid side-effects.
```
BackEndTea, line 109: Assuming all values in the array are scalars or arrays (not objects), then it won't have side effects.

Arrays in PHP are copy on write, meaning they are passed by ref as long as you make no updates, but i the array is updated then a copy will be used. (unless you explicitly pass it by ref):

https://3v4l.org/ks9hU

library/EngineBlock/Corto/Model/Consent.php
```diff
@@ -60,63 +57,72 @@ class EngineBlock_Corto_Model_Consent
...
         $this->_consentEnabled = $consentEnabled;
     }
 
-    public function explicitConsentWasGivenFor(ServiceProvider $serviceProvider)
+    public function explicitConsentWasGivenFor(ServiceProvider $serviceProvider): bool
     {
-        return !$this->_consentEnabled ||
-            $this->_hasStoredConsent($serviceProvider, ConsentType::TYPE_EXPLICIT);
+        $consent = $this->_hasStoredConsent($serviceProvider, ConsentType::TYPE_EXPLICIT);
+        return !$this->_consentEnabled || $consent->given();
```
thijskh, line 88: This means now queries (and hash calculations) will be done even if the consent feature toggle is off. This also goes for the other public methods that use this construct.

src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
```diff
@@ -161,4 +163,144 @@ public function deleteOneFor(string $userId, string $serviceProviderEntityId): b
...
+        }
+    }
+
+    /**
+     * @throws RuntimeException
+     */
+    public function storeConsentHash(array $parameters): bool
+    {
+        $query = "INSERT INTO consent (hashed_user_id, service_id, attribute_stable, consent_type, consent_date, deleted_at)
+                  VALUES (?, ?, ?, ?, NOW(), '0000-00-00 00:00:00')
```
thijskh, line 241: So it's stored as being deleted? Because a non-null value is being set for deleted_at?

library/EngineBlock/Corto/Model/Consent.php
```diff
@@ -128,98 +134,69 @@ protected function _getConsentUid()
...
         }
+
+        $parameters = [
+            $consentUuid,
+            $serviceProvider->entityId,
+            $this->_getAttributesHash($this->_responseAttributes),
+            $consentType,
+        ];
+
+        $hasUnstableConsentHash = $this->_hashService->retrieveConsentHash($parameters);
```
thijskh, line 194: I am wondering about the approach. This currently queries first for the new hash. If not found then for the old hash. And if also not found will present consent screen.

Why is this done over the following approach
- oldHash = _getAttributesHash()
- newHash = _getStableAttributesHash()
- SELECT * FROM consent where attributes_stable = :newHash OR attributes = :oldHash
so you always know in one query whether consent was given?

database/DoctrineMigrations/Version20220503092351.php
```diff
@@ -0,0 +1,30 @@
...
+ * 2. Changed the `attribute` column, has been made nullable
+ */
+final class Version20220503092351 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE consent ADD attribute_stable VARCHAR(80) NOT NULL, CHANGE attribute attribute VARCHAR(80) DEFAULT NULL');
```
thijskh, line 20: I'm not yet decided on the pros and cons of adding a new hash field next to the old field.

I believe the problem is solvable with just one database field. Not sure yet if that is preferable though.

library/EngineBlock/Corto/Model/Consent.php
MKodde: A very good point! Will change this to only search for previous consent when consent is enabled.

src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
MKodde: That has to do with the following.

The deleted_at column became part of the primary key, because we can now have multiple rows for a given user for a given service. Where one of those rows is an active consent row, and the rest are soft deleted versions of previously given consent. 

Having a dattime as part of a key requires it to be not nullable. But we do not want the column to be empty. Thats why a null-datetime is set for those values. DBAL will translate this to 'null'. But for the DBMS is satisfied. At least that's my understanding.

library/EngineBlock/Corto/Model/Consent.php
MKodde: Also, a good point! Having this in one query is much better. Will update this.

database/DoctrineMigrations/Version20220503092351.php
MKodde: I'm open for other suggestions! Having it in one single column will work too in my opinion. The 'OR' based query solution would work just as well on a single column.

baszoetekouw:
@MKodde @johanib I seems this issue (unstable consent hashes, causing too many consent screen because the orider of the attributes values received form the IdP changes between authentications) might become relevant again.  

What would be the best approach to revive this?  Try to bring this PR up to date, or start fresh implementing this using the ideas from this PR?

Issue #632, opened 2019-02-07
ARP being ran twice


tvdijen:
According to [this](https://github.com/OpenConext/OpenConext-engineblock/wiki/Attribute-Manipulations#openconext-sp) documentation, the processing order is:
first ARP, then AM.

It seems however, ARP is being ran after AM for the second time. Logging seems to confirm this as well (see first + last line):

`[2019-02-07 10:01:06] engineblock.INFO: Applying attribute release policy for https://acc-anvs.mavimcloud.com/saml2 {"session_id":"ea5utk4d5iskv2v94l440ofe05","request_id":"5c5bf3d294feb"} []
[2019-02-07 10:01:06] engineblock.NOTICE: AssertionConsumerServiceLocation 'https://acc-anvs.mavimcloud.com/Saml2/Acs' or ProtocolBinding '' were mentioned in request, but not both! Ignoring...  {"session_id":"ea5utk4d5iskv2v94l440ofe05","request_id":"5c5bf3d294feb"} []
[2019-02-07 10:01:06] engineblock.NOTICE: AssertionConsumerServiceLocation 'https://acc-anvs.mavimcloud.com/Saml2/Acs' or ProtocolBinding '' were mentioned in request, but not both! Ignoring...  {"session_id":"ea5utk4d5iskv2v94l440ofe05","request_id":"5c5bf3d294feb"} []
[2019-02-07 10:01:06] engineblock.INFO: Using internal binding for destination https://engine.sson.accsscict.rijksweb.nl/authentication/idp/provide-consent {"session_id":"ea5utk4d5iskv2v94l440ofe05","request_id":"5c5bf3d294feb"} {"url_params":{"EntityCode":"main","ServiceName":"provideConsentService","RemoteIdPMd5Hash":""}}
[2019-02-07 10:01:06] engineblock.INFO: Calling service 'provideConsentService' {"session_id":"ea5utk4d5iskv2v94l440ofe05","request_id":"5c5bf3d294feb"} []
[2019-02-07 10:01:06] engineblock.INFO: Using internal binding for destination https://engine.sson.accsscict.rijksweb.nl/authentication/proxy/processed-assertion {"session_id":"ea5utk4d5iskv2v94l440ofe05","request_id":"5c5bf3d294feb"} {"url_params":{"EntityCode":"main","ServiceName":"processedAssertionConsumerService","RemoteIdPMd5Hash":""}}
[...]

PR #1139, opened 2021-05-28
Apply ARP only once in the filter chain.

thijskh:
Currently EB applies the ARP two times in the filter chain. The second time it is just done because ePTI is first added unconditionally, then the entire ARP is applied to decide whether this is to be let through. So if you have an ARP that results in (a,b) in the first pass, the AddEduPeronTargetedId filter command makes the attribute set be (a,b,ePTI), the second ARP is applied and the attribute set is again (a,b).

This has several drawbacks:
- Anything added in the attribute manipulation is filtered out by the second ARP run. This usually means that you need to disable the ARP when having an AM. This has quite some consequences including confusing display in the Consent screen and in Dashboard.
- Attribute aggregation is configured in the ARP so you cannot use AA with an Attribute Manipulation now.
- Complexity and elevated mental load in understanding the filter chain and EB's processing for people working with EB as admins or developers.
- It does not seem optimal performance wise, but probably this is not that significant.

To resolve these issues, the decision on whether to add the ePTI is moved into the AddEduPersonTargetedId filter command. First, we added some tests to confirm the current ePTI behaviour (since these were not present yet). Then, the second ARP step is removed and in the AddEduPersonTargetedId command it is checked whether the ePTI is in the ARP. Only when it is, it is generated and released. The ePTI command is still well understandable and the logic of releasing the ePTI and its value is now kept together. The added tests still pass afterwards.

Closes: #632

### DIFF
--- a/library/EngineBlock/Corto/Filter/Command/AddEduPersonTargettedId.php
+++ b/library/EngineBlock/Corto/Filter/Command/AddEduPersonTargettedId.php
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+use SAML2\Constants;
+
 class EngineBlock_Corto_Filter_Command_AddEduPersonTargettedId extends EngineBlock_Corto_Filter_Command_Abstract
     implements EngineBlock_Corto_Filter_Command_ResponseAttributesModificationInterface
 {
@@ -39,6 +41,14 @@ class EngineBlock_Corto_Filter_Command_AddEduPersonTargettedId extends EngineBlo
             $this->_server->getRepository()
         );
 
+        // Find out if the EduPersonTargetedId is in the ARP of the destination SP.
+        // If the ARP is NULL this means no ARP = let everything through including ePTI.
+        // Otherwise only add ePTI if it's acutally in the ARP.
+        $arp = $destinationMetadata->getAttributeReleasePolicy();
+        if (!is_null($arp) && !$arp->hasAttribute(Constants::EPTI_URN_MACE)) {
+            return;
+        }
+
         // Resolve what NameID we should send the destination.
         $resolver = new EngineBlock_Saml2_NameIdResolver();
         $nameId = $resolver->resolve(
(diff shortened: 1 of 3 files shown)


Badlapje:
@thijskh ik heb deze ff gerebased zodat ie niet meer faalt op de security check

library/EngineBlock/Corto/Filter/Command/AddEduPersonTargettedId.php
```diff
@@ -48,8 +58,6 @@ public function execute()
             $this->_collabPersonId
         );
 
-        $this->_responseAttributes['urn:mace:dir:attribute-def:eduPersonTargetedID'] = array(
-            $nameId
-        );
+        $this->_responseAttributes[Constants::EPTI_URN_MACE] = [ $nameId ];
```
MKodde, line 61: :100:

library/EngineBlock/Corto/Filter/Command/AddEduPersonTargettedId.php
```diff
@@ -39,6 +41,14 @@ public function execute()
             $this->_server->getRepository()
         );
 
+        // Find out if the EduPersonTargetedId is in the ARP of the destination SP.
+        // If the ARP is NULL this means no ARP = let everything through including ePTI.
+        // Otherwise only add ePTI if it's acutally in the ARP.
+        $arp = $destinationMetadata->getAttributeReleasePolicy();
+        if (!is_null($arp) && !$arp->hasAttribute(Constants::EPTI_URN_MACE)) {
+            return;
+        }
```
MKodde, line 50: Makes sense!

MKodde (approved):
Looks like  a smart optimization! Disclaimer: I have not tested this in browser, this was a code review only.

thijskh:
Thanks for the review. The added behat tests already give quite some confidence. I will merge and we'll test further in our different environments.

PR #1142, opened 2021-06-02
Fix npm audit problems

Badlapje:
There are quite a lot of security problems being thrown up by npm audit. This PR is to fix those.

After trying out a better fix, we opted to disable the security check for the devdependencies.  Currently upgrading those is not compatible with maintaining support for IE8

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -130,7 +130,7 @@ jobs:
                       ./bin/securityChecker.sh && \
                       cd theme && \
                       echo -e "\nNPM Audit\n" && \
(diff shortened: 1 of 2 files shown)


MKodde (approved):
Good job. Please (if not already so) create a technical debt ticket on Pivotal stating we are currently only validating production dependencies. So we can fix this in the (hopefully near) future.

Badlapje:
> Good job. Please (if not already so) create a technical debt ticket on Pivotal stating we are currently only validating production dependencies. So we can fix this in the (hopefully near) future.

Done!

PR #1145, opened 2021-06-08
Check if there is a valid authentication in the AuthenticationState

Stephan-Kok:
In this pull request we have added functionality to check if there is a valid and completed authentication. 

In AuthenticationState.php there is a new public method 'isAuthenticated()' that returns true if there is any AuthenticationProcedure in the AuthenticationProcedureMap with a completed authentication.

After discussion with @MKodde we found that there is an Issue in the AuthenticationProcedure.php were the function 'hasBeenAuthenticated()' returns false when an authentication has been successful and returns true when an authentication has not yet been completed. For this reason there have been some changes to the core of the authentication process.

- The IDP will only be set in the AuthenticationProcedure when an authentication has been successful (This is done after the consent has been given by the user)
- Before the consent is checked, we validate the authentication request if the authentication exists and if it is not already completed

We have tested the code on our development environments and the authentication flow still works. However, since these are changes to the core of the authentication please test the changes thoroughly.

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
+++ b/src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
@@ -49,7 +49,7 @@ final class AuthenticationState implements AuthenticationStateInterface
      * @return void
      * @throws AssertionFailedException
      */
-    public function startAuthenticationOnBehalfOf($requestId, Entity $serviceProvider)
+    public function startAuthenticationOnBehalfOf($requestId, Entity $serviceProvider): void
     {
         Assertion::string($requestId, 'The requestId must be a string (XML ID) value');
         $currentAuthenticationProcedure = AuthenticationProcedure::onBehalfOf($serviceProvider);
@@ -78,12 +78,14 @@ final class AuthenticationState implements AuthenticationStateInterface
     }
 
     /**
+     * Validates if an request can be found in session and if the request is not already authenticated
+     *
      * @param string $requestId
-     * @param Entity $identityProvider
      * @return void
      * @throws AssertionFailedException
+     * @throws LogicException
      */
-    public function authenticatedAt($requestId, Entity $identityProvider)
+    public function validateAuthenticationRequest($requestId): void
     {
         Assertion::string($requestId, 'The requestId must be a string (XML ID) value');
 
@@ -92,28 +94,8 @@ final class AuthenticationState implements AuthenticationStateInterface
         if ($currentRequest === null) {
             throw new LogicException(
                 sprintf(
-                    'The requested authentication procedure with requestId "%s" couldn\'t be found in the session storage.',
-                    $requestId
-                )
-            );
-        }
-
-        $currentRequest->authenticatedAt($identityProvider);
-    }
-
-    /**
-     * @param string $requestId
-     * @return void
-     * @throws AssertionFailedException
-     */
-    public function completeCurrentProcedure($requestId)
-    {
-        Assertion::string($requestId, 'The requestId must be a string (XML ID) value');
-        $currentRequest = $this->authenticationProcedures->find($requestId);
-        if ($currentRequest === null) {
-            throw new LogicException(
-                sprintf(
-                    'The requested authentication procedure with requestId "%s" couldn\'t be found in the session storage in order to complete.',
+                    'The requested authentication procedure with requestId "%s" couldn\'t be found in the ' .
+                    'session storage.',
                     $requestId
                 )
             );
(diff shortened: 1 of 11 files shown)


src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -92,33 +94,44 @@ public function authenticatedAt($requestId, Entity $identityProvider)
...
                 )
             );
         }
 
-        $currentRequest->authenticatedAt($identityProvider);
+        if ($currentRequest->hasBeenAuthenticated()) {
+            throw new LogicException(
+                sprintf(
+                    'The requested authentication procedure with requestId "%s" has already been authenticated.',
+                    $requestId
```
thijskh, line 108: I'm worried this makes us robust. We currently purposefully allow someone to hit the 'reload' or 'back' button and reposting an assertion. I wonder whether doing this would now create an unhandled exception for the user?

tests/unit/OpenConext/EngineBlockBundle/Authentication/AuthenticationStateTest.php
```diff
@@ -46,7 +46,7 @@ public function an_authentication_procedure_cannot_be_authenticated_if_it_has_no
         $this->expectException(LogicException::class);
         $this->expectExceptionMessage('The requested authentication procedure with requestId "_00000000-0000-0000-0000-000000000000" couldn\'t be found in the session storage.');
 
-        $authenticationState->authenticatedAt('_00000000-0000-0000-0000-000000000000', $identityProvider);
+        $authenticationState->validateAuthenticationRequest('_00000000-0000-0000-0000-000000000000', $identityProvider);
```
MKodde, line 49: The IdP parameter can be dropped here

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -92,33 +94,44 @@ public function authenticatedAt($requestId, Entity $identityProvider)
...
     /**
+     * Completes the authentication procedure and sets the authenticated IDP and time
+     *
      * @param string $requestId
+     * @param Entity $identityProvider
      * @return void
      * @throws AssertionFailedException
      */
-    public function completeCurrentProcedure($requestId)
+    public function completeCurrentProcedure($requestId, Entity $identityProvider): void
```
MKodde, line 122: Nit: As we are now able to write php7 code. Consider adding a `string` type definition for the `requestId` and remove the `@param` and `@retrun` annotations from the PHPDoc.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
```diff
@@ -128,6 +141,22 @@ public function completeCurrentProcedure($requestId)
...
         }
 
+        $currentRequest->authenticatedAt($identityProvider);
         $currentRequest->completeOn(new DateTimeImmutable);
     }
+
+    /**
+     * Validates if the current session contains a valid authentication
+     *
+     * @return bool
```
MKodde, line 151: Nit: feel free to remove this @return annotation as the return type definition is good enough for us. We only created the PHPDoc for IDE autocomplete features. Not for actual PHPDoc generation.

MKodde (changes requested):
Nice work! I share Thijs' concern about the increased strictness. And I found some small nitpicks in your changes.

src/OpenConext/EngineBlockBundle/Authentication/AuthenticationState.php
Stephan-Kok: Hopefully this explains my thought process. 

**Current situation:**
we would 'authenticate' the user before showing the consent screen. Meaning:
- Fail if there is no request for the authenticationProcedures
- Setting the identityProvider in the AuthenticationProcedure '$currentRequest->authenticatedAt($identityProvider);'. 

After consent has been given and the authentication is complete we would complete the authentication procedure by setting the completion on date. Meaning:
- Fail if there is no request for the authenticationProcedures
- Fail if the identityProvider is not set '$currentRequest->hasBeenAuthenticated();'
	- This would throw the error 'The requested authentication procedure with requestId "%s" has already been authenticated.' but would actually mean: 'The requested authentication procedure is not yet authenticated'. Since it will be thrown if no identityProvider has been set yet.
- Complete the authentication '$currentRequest->completeOn(new DateTimeImmutable);'

**Suggested change:**
Before consent screen:
- Fail if there is no request for the authenticationProcedures
- Fail if the authenticationProcedure is already authenticated

After consent screen: (on the already existing command // Finally, mark the authentication procedure as being complete.)
- Fail if there is no request for the authenticationProcedures
- Fail if the authenticationProcedure is already authenticated
- Complete authentication by setting both the identityProvider and completeOn date

This way there is a clear point when a user is authenticated.


As to your question towards making this less robust, I do not see any changes with the previous implementation. 

When pressing the back button the session would have been lost either way with the implementation of '$this->_processingStateHelper->clearStepByRequestId($receivedRequest->getId());' in ProcessedAssertionConsumer.php. 

If you had meant before pressing consent, then you can still authenticate at different IdentityProviders since a 'startAuthenticationOnBehalfOf($requestId, Entity $serviceProvider)' creates new AuthenticationProcedures each time. The only change I am suggesting is to fix the logic behind a single AuthenticationProcedure; to prevent that there is a AuthenticationProcedure in the AuthenticationProcedureMap that is marked as 'authenticated' and is not actually completed.

thijskh: Thanks for the explanation. If I rePOST the assertion from the IdP, would that still work?

Stephan-Kok: You can still press refresh and rePOST assestions on the `authentication/sp/consume-assertion` endpoint. It is also possible to go back and select another IDP and authenticate with that IDP.

thijskh: The following scenario now fails with "engineblock.ERROR: The requested authentication procedure with requestId "CORTOc3af55c6a21f8b9217fdb8ba6e44bf1c16c1df70" has already been authenticated":

1. Start authentication, select IdP, log in at IdP, be presented consent
2. Accept consent and continue to SP
3. Press "back" on SP and re-post assertion to consume-assertion.

thijskh (commented):
Thanks. I think we are fine now. I will give it some real life testing to confirm it.

Was just wondering why some methods have been moved around without a visible change in their content.

When approved I think we should squash the commits to avoid changing something and reverting it again later as part of the same change.

thijskh (approved):
Yes, seems good now. Would be nice to squash before merging, and also if we can avoid moving around methods without actual changes in their code?

thijskh:
Will you be squashing the PR and looking into the methods that were just moved around? Otherwise it's good to go I think.

Stephan-Kok:
I squashed the PR, updated it with the current master and validated the changes on my local machine.

PR #1152, opened 2021-07-19
Add feature for attribute filter

Jong-Vincent:
In this pull request we have added the possibility to define a set of attributes which are always filtered from the attributes that are sent by an Identity Provider.

These attributes are subsequently also not shown in the consent screen and are not propagated to the Service Provider.

This can for example be useful when it is required to filter certain attributes for every Service Provider. For example in our case we  require certain attributes from the IdP but these attributes should never be sent to any Service Provider.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -24,6 +24,7 @@ open_conext_engine_block:
         eb.run_all_manipulations_prior_to_consent: "%feature_run_all_manipulations_prior_to_consent%"
         eb.block_user_on_violation: "%feature_block_user_on_violation%"
         eb.enable_sso_notification: "%feature_enable_sso_notification%"
+        eb.feature_filter_attributes: "%feature_filter_attributes%"
 
 swiftmailer:
     transport: "%mailer_transport%"
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -212,6 +212,7 @@ parameters:
     feature_api_deprovision: true
     feature_run_all_manipulations_prior_to_consent: false
     feature_block_user_on_violation: false
+    feature_filter_attributes: false
 
     ##########################################################################################
     ## PROFILE SETTINGS
@@ -265,3 +266,11 @@ parameters:
     sso_notification_encryption_key: <xxx>
     ## The encryption key salt used to decrypt the SSO notification
     sso_notification_encryption_key_salt: <xxx>
+
+    ##########################################################################################
+    ## FILTER ATTRIBUTES SETTINGS
+    ##########################################################################################
+    ## The attribute names added to this list are filtered from the list of attributes sent
+    ## by the IdP before showing the consent screen and sent to the SP
+    filter_attributes:
+        - example
--- a/library/EngineBlock/Application/DiContainer.php
+++ b/library/EngineBlock/Application/DiContainer.php
@@ -533,4 +533,12 @@ class EngineBlock_Application_DiContainer extends Pimple
     {
         return $this->container->get('engineblock.service.sso_notification');
     }
+
+    /**
+     * @return array
+     */
+    public function getFilterAttributes(): ?array
+    {
+        return $this->container->getParameter('filter_attributes');
+    }
 }
--- a/library/EngineBlock/Corto/Filter/Command/FilterAttributes.php
+++ b/library/EngineBlock/Corto/Filter/Command/FilterAttributes.php
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * Copyright 2021 Stichting Kennisnet
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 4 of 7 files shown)


thijskh:
Thanks. I'd like to understand the use case better though.

Currently EB works like this:
- Accepts any attribute from the IdP
- Has an Attribute Release Policy per SP.
- Will only ask for consent for the attributes enabled in the SPs ARP
- Will only release the same attributes in the SPs ARP to the SP

So in this scenario any auxiliary attribute sent by the IdP is never encountered in Consent screens or sent to any SP. Unless it's added to an ARP. I wonder why the additional global filtering is needed for.

Jong-Vincent:
Hi @thijskh, we require an employee number from the IdP to construct a uid. The uid is then sent to the SP. However, the employee number itself should not be sent to **any** SP. I believe this is the part that is causing the confusion (I will update it in the original text).

thijskh:
Given the following platform:

* Engineblock with 1 idp and 2 sp’s:

* IdP which releases:
** employeenr
** givenname
** surname
** mail

* SP1 with ARP (uid, givenname, surname)
* SP2 with ARP (mail)

When logging into SP1, user sees consent for uid, givenname, surname and these are released to the SP.

When logging into SP2, user sees consent for mail, and this is released to the SP.

In this setup employeenr is nowhere visible in consent or released to any SP. So no global filter is needed.

How would you need to change the setup to introduce a need for the filtering feature?

Jong-Vincent:
In our case we also have a number of Service Providers which do not have an ARP and should also not receive the employee number. Without global filtering we would be forced to create an ARP for all Service Providers for an attribute which is not part of that SP's ARP.

tvdijen:
You could unset the attribute in Attribute Manipulation, but you should really consider setting an ARP for those service providers as well... Any privacy officer will have your head for not applying the "principle of least privilege"

thijskh:
Why are there SPs without an ARP? What is the idea behind that?

Jong-Vincent:
Perhaps a clarification is in order first:

In our implementation, 'no ARP' means that a Service Provider receives a limited and predefined set of attributes by default. This set of attributes is also included in our agreement with the Identity Providers. Strictly speaking, there is therefore a basic ARP that applies by default to all Service Providers.

In this case, 'no ARP' does not mean that all attributes are released to a Service Provider.

The reason we want to exclude the employeeNumber attribute from propagation to Service Providers is because, along with the NameID, it forms the basis of the hashed uid that is released to Service Providers. As an extra layer of protection, we therefore do not expose the employeeNumber attribute. Because we apply this global filter mechanism, the employeeNumber attribute cannot be accidentally included in an ARP.

Finally, we provide a publicly available and transparent reference Service Provider to whom all attributes are released. This Service Provider implementation is used in certain cases for testing and support purposes. For this, all attributes are always released (literally the only Service Provider without an ARP). However, we don't want to show the employeeNumber attribute in this application, because, as mentioned above, this attribute is also the basis for the hashing of the uid.

tvdijen:
>> In this case, 'no ARP' does not mean that all attributes are released to a Service Provider.

The 'No ARP' checkbox does implicate this.. IDPs may release more than you're asking for, even accidentally.. If you in turn release them to an SP, as a data processor you are responsible.. My suggestion would be to explicitly set an ARP for the attributes agreed on by the agreement.. The testing service provider is within your control, or so it seems, so that's not an issue.. You can use Attribute Manipulation to filter out anything unwanted..

That said, I'm not a fan of this PR.. Adds complexity to something that's already possible

thijskh:
Hi,

- The "basic ARP" SPs will still have an ARP so will not receive this employeeID attribute in any case.
- If the attribute is to be never released to an SP, you just not add it to the attribute schema known by Manage. This means that there is no way to add this attribute to any ARP. Also not accidentally. The option is just not there.
- For the special case SP that really has "no ARP" applied (this is similar to our SP Profile) you can either:

1. accept this (if you want a demo SP that shows all attributes, maybe indeed it should show all attributes)
2. solve it in the SP itself (filter it out there)
3. as TIm suggests, solve the special case in an AM instead of introducing a global feature for what is in essence just one special SP

Jong-Vincent:
First of all, thank you both for the considerations and suggestions. 

In our current implementation of Engineblock we still see a use-case for this filter as an added layer of protection which lowers the risk if e.g. by any chance configuration mistakes are made in Manage. We will explore the usage of default, minimal ARPs as well as prohibiting "No ARP"-functionality to be submitted to a production environment (which is currently not validated). We understand your hesitation to go forward with this PR since it indeed adds some level of additional complexity, which for your own described usage does not add much value. We will keep this feature in our fork for now. Feel free to close this PR.

thijskh:
I still do not understand "any chance configuration mistakes are made in Manage" because this mistake is just not possible to make. The option is not offered.

Some kind of validation that indeed production SPs always have an ARP would be interesting to pursue. We now check that out of band in a cron process against the Manage API which checks that all production SPs have either: an ARP or an AM or are on the allowlist of no-ARP SPs (= only special cases: profile, dashboard, trusted proxies). The "or AM" may be less needed after 6.6.3 is released.

PR #1153, opened 2021-07-21
Make the original AuthnRequest available to attribute manipulations.

thijskh:
The Attribute Manipulations could use any items from the IdP, SP
metadata and prepared response, but could not read the AuthnRequest.
Adding `$requestObj` as an input adds a generic way to access information
from the request. It is (obviously) read only.

We're adding it only in the object form to mirror how the response is
passed, but do not convert it to any legacy format; that seems
unnecessary for a new feature.

There's a behat test that verifies that the AM can indeed read something
from the `$requestObj` variable - the AM itself in this scenario makes
not much sense but shows that we can read something from the object and
do something with it.


Badlapje (approved):
Looks good to me.  Nice work!

PR #1154, opened 2021-07-22
Unset id field which is autogenerated and should not be added to the SQL statement.

tbeekman:
The id field is added to the object which is used to construct the update or insert statements (using a NULL value on inserts). MariaDB understands that the ID is auto-generated, and ignores the field. However, we are migrating towards PostgreSQL which does not automatically ignore this field. This PR proposes to explicitly unset the ID field to prevent issues with databases other than MariaDB/MySQL without affecting the current inner workings. We will provide additional documentation in a later stage describing how Engineblock can be used with PostgreSQL instead of MariaDB.


pablothedude (approved):
That should do the trick.

thijskh:
Thanks!

PR #1155, opened 2021-07-26
Add configuration options for requested attributes in proxy SP metadata

Jong-Vincent:
Currently an IdP based on SimpleSAMLphp (v1.19.1) sends a limited list of attributes to an SP which defines RequestedAttributes, namely only the attributes which are defined in the RequestedAttributes. This limits the use of attribute conditions within Engineblock since the min condition, which adds the attribute to RequestedAttributes, can not be used. 
We have added an option to allow control over the RequestedAttribute list added to the proxy SP metadata of Engineblock where options 'all', 'required' and 'none' are supported. Default has been set to 'all' which is how currently Engineblock works. This may turn out to be a temporary solution if this is handled in SimpleSAMLphp but we wanted to provide this all the same.

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -54,6 +54,10 @@ parameters:
         - http
         - https
 
+    ## Add RequestedAttributes to the AttributeConsumingService of the SP Proxy metadata of Engineblock, default is all
+    ## Options are 'all' (optional and required attributes), 'required' (only required attributes) or 'none'
+    metadata_add_requested_attributes: all
+
     ##########################################################################################
     ## PHP SETTINGS
     ##########################################################################################
--- a/src/OpenConext/EngineBlock/Xml/MetadataRenderer.php
+++ b/src/OpenConext/EngineBlock/Xml/MetadataRenderer.php
@@ -71,6 +71,10 @@ class MetadataRenderer
      * @var LanguageSupportProvider
      */
     private $languageSupportProvider;
+    /**
+     * @var string
+     */
+    private $addRequestedAttributes;
 
     public function __construct(
         LanguageSupportProvider $languageSupportProvider,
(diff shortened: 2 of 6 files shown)


thijskh (changes requested):
Thanks! It looks good but I would expect tests to be added for each of the three variants of the setting to confirm that it works and remains working.

tvdijen:
Why not fix this in SSP?  I believe the metadata-converter shouldn't worry about RequestedAttributes and ignore them.. It's an internal SSP feature to have an ARP-like feature...

thijskh:
@tvdijen Is SSP the only consumer with this issue? If it also affects other products (e.g. Shibboleth) it might make more sense to change it in EB.

tvdijen:
I'm not familiar with Shibboleth IDPs and how they process SP-metadata to be honest...

Right now I manually edit the metadata-template to remove RequestedAttributes completely, so this change would actually make my life easier.
I still think, however, that SSP shouldn't convert RequestedAttributes to an implicit ARP

Jong-Vincent:
We are in favor of solving this in SSP as well but as @thijskh suggested it might also affect other products
such as Shibboleth. Unfortunately we do not have enough experience with Shibboleth as an IdP, we know that ADFS does not experience this issue. For the time being, we may want to implement this PR and reconsider it when SSP does not convert RequestedAttributes to an implicit ARP. We are more than happy to support the implementation of this feature in SSP if required.

@thijskh For this PR I have also added unit tests to verify the functionality.

thijskh (approved):
Thanks for adding the tests. I agree with you re fixing this in SSP. We can merge this now and can always take it out later if SSP does it differently _and_ this new version is sufficiently wide deployed at IdPs, i.e. it's no longer needed. It does not hurt much to add it now.

PR #1157, opened 2021-08-18
Reset attributeValueTypes after an Attribute Manipulation

thijskh:
AttributeValueTypes is an array contains the types (string, bool, int..) for each value of all released attributes in the attributes array. The ARP ensures the two arrays remain consistent after filtering out some values. However, the Attribute Manipulation also has the option to change the released attributes in the `$attributes` array passed to it. AM's do however not have a way to change the corresponding list of types (it is not available to the AM code).

If this is not done, in the case that the resulting set has less values than the original set after ARP, one gets [an Exception from the SAML2 library](https://github.com/simplesamlphp/saml2/blob/04d0ed6c5e51cbb328afe1bdcba2cc7bd84d7e87/src/SAML2/Assertion.php#L1690) that there are more valuetypes than values.

Earlier, this would always work because after AM the ARP would be reapplied which kind of sorted out the attribute types. Not entirely correctly btw, more by chance (the number of valuetypes is adjusted to the number of values - regardless), but in practice this nearly always went well because we never deal with anything but `xsi:string`. When the second ARP step was removed, this reordering of the value types was also not present anymore and the saml2 Exception appeared in the case of an IdP releasing multivalued attribute which was by an AM reduced to less (but not 0) values.

The solution is (and this makes it actually more correct than it was), is that if anything changed about the attributes after AM, we just reset the valuetypes. This triggers a feature in the saml2 library [that if there are no types](https://github.com/simplesamlphp/saml2/blob/04d0ed6c5e51cbb328afe1bdcba2cc7bd84d7e87/src/SAML2/Assertion.php#L1712), the library will assign the correct types based on the PHP types of the values. This means that you can even decide in the AM what types you'd want by setting $attributes correctly.

The approach is especially future proof because the 5.x branch of saml2 will do away with valuetypes altogether and will always use the PHP types of the attribute values, the behaviour we now trigger explicitly will then be the only behaviour available.

The issue was not discovered in development because the functional test suite lacked tests for multivalued attributes being reduced to lesser values, both by ARP and (in this case) by an AM. This PR first adds such tests. The testsuite fails, which shows the issue we encountered being present. After applying the commit with the fix, the testsuite passes again.

### DIFF
--- a/library/EngineBlock/Corto/Filter/Command/RunAttributeManipulations.php
+++ b/library/EngineBlock/Corto/Filter/Command/RunAttributeManipulations.php
@@ -21,7 +21,8 @@
  */
 class EngineBlock_Corto_Filter_Command_RunAttributeManipulations extends EngineBlock_Corto_Filter_Command_Abstract
     implements EngineBlock_Corto_Filter_Command_ResponseModificationInterface,
-    EngineBlock_Corto_Filter_Command_ResponseAttributesModificationInterface
+    EngineBlock_Corto_Filter_Command_ResponseAttributesModificationInterface,
+    EngineBlock_Corto_Filter_Command_ResponseAttributeValueTypesModificationInterface
 {
     const TYPE_SP  = 'sp';
     const TYPE_REQUESTER_SP = 'requester-sp';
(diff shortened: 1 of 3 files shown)


Badlapje (approved):
Very clear explanation of the problem being solved an how it was solved 💯

PR #1160, opened 2021-09-20
Support removal of consent by collab person id and SP entity id

MKodde:
An API endpoint is created to remove a single consent row based on a user id and a SP entity id. This allows the profile end user to retract consent for services they have previously granted consent to.

https://www.pivotaltracker.com/story/show/179490276

### DIFF
--- a/library/EngineBlock/User.php
+++ b/library/EngineBlock/User.php
@@ -59,7 +59,8 @@ class EngineBlock_User
     {
         $pdo = $this->_getDatabaseConnection();
         $query = 'SELECT service_id FROM consent
-                    WHERE hashed_user_id = ?';
+                    WHERE hashed_user_id = ?
+                    AND ISNULL(deleted_at)';
         $parameters = array(
             sha1($this->getUid())
         );
--- a/library/EngineBlock/User.php
+++ b/library/EngineBlock/User.php
@@ -60,7 +60,7 @@ class EngineBlock_User
         $pdo = $this->_getDatabaseConnection();
         $query = 'SELECT service_id FROM consent
                     WHERE hashed_user_id = ?
-                    AND ISNULL(deleted_at)';
+                    AND deleted_at IS NULL';
         $parameters = array(
             sha1($this->getUid())
         );
--- a/src/OpenConext/EngineBlock/Service/ConsentService.php
+++ b/src/OpenConext/EngineBlock/Service/ConsentService.php
@@ -96,11 +96,16 @@ final class ConsentService implements ConsentServiceInterface
 
     public function deleteOneConsentFor(CollabPersonId $id, string $serviceProviderEntityId): bool
     {
+        $collabPersonId = $id->getCollabPersonId();
         try {
-            return $this->consentRepository->deleteOneFor($id->getCollabPersonId(), $serviceProviderEntityId);
+            return $this->consentRepository->deleteOneFor($collabPersonId, $serviceProviderEntityId);
         } catch (Exception $e) {
             throw new RuntimeException(
-                sprintf('An exception occurred while removing consent for a service provider("%s")', $e->getMessage()),
+                sprintf(
+                    'An exception occurred while removing consent for a service provider("%s") and user ("%s").',
+                    $serviceProviderEntityId,
+                    $collabPersonId
+                ),
                 0,
                 $e
             );
--- a/src/OpenConext/EngineBlock/Service/ConsentService.php
+++ b/src/OpenConext/EngineBlock/Service/ConsentService.php
@@ -23,9 +23,11 @@ use OpenConext\EngineBlock\Authentication\Dto\Consent;
 use OpenConext\EngineBlock\Authentication\Dto\ConsentList;
 use OpenConext\EngineBlock\Authentication\Model\Consent as ConsentEntity;
 use OpenConext\EngineBlock\Authentication\Repository\ConsentRepository;
+use OpenConext\EngineBlock\Authentication\Value\CollabPersonId;
 use OpenConext\EngineBlock\Exception\RuntimeException;
 use OpenConext\Value\Saml\EntityId;
 use Psr\Log\LoggerInterface;
+use function sprintf;
 
 final class ConsentService implements ConsentServiceInterface
 {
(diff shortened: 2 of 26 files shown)


src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
```diff
@@ -99,4 +100,31 @@ public function deleteAllFor($userId)
...
+                ]
+            );
+            return $result->rowCount() > 0;
+        } catch (DBALException $exception) {
+            throw new RuntimeException(
+                'Could not delete user consent from the database for a specific SP',
+                0,
+                $exception
+            );
+        }
```
thijskh, line 128: I think we want to log that we did it.

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/DeprovisionControllerTest.php
```diff
@@ -278,6 +353,54 @@ public function all_user_data_for_collab_person_id_is_retrieved_and_deleted($met
...
+            'PHP_AUTH_USER' => $this->getContainer()->getParameter('api.users.deprovision.username'),
+            'PHP_AUTH_PW' => $this->getContainer()->getParameter('api.users.deprovision.password'),
+        ]);
+
+        $this->addServiceProviderUuidFixture($spUuid1, $spEntityId1);
+        $this->addUserFixture($userId, $userUuid);
+        $this->addSamlPersistentIdFixture($userUuid, $spUuid1, $persistentId1);
+        $this->addConsentFixture($userId, $spEntityId1, $attributeHash, $consentType, $consentDate);
+
+        $client->request('GET', 'https://engine-api.vm.openconext.org/remove-consent/urn:collab:person:test/https://my-first-sp.test');
```
thijskh, line 396: I'm not so sure about putting an URL in a URL. I can foresee all kinds of issues if we do not encode this. I am also not so sure that "GET" is the correct method? So maybe we solve both in one go and change it to a POST where we can supply the variables in the body?

src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
MKodde: I agree, additional logging for the 'happy flow' is meaningful here. May I suggest to perform this logging action in the controller action?

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/DeprovisionControllerTest.php
MKodde: Yes! I think using the DELETE verb or maybe POST is better than using a GET request. Delete would be most expressive from a restful point of view. But this would not eliminate the use of query parameters.

thijskh (changes requested):
The delete is too final. See https://www.pivotaltracker.com/story/show/180576312 for how we envision how soft-delete should work.

library/EngineBlock/User.php
```diff
@@ -59,7 +59,8 @@ public function getConsent()
     {
         $pdo = $this->_getDatabaseConnection();
         $query = 'SELECT service_id FROM consent
-                    WHERE hashed_user_id = ?';
+                    WHERE hashed_user_id = ?
+                    AND ISNULL(deleted_at)';
```
thijskh, line 63: Not sure this is standard SQL? Should that be `deleted_at IS NULL` instead?

src/OpenConext/EngineBlock/Service/ConsentService.php
```diff
@@ -92,6 +94,19 @@ public function countAllFor($userId)
         return count($consents);
     }
 
+    public function deleteOneConsentFor(CollabPersonId $id, string $serviceProviderEntityId): bool
+    {
+        try {
+            return $this->consentRepository->deleteOneFor($id->getCollabPersonId(), $serviceProviderEntityId);
+        } catch (Exception $e) {
+            throw new RuntimeException(
+                sprintf('An exception occurred while removing consent for a service provider("%s")', $e->getMessage()),
```
thijskh, line 103: While we're at it, add `$id` and `$serviceProviderEntityId` to the message?

src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
```diff
@@ -61,6 +66,8 @@ public function findAllFor($userId)
                 consent
             WHERE
                 hashed_user_id=:hashed_user_id
+            AND
+                ISNULL(deleted_at)
```
thijskh, line 70: as above

src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
```diff
@@ -95,8 +102,63 @@ public function deleteAllFor($userId)
...
+        $sql = '
+            UPDATE
+                consent
+            SET
+                deleted_at = NOW()
+            WHERE
+                hashed_user_id = :hashed_user_id
+            AND
+                service_id = :service_id
+            AND ISNULL(deleted_at)
```
thijskh, line 132: as above

src/OpenConext/EngineBlockBundle/Authentication/Repository/DbalConsentRepository.php
```diff
@@ -95,8 +102,63 @@ public function deleteAllFor($userId)
 
         try {
             $this->connection->executeQuery($sql, ['hashed_user_id' => sha1($userId)]);
+            $this->logger->info(sprintf('Removed consent for hashed user id %s (%s)', sha1($userId), $userId));
```
thijskh, line 105: I tend to make this a `notice` for now. Can always downgrade if that's too chatty

src/OpenConext/EngineBlockBundle/Controller/Api/ConsentController.php
```diff
@@ -88,4 +95,47 @@ public function userAction($userId, Request $request)
...
         return new JsonResponse($consentList, Response::HTTP_OK);
     }
+
+    public function removeAction(Request $request): JsonResponse
+    {
+        if (!$request->isMethod(Request::METHOD_POST)) {
+            throw ApiMethodNotAllowedHttpException::methodNotAllowed($request->getMethod(), [Request::METHOD_POST]);
+        }
+
+        if (!$this->featureConfiguration->isEnabled('api.consent_remove')) {
```
thijskh, line 105: Likely it makes sense to disable all consent endpoints if consent is disabled entirely. But I'll make a separate story for that

thijskh (approved):

library/EngineBlock/User.php
MKodde: Good point, I changed it to use the IS NULL syntax. Using the method was safe too, but not semantically correct.

PR #1161, opened 2021-09-20
Verify GHA Test intergration works as intended

MKodde:
The build of the PHP FPM container was not passing. This because NPM could not be installed. This was resolved by pulling NPM and Node from the official Node container. NPX needed a manual installation afterwards.

### DIFF
--- a/docker/php-fpm/Dockerfile-php74
+++ b/docker/php-fpm/Dockerfile-php74
@@ -6,11 +6,16 @@ ARG NPM_GID=1000
 # Copy phpfpm config
 COPY docker/php-fpm/app.ini /usr/local/etc/php/conf.d/
 
+# Yank the node and npm binaries from the official Node docker container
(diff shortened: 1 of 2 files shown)


docker/php-fpm/Dockerfile-php74
```diff
@@ -1,3 +1,4 @@
+FROM node:10 AS node
```
quartje, line 1: No need for this line, you can just do this in line 10 and 11
COPY --from=node:10 /usr/local/lib/node_modules /usr/local/lib/node_modules
COPY --from=node:10 /usr/local/bin/node /usr/local/bin/node

quartje (commented):
Good work! This seems to be fixing the issue . I've commented on a minor detail.

docker/php-fpm/Dockerfile-php74
MKodde: The more you know! Will update and test that!

quartje (approved):

PR #1162, opened 2021-09-24
Allow to specify a community Uuid in Manage.

thijskh:
All SPs in the community will get the same persistent NameID.

Example applications:
- A group of SP's is in fact the same logical service provider, divided over several technical SP entityIDs. They would all get different persistent IDs. Now they can be marked as a group and have a common ID, that is still scoped to this group and not relatable to IDs outside of the group.
- A service provider migrates to a different entityID. The second entity can keep the same persistent IDs that the first entity used to receive.

Use it by setting the field `coin:communityUuid` in Manage to a new Uuid, and set this same Uuid to any other entity that you want to be part of this community. For the migration case, set the Uuid to be the current Uuid of the SP in the service_provider_uuid table in Engineblock.

If an SP has this new metadata field, its identifiers will be based on this Uuid from now on. This means that:
- If you add the field to an SP, all current identifiers will change (if they exist), to become part of the new community. If you don't want this, choose the existing service_provider_uuid as the value.
- If you remove the field from an SP, nothing will change. If you want to split up a community, you need to explicitly assign a new Uuid as its community.

This requires also a (onliner) update to Manage to send the new field `coin:communityUuid` in the EB push.

### DIFF
--- a/library/EngineBlock/Saml2/NameIdResolver.php
+++ b/library/EngineBlock/Saml2/NameIdResolver.php
@@ -233,21 +233,31 @@ class EngineBlock_Saml2_NameIdResolver
         return $persistentId;
     }
 
-    protected function _getServiceProviderUuid($spEntityId)
+    protected function _getServiceProviderUuid(string $spEntityId): string
     {
-        $uuid = $this->_fetchServiceProviderUuid($spEntityId);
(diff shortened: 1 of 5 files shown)


thijskh:
Missing:

- [ ] Tests
- [ ] DB migration for the primary key

thijskh:
Not pursuing this PoC right now

PR #1164, opened 2021-10-19
Add "consent disable" feature

tbeekman:
This PR introduces the feature "feature_disable_consent". When set to true, the consent functionality is completely disabled (in contrast to setting "consent_store_values" to false). This means the following:
- the consent page will not be shown
- the database will not written to
- the database will not be queried

This has some benefits for usage of EngineBlock in cases where the consent functionality is not used. There will be no roundtrip to the database which optimizes performance. Furthermore, since no consents need to be written to the database, a read-only user can be used to access the "eb" database, which opens up opportunities for database scaling. 

If the feature is set to false, the existing functionality will work as it currently does.


tbeekman:
Hi @thijskh,

Thanks for the review. As for mentioning this feature in a positive sense, good addition. The ability to disable a feature is not a feature on its own. We refactored the code and added the commit to this PR.

As for the functional test; would be a great addition. However, we currently do not have these functional tests running on our own environment, so adding such a test would be quite a tedious task. Would be great however if someone from your team could add this instead.

PR #1166, opened 2021-10-21
Save custom objects using custom doctrine type mappings switch to eb6

Stephan-Kok:
This is the pull request for step 2 (**switch to eb6**) in the implementation of the Custom doctrine database types.

In this PR we no longer use sso_provider_roles_eb5 for the main process, but instead we use sso_provider_roles_eb6. The push still updates both, sso_provider_roles_eb5 and sso_provider_roles_eb6 so if other nodes are outdated, that is not a problem.

Please see https://github.com/OpenConext/OpenConext-engineblock/pull/1159 for step 1

The required steps per release are:
- **prepare eb6** - Introduce sso_provider_roles_eb6 and temporary push to both sso_provider_roles_eb5 and sso_provider_roles_eb6. Still use sso_provider_roles_eb5 for the main process
- **switch to eb6** - temporary push to both sso_provider_roles_eb5 and sso_provider_roles_eb6. Use sso_provider_roles_eb6 for the main process
- **Remove eb5** - Clean up and remove sso_provider_roles_eb5

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -95,9 +95,19 @@ doctrine:
                 mapping_types:
                     enum: string
         types:
+            engineblock_attribute_release_policy: OpenConext\EngineBlockBundle\Doctrine\Type\AttributeReleasePolicyType
+            engineblock_certificate_array: OpenConext\EngineBlockBundle\Doctrine\Type\CertificateArrayType
             engineblock_collab_person_id: OpenConext\EngineBlockBundle\Doctrine\Type\CollabPersonIdType
             engineblock_collab_person_uuid: OpenConext\EngineBlockBundle\Doctrine\Type\CollabPersonUuidType
+            engineblock_contact_person_array: OpenConext\EngineBlockBundle\Doctrine\Type\ContactPersonArrayType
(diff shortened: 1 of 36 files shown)


thijskh:
This might be completely out of scope, but I'll drop the question here anyway. Instead of converting all these things to stuff them in the RDBMS in a different way than before, how would you feel about moving to a schemaless backend, at least for the sso provider roles?

We have a MongoDB already running within the platform.

quartje:
> This might be completely out of scope, but I'll drop the question here anyway. Instead of converting all these things to stuff them in the RDBMS in a different way than before, how would you feel about moving to a schemaless backend, at least for the sso provider roles?
> 
> We have a MongoDB already running within the platform.

A schemaless approach for this particular table is certainly a good idea. However, having two database engines in stead of one in the critical path for all logins doesn't really sound uptime enhancing. And schemaless doesn't solve the problem of back- and forward compatibility here. 
Maybe there's a way to do schemaless with mySQL?

tvdijen:
Personally, I never understood the half move to MongoDB.. Sure, NoSQL is great, but it added a lot of complexity on top of yet a complex eco-system..
The json-content in SQL solves most of the BC, but it makes it quite impossible to query the db...

PR #1172, opened 2022-02-16
Cover internal-CollabPersonId behaviour in test

MKodde:
The different scenarios with and without a trusted proxy are covered.

1. A normal situation where no TP is concerned should yield no internal    collab person id
2. With TP, different NameId formats are not transformed, but remain    unmangled in the assertion. The internal collab person id will contain an unspecified string representation of the users identifier
3. EPTI can still be present. It is set with the same value found in the NameId. So no trnasformations there either.

For now these changes have only been covered in behat tests. And a test that was already touching this scenario was (re)moved from SpProxy.feature.

Note that the behat test are not passing yet, at least the three ones covering the new situation will not. They are marked WIP

See: https://www.pivotaltracker.com/story/show/181121279


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
```diff
@@ -0,0 +1,57 @@
...
+    And SP "Step Up TP" signs its requests
+    And SP "Step Up TP" does not require consent
+    And SP "Step Up TP" uses the Unspecified NameID format
+    When I log in at "Step Up TP"
+    And I pass through EngineBlock
+    And I pass through the IdP
+    And I pass through EngineBlock
+    Then the response should match xpath '/samlp:Response/saml:Assertion/saml:Subject/saml:NameID[@Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"]'
+    And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:dir:attribute-def:eduPersonTargetedID"]/saml:AttributeValue/saml:NameID[@Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"]'
+    And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:surf.nl:attribute-def:internal-collabPersonId"]'
```
thijskh, line 41: For this atrribute we should check that it in fact contains `urn:collab:person:...` You could make a separate feature to check the contents of the new attribute extensively (or do it in unit tests). Check that internal-collabPersonId:
- Has exactly one value
- Is of type xs:string
- Contains `urn:collab:person:...`

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
```diff
@@ -0,0 +1,57 @@
...
+
+  # In a TP authentication, the internal-collabPersonId is present. NameId format is not touched
+  @WIP
+  Scenario: User logs in via trusted proxy in that case the internalCollabPersonId should be present and NameId format is not modified
+    Given SP "Step Up TP" is authenticating for SP "SelfService"
+    And SP "Step Up TP" is a trusted proxy
+    And SP "Step Up TP" signs its requests
+    And SP "Step Up TP" does not require consent
+    And SP "Step Up TP" uses the Transient NameID format
```
thijskh, line 50: As above. The NameIdFormat should be determined by the end-SP.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
```diff
@@ -0,0 +1,57 @@
...
+    When I log in at "Step Up TP"
+    And I pass through EngineBlock
+    And I pass through the IdP
+    And I pass through EngineBlock
+    Then the response should match xpath '/samlp:Response/saml:Assertion/saml:Subject/saml:NameID[@Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient"]'
+    And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:dir:attribute-def:eduPersonTargetedID"]/saml:AttributeValue/saml:NameID[@Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient"]'
+    And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:surf.nl:attribute-def:internal-collabPersonId"]'
```
thijskh, line 57: FWIW, the epTID is released in all these cases because the end SP has no ARP. This is correct and tested in the epTID feature. Should the end-SP have an ARP (without epTID enabled) then no epTID will be released.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
```diff
@@ -0,0 +1,57 @@
...
+    And SP "Step Up TP" uses the Unspecified NameID format
+    When I log in at "Step Up TP"
+    And I pass through EngineBlock
+    And I pass through the IdP
+    And I pass through EngineBlock
+    Then the response should match xpath '/samlp:Response/saml:Assertion/saml:Subject/saml:NameID[@Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"]'
+    And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:dir:attribute-def:eduPersonTargetedID"]/saml:AttributeValue/saml:NameID[@Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"]'
+    And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:surf.nl:attribute-def:internal-collabPersonId"]'
+
+  # In a TP authentication, the internal-collabPersonId is present. NameId format is not touched
```
thijskh, line 43: Description is same as feature above.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/SpProxy.feature
```diff
@@ -188,20 +188,6 @@ Feature:
...
-      And SP "Step Up" signs its requests
-      And SP "Step Up" does not require consent
-      And SP "Step Up" uses the Unspecified NameID format
-     When I log in at "Step Up"
-      And I select "AlwaysAuth" on the WAYF
-      And I pass through EngineBlock
-      And I pass through the IdP
-      And I pass through EngineBlock
-     Then the response should match xpath '/samlp:Response/saml:Assertion/saml:Subject/saml:NameID[@Format="urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"]'
-      And the response should match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:dir:attribute-def:eduPersonTargetedID"]/saml:AttributeValue/saml:NameID[@Format="urn:oasis:names:tc:SAML:2.0:nameid-format:persistent"]'
```
thijskh, line 203: As you see here, the formats are different in the old scenario, FWIW.

thijskh (changes requested):

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
```diff
@@ -0,0 +1,57 @@
+Feature:
+  In order to pass an identifier to the trusted proxy
+  As engineblock
+  I need to show correct behavour in different situations
```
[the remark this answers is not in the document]
MKodde: Good suggestion!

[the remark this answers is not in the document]
thijskh: I wrote "iff" on purpose :-)

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
```diff
@@ -0,0 +1,57 @@
...
+    And the response should not match xpath '/samlp:Response/saml:Assertion/saml:AttributeStatement/saml:Attribute[@Name="urn:mace:surf.nl:attribute-def:internal-collabPersonId"]'
+
+  # In a TP authentication, the internal-collabPersonId is present. NameId format is not touched
+  @WIP
+  Scenario: User logs in via trusted proxy in that case the internalCollabPersonId should be present
+    Given SP "Step Up TP" is authenticating for SP "SelfService"
+    And SP "Step Up TP" is a trusted proxy
+    And SP "Step Up TP" signs its requests
+    And SP "Step Up TP" does not require consent
+    And SP "Step Up TP" uses the Unspecified NameID format
```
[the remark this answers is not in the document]
MKodde: Good catch!

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
MKodde: In retrospect, the Scenario description should be sufficient in giving context to the meaning of the test. Removed the duplicate comments.

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/InternalCollabPersonId.feature
MKodde: Points 1 and 2 can be easily achieved. Point three is somewhat more challenging. 

The level of control over the user attributes being released is limited at this point. I'm looking into this. As of now, the user identifier is a random hash, different every run.

thijskh: I don't follow, internal-collabPersonId is always of the form `urn:collab:person:example.com:userid`

MKodde: In the end we might like to add something to modify the identity of the assertion via a step definition. How we want to do this is something I'd like to discuss with you on our upcoming peer session where we will discuss the additional TP test coverage.

thijskh: Regardless of the content, I think it also still misses the "has exactly one value" test.

MKodde: Yes, that one I missed. Thanks for the warning!

thijskh (approved):

PR #1173, opened 2022-02-16
Upgrade Twig to 2.x

thijskh:


thijskh:
The testsuite fails only on the security audit which highlights a vulnerability irrelevant to EB...

theme/base/templates/layouts/scripts/default.html.twig
```diff
@@ -14,7 +14,7 @@
         <link rel="preload" href="/fonts/sourcesanspro-bold-webfont.woff2?v={{ assetsVersion }}" as="font" type="font/woff2" crossorigin>
         <link rel="preload" href="/fonts/nunito-semibold.woff2?v={{ assetsVersion }}" as="font" type="font/woff2" crossorigin>
         <link rel="preload" href="/fonts/sourcesanspro-regular-webfont.woff2?v={{ assetsVersion }}" as="font" type="font/woff2" crossorigin>
-        <link href="/stylesheets/application.css?v={{ assetsVersion }}" rel="stylesheet" type="text/css"/>
+        <link href="/stylesheets/application.css?v={{ assetsVersion }}" rel="stylesheet">
```
MKodde, line 17: On multiple occasions you removed the `type` attribute from the `link` elements. No problem, all browsers will implement this correctly and will automatically assume a stylesheet is of type text/css. 

I wonder why you removed this explicit type definition.

theme/base/templates/modules/Default/Partials/notConverted/footer.html.twig
```diff
@@ -1,4 +1,6 @@
+{% import "@themeLayouts/scripts/language-switcher.html.twig" as languageSwitcher %}
```
MKodde, line 1: Nice move to load the language switcher only in the templates where it is actually used :+1:

MKodde (approved):
Thanks for updating Twig to a modern version! See two points in the code review below.

theme/base/templates/layouts/scripts/default.html.twig
thijskh: It is considered [recommended practice in HTML5 to leave it out](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-type).

theme/base/templates/modules/Default/Partials/notConverted/footer.html.twig
thijskh: It's actually required in Twig 2. Importing a macro makes it only available to the exact template file it's imported in, not child/included templates.

PR #1174, opened 2022-02-17
Trusted proxy documentation update

thijskh:
Describe changed behaviour


thijskh:
I've changed it for consistency. We can squash-merge this when #1175 is merged.

PR #1175, opened 2022-02-17
Apply trusted proxy related output filtering

MKodde:
For now this includes adding the internal-collabPersonId to the
assertion.

Other modifications to get the required behavior:
1. Improved logging in output filtering for the affected output filters (nameId, eduPersonTargetedId and internalCollabPersonId)
2. Add internal-collabPersonId value check in behat step definition
3. Rename classname with typo AddEduPersonTarget~~t~~edId


MKodde:
Hi @thijskh, these changes make the tests go green. And they seem to adhere to the description found in the story.

What bothers me slightly is that no 'special case' code was removed. Only extra special case code was added :clown_face: 

The NameId behavior was changed. The released name Id now always follows that of the 'destination' SP. Previously we just addressed the SP next in line. Regardless of it being TP or not.

Aside from a fixed typo and additional logging. Nothing changed in the EPTI output filter.

Did I miswrite the Behat tests? Am I overlooking a 'special case' that you are aware of and I am not?

library/EngineBlock/Corto/Filter/Command/ApplyTrustedProxyBehavior.php
```diff
@@ -0,0 +1,63 @@
...
+     * urn:mace:surf.nl:attribute-def:internal-collabPersonId
+     */
+    public function execute()
+    {
+        $this->logger->info('Executing the ApplyTrustedProxyBehavior output filter');
+        if (!$this->_serviceProvider->getCoins()->isTrustedProxy()) {
+            return;
+        }
+        $this->logger->info('Adding internal-collabPersonId to the attributes, set with collabPersonId');
+        $this->_responseAttributes['urn:mace:surf.nl:attribute-def:internal-collabPersonId'] = [$this->_collabPersonId];
```
thijskh, line 56: attribute name could be from Constants, to mirror how the eduPersonTargetedId command does it.

thijskh (commented):
The change looks like it will do the job. Good thing to add the logging.

You ask why no special case is removed? I think this is the explanation:
- Previously, the NameID was always just the NameId format configured for the next hop SP (no special casing for TP). This means it would work iff you configured the TP correctly with an unspecified nameformat: the TP would as any other directly connected SP always get an unspecified nameID. So in a sense the "special case" was that you needed to know to set this correctly for the TP entity.
- epTID already contained the special casing of TP: when TP involved use the end-SP's nameidformat.

The chosen solution just makes the algorithm the same for NameId and epTID. So in that sense there's "less special case": the two things now work the same.

The only thing I'd remark about that, is that the NameId filter and epTID filter are now very similar. We could do two things:

1. Accept that, to keep this change targeted
2. Refactor it. Maybe it's not hard to flatten them into one Command.

library/EngineBlock/Corto/Filter/Command/ApplyTrustedProxyBehavior.php
MKodde: Yes, that is one of the other tasks I added on the story. I'm going to add that on this PR or maybe on a separate one.

library/EngineBlock/Corto/Filter/Command/ApplyTrustedProxyBehavior.php
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * When a TP is involved in an authentication, a special case is triggered:
+     *   the NameID is set to the collabPersonId ("urn:collab:person:...") and
+     *   the epTID to the remote SP's NameID as configured (persistent,
+     *   transient...).
+     *
+     * We want to move the information about the collabPersonId that the TP needs
+     * to a dedicated attribute. And then not make special cases for NameID, let
+     * this work just like any other SP.
```
thijskh, line 45: I would personally not describe the development process/history in the docblock. Just say what it does now. History is in commit messages and/or stories and/or PRs. So more something along the lines of

"Provide a dedicated internal attribute so the TP can identify the user uniquely. The 'regular' ARP/nameId attributes cannot always be used for this, because these differ per and may be scoped for the end-SP."

thijskh (approved):
Seems sound, just a personal preference about docblock comments

library/EngineBlock/Corto/Filter/Command/ApplyTrustedProxyBehavior.php
MKodde: Again, good suggestion. I just copy pasted this and never looked back to the relevance of the text that was pasted.

PR #1178, opened 2022-02-23
Initially sort the displayed IdPs in WAYF correctly.

thijskh:
The IdPs in the WAYF are supposedly sorted: by count (when used before)
or by name (remaining IdPs). This happens in JS and happens correctly
when you use the sort field or clear it. However, the initial state
(the screen as shown when not searching for anything yet) is unsorted.

This is solved in the twig template by sorting the used IdPs (by
count, descending - most often used first) and the remainig IdPs by
displayTitle(lowercase). This is the same algorithm as JS used. Solving
this in the template/HTML means it's also solved for non-JS scenarios.
and there's no need to 'resort' the initial screen before it can be
displayed.


Badlapje (approved):
Looks like a nice way to handle this 👍

PR #1179, opened 2022-02-24
Improve Trused Proxy Behat test coverage

MKodde:
Thanks to a good sit-down with @thijskh several relevant trusted proxy behavioral tests where identified. Most of which are variations on existing tests that are not TP related. The tests have been specified in the tasks of the story linked below. This PR adds these tests.

### DIFF
--- a/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/PolicyEnforcement.feature
+++ b/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/PolicyEnforcement.feature
@@ -47,7 +47,7 @@ Feature:
      When I log in at "Dummy SP"
       And I pass through EngineBlock
       And I pass through the IdP
-      And I should not see "Error - No access"
+      And I should not see "Error - Access denied"
 
   Scenario: Access is permitted because of a Not Applicable policy
     Given SP "Dummy SP" requires a policy enforcement decision
@@ -55,7 +55,7 @@ Feature:
      When I log in at "Dummy SP"
       And I pass through EngineBlock
       And I pass through the IdP
-      And I should not see "Error - No access"
+      And I should not see "Error - Access denied"
 
   Scenario: Access is denied because of a Deny policy, End-SP behind Trusted Proxy
     Given SP "Stepup SelfService" requires a policy enforcement decision
@@ -90,7 +90,7 @@ Feature:
     When I log in at "Stepup Gateway"
     And I pass through EngineBlock
     And I pass through the IdP
-    And I should not see "Error - No access"
+    And I should not see "Error - Access denied"
 
   Scenario: Access is permitted because of a Not Applicable policy, End-SP behind Trusted Proxy
     Given SP "Stepup SelfService" requires a policy enforcement decision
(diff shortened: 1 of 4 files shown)


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/PolicyEnforcement.feature
```diff
@@ -54,3 +56,49 @@ Feature:
...
+  Scenario: Access is denied because of an Indeterminate policy, End-SP behind Trusted Proxy
+    Given SP "Stepup SelfService" requires a policy enforcement decision
+    And SP "Stepup Gateway" is authenticating for SP "Stepup SelfService"
+    And SP "Stepup Gateway" is a trusted proxy
+    And SP "Stepup Gateway" signs its requests
+    And pdp gives an indeterminate response
+    When I log in at "Stepup Gateway"
+    And I pass through EngineBlock
+    And I pass through the IdP
+    And I should see "Error - Access denied"
```
thijskh, line 81: This checks for "I should see "Error - *Access denied*"...

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/PolicyEnforcement.feature
```diff
@@ -54,3 +56,49 @@ Feature:
...
+  Scenario: Access is permitted because of a Permit policy, End-SP behind Trusted Proxy
+    Given SP "Stepup SelfService" requires a policy enforcement decision
+    And SP "Stepup Gateway" is authenticating for SP "Stepup SelfService"
+    And SP "Stepup Gateway" is a trusted proxy
+    And SP "Stepup Gateway" signs its requests
+    And pdp gives a permit response
+    When I log in at "Stepup Gateway"
+    And I pass through EngineBlock
+    And I pass through the IdP
+    And I should not see "Error - No access"
```
thijskh, line 93: ...but this checks for "I should **not** see "Error - *No access*"?

thijskh (commented):
Looks good, good to have more assurance about the current behaviour so we're more certain that this will remain working as specified in the future.

This PR could also remove this test which seems duplicated.

https://github.com/OpenConext/OpenConext-engineblock/blob/6478395f3fe66602d02bdf4701f88b068e5bc1c6/src/OpenConext/EngineBlockFunctionalTestingBundle/Features/SpProxy.feature#L116

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/PolicyEnforcement.feature
MKodde: This tests behavior similar to that found in scenario: `Scenario: Access is permitted because of a Permit policy:44`

Here the PDP gives a `permit` response. Hence the test:  "I should *not* see "Error - No access".

MKodde: :zap: But maybe you are hinting on the fact that the error text is not the same? And needs updating.

thijskh: Maybe if the error text is wrong... it will always be "not seen"

thijskh (approved):

PR #1180, opened 2022-03-01
Show NameID for end-SP in consent

thijskh:
The consent screen always displayed the NameID for the next-hop SP, regardless of whether this was a trusted proxy. All other information in the consent screen was neatly derived from the end-SP. This makes it so that the NameID to display is also read from the end-SP instead of the next-hop SP.

This problem has always existed, but now that we've reorganised the way we deal with trusted proxies it became more apparent and also that the fix was easy.

This PR first created some (initially failing) tests for the output in the consent screen. Then the actual fix which is to use the already present variable for the end-SP metadata instead of the next hop SP metadata.


MKodde (approved):
Nothing but net!

One minor note would be that your use of the `serviceProviderMetadata` (first SP in requester chain) might look somewhat confusing to a less informed person reading this code.

Maybe I should have addressed this feedback myself while writing that code in the first place. But maybe you'd be so kind to add a small explanation to line `125` as to why we create a `serviceProviderMetadata` and why this is the first SP from the `spMetadataChain`.

Feel free to leave this as-is.

And kudos for the behat tests!

PR #1187, opened 2022-04-28
Prevent removal of form from IdP when previously removed

MKodde:
**Description of issue:**
If you press an unconnected IdP the RA form is shown, but you click the greyed out IdP (which remains visible) again you get errors in the JS console

**Solution:**
Only remove the form from the IdP clone when the form still exists. Clicking on a disconnected idp triggers the script to handle the request access form interactions. One of these actions is to remove the form triggering SSO on the targetted IdP. When the form is already removed in a previous click, a successive click would trigger a console error.

**See:**
https://www.pivotaltracker.com/n/projects/1425900/stories/180935719


Badlapje (approved):
The perfect PR: hardly a line to review and well documented 🚀 :shipit:

PR #1188, opened 2022-04-28
Return HTTP status code 400 on validation error

MKodde:
Previously a 200 response was returned. Which was interpreted by the JS application as being a valid, all is good response. This was changed in this commit. A 400 status code coreesponds to 'Bad request' which seems most appropriate from a semantical point of view.

See task 2 of: https://www.pivotaltracker.com/story/show/180935719


Badlapje (approved):
🚢 it 👍

PR #1189, opened 2022-04-28
Set IdP name on the wayf__idp row

MKodde:
This display name is assumed to be set on the idp row when sending the request access mail message. Not having it set here will result in the mail message not being sent.

See point 3 of: https://www.pivotaltracker.com/story/show/180935719


Badlapje (approved):
Looks good to me 👍 Nice work!

PR #1191, opened 2022-05-02
Disable the Consent-related API endpoints if EB's consent feature is off

thijskh:
Makes behaviour more explicit and does not expose code paths that are not needed.

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Controller/Api/ConsentController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/Api/ConsentController.php
@@ -70,6 +70,10 @@ final class ConsentController
             throw ApiMethodNotAllowedHttpException::methodNotAllowed($request->getMethod(), [Request::METHOD_GET]);
         }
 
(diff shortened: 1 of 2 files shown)


thijskh:
It seemed so simple but it appears I broke the tests of the existing api feature flags... but why? @MKodde any idea?

MKodde:
Nice addition! 

The test breaks because of the `disableConsentApiFeatureFor` and `disableRemoveConsentApiFeatureFor` methods that need to take into account the existance of the `eb.feature_enable_consent` feature flag. The test only knows about the feature flags that are explicitly configured in test. The controller actions that where tested in the failing tests referenced to feature flags that where not configured.

thijskh:
Build only fails on security (non)vulnerability in deps

MKodde (approved):

MKodde:
> Build only fails on security (non)vulnerability in deps

Fixing that in #1190

thijskh:
@tbeekman Makes sense?

PR #1197, opened 2022-06-24
Support subjectid

thijskh:


theme/base/templates/modules/Authentication/View/Proxy/debug-idp-response.html.twig
```diff
@@ -120,9 +120,6 @@
                         <th>
                             {{ 'value'|trans }}
                         </th>
-                        <th>
-                            {{ 'remarks'|trans }}
-                        </th>
```
MKodde, line 125: 1. Table heading column was removed :heavy_check_mark: 
2. The remark was moved to a new table row the current row (if it exists) :heavy_check_mark: 

You might need to remove the obsolete 'remark' translation. Or maybe set it on the empty table data cell on line 162

theme/base/templates/modules/Authentication/View/Proxy/debug-idp-response.html.twig
```diff
@@ -67,6 +67,17 @@
...
                         </td>
                     </tr>
+                    <tr>
+                        <th>
+                            {{ 'allowed_scopes'|trans }}
+                        </th>
+                        <td>
+                            {% for scope in idp.shibMdScopes %}
+                                {{ scope.allowed }}
+                                {% if scope.regex %}(regex){% endif %}
```
MKodde, line 77: One could argue this does not need translation: `(regex)`

I still strongly suggest to still create a translation entry for this string. If only to follow procedure.

MKodde: I later saw that this debug page is not concerned with being translatable. Feel free to ignore the message above :muscle:

theme/base/templates/modules/Authentication/View/Proxy/debug-idp-response.html.twig
```diff
@@ -140,7 +140,7 @@
                                 {% if validationResult.isValid(attributeName) %}
                                     <i class="c-icon c-icon-checkmark fa-check"></i>
                                 {% else %}
-                                    {% if validationResult.errors(attributeName).length > 0 %}
+                                    {% if validationResult.errors(attributeName)|length > 0 %}
```
MKodde, line 130: :medal_military:

MKodde (commented):
Nice work! See some feedback below

theme/base/templates/modules/Authentication/View/Proxy/debug-idp-response.html.twig
thijskh: I indeed took the more pragmatic approach here that other technical terms are not translated currently in the debug page so 'regex' seemed to fall in that category. So I'll leave it at that.

PR #1199, opened 2022-08-29
Make sure rememberChoice is called

ArnoutvdKnaap:
Added the call to rememberChoice functionality in subMitForm.js


MKodde (commented):
This looks like a very straight forward solutions and those are the best :+1: 

One question I was not directly able to answer myself yet; Shouldn't the rememberChoice call not be behind the feature flag? Preventing the calling of the remember choice feature when the feature is disabled? 

I tested this in my dev env and no side effects where encoin

ArnoutvdKnaap:
currently it is not behind one as the function itself was already checking for the rememberChoice checkbox which is behind a feature check

MKodde (approved):
Thanks for verifying that!

PR #1211, opened 2022-12-12
Set default value on deleted_at column

MKodde:
Having the default set to 0000-00-00 ensures we can make this column backwards compatible with a pre 6.8.0 release


thijskh:
It seems to do the trick. I do not quite follow yet why it works, because the code checks for "IS NULL", are rows with 0000-00-00 then considered to match that expression?

thijskh (approved):

PR #1212, opened 2022-12-12
Improve logging of RequesterID feature and add tests

thijskh:
A feature exists in EB where it can sub-scope the available IdPs for an SP if this SP passes it as a RequesterID in the authentication request and we know this remote SP: the allowed IdPs of this remote SP are then used to filter the allowed IdPs for the authenticating SP.

The logging associated with this feature was confusing. When creating the feature it was assumed that SPs would only send RequesterIDs if they would be using this feature, so  a warning was added if the mentioned SP is unknown. We now know that this is a common and legitimate case, so downgrade this warning to an info level message.

There's some discussion about whether this feature should be kept at all. For now we do not know for certain that no-one using OpenConext actually uses it. It is not very painful to keep (has been there for 10 years without change). For starters, add logging to when the feature actually _is_ used, so we get information about this. And add test coverage so this (and other features) are at least covered both when its used and when it should not have an effect.

### DIFF
--- a/library/EngineBlock/Corto/Adapter.php
+++ b/library/EngineBlock/Corto/Adapter.php
@@ -208,6 +208,9 @@ class EngineBlock_Corto_Adapter
             $serviceProvider = $repository->findServiceProviderByEntityId($requesterId);
 
             if ($serviceProvider) {
+                $this->_getLogger()->info(
+                    "SP passes RequesterID '$requesterId', using it to sub-scope the available IdPs"
+                );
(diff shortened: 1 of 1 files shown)


tvdijen:
Works for me!  As for the discussion whether to keep this functionality or not.. For people using this, it would only be a matter to mark an SP as a trusted proxy to have the exact same functionality, right?

thijskh:
Yes, but the trusted proxy gains more powers than just influencing the ACL.

PR #1215, opened 2023-01-03
Prefer [] operator for adding one element to an array

thijskh:


MKodde (commented):
Nice optimization work, did you leave out: `EngineBlock_Application_ErrorHandler::pushExitHandler` intentionally?

thijskh:
Yes, because of the push/pop methods that are implemented symmetrically with arry_push/pop, seemed ok there.

PR #1219, opened 2023-02-13
Limit value length of metadata fields to max 255 chars.

thijskh:
The current database scheme has varchar(255) for several metadata types. Traditionally, MySQL would just truncate silently any too long values. Modern installations have strict mode enabled which instead generates a justified error. To make EB usable in this configuration, limit the length of some fields that (a) have a trend to sometimes be long, (b) which can be truncated without real loss.

One could consider to enlarge columns but as far as I'm aware description is not used and 255 chars of keywords should be enough for everyone. So for now lets keep it to this simple fix.

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
@@ -162,7 +162,7 @@ class PushMetadataAssembler implements MetadataAssemblerInterface
      */
     private function limitValueLength(string $value): string
     {
-        if(strlen($value) < self::FIELDS_MAX_LENGTH) {
+        if (strlen($value) < self::FIELDS_MAX_LENGTH) {
             return $value;
         }
         $this->logger->info(sprintf("Push Metadata Assembler: truncating too long value: '%s'", $value));
--- a/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
@@ -48,6 +48,11 @@ class PushMetadataAssembler implements MetadataAssemblerInterface
      */
     private $allowedAcsLocationsValidator;
 
+    /**
+     * Maximum length that will fit in certain string type fields.
+     */
+    private const FIELDS_MAX_LENGTH = 255;
+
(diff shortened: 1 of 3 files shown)


src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
```diff
@@ -418,12 +434,15 @@ private function assembleSingleLogoutServices($connection)
         ));
     }
 
-    private function setPathFromObjectString($from, $to)
+    private function setPathFromObjectString($from, $to, $limitlength = false)
```
tvdijen, line 437: bool typehint?

src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
```diff
@@ -142,6 +147,17 @@ function ($allowedConnection) {
         return $roles;
     }
 
+    /**
+     * Shorten possibly lengthy fields that need to fit into a varchar(255).
+     * If they are longer than this, we likely do not use the contents for
+     * anything useful so it's OK to just chop it off. Otherwise the assembler
+     * will get errors when using MySQL with strict mode enabled.
```
MKodde, line 154: 1. I'm happy with this explanation, it would help me greatly while figuring out why my fields are truncated. 

2. Would logging this occurrence give added value? That way you can more easily remedy these issues (if need be).

3. Another though: If truncation occurs in the real world. Shouldn't we increase the length of the database fields instead? In other words, shouldn't we match the length requirements Manage states for these fields?

MKodde (approved):
Nice work. See my three comments below. And additionally, (sh|c)ould add test coverage of you functional change in `PushmetadataAssemblerTest`.

src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
thijskh: As for 3. Indeed the more robust approach is to reimagine the sso provider roles database structure, extend some fields, define some limits. But that has much more impact for which there is now no time. So the choice is:

1. Fix this in a simple way for fields that are not really used in an important fashion anyway
2. Not fix this and wait for when we have time for a perfect solution

thijskh:
I have implemented the feedback, will square after review

MKodde (approved):
Thanks for clarifying your motivation and adding the additional logging! :ship:

PR #1220, opened 2023-02-20
Move metadata certificate endpoint to Symfony

thijskh:
The certificate endpoint was still served by a service in the Corto framework, as the only part of the metadata URLs. So also move this functionality into the Symfony structure, simplifying it as we go and add functional tests for it.


src/OpenConext/EngineBlockBundle/Controller/MetadataController.php
```diff
@@ -55,11 +55,7 @@ public function __construct(
         $this->router = $router;
```
MKodde, line 55: Both the router and the ebApplicationSingleton seem unused in this controller. Consider removing them under the boyscout rule.

src/OpenConext/EngineBlockBundle/Controller/MetadataController.php
```diff
@@ -124,4 +105,17 @@ public function stepupMetadataAction($keyId = null)
...
+
+    public function signingCertificateAction(string $keyId = null): Response
+    {
+        if (empty($keyId)) {
+            $keyId = KeyPairFactory::DEFAULT_KEY_PAIR_IDENTIFIER;
+        }
+
+        $cert = $this->metadataService->certificate($keyId);
+        $response = new Response($cert);
+        $response->headers->set('Content-Type', 'application/x-pem-file');
```
MKodde, line 117: This will probably work as the browser itself will probably pick this up as a file download. But to ensure good compatibility for any browser. And to adhere to the best practices consider setting a disposition: attachment or use the binary file response type:

`https://symfony.com/doc/3.4/components/http_foundation.html#serving-files`

MKodde (changes requested):
Nice work. See some suggestions below. Chief amongst which the suggestion regarding serving of files.

src/OpenConext/EngineBlockBundle/Controller/MetadataController.php
thijskh: Note that we already serve the XML files in this way in this same file. I wanted to keep it consistent so do it the same for XML and PEM since I personally do not perceive a difference here. In the interest of simplicity I opted to make the PEM serving the same as the XML serving.

src/OpenConext/EngineBlock/Xml/MetadataProvider.php
```diff
@@ -141,4 +149,9 @@ public function metadataForStepup(string $keyId): string
 
         return $this->renderer->fromServiceProviderEntity($serviceProvider, $keyId);
     }
+
+    public function certificate(string $keyId): string
+    {
+        return $this->keyPairFactory->buildFromIdentifier($keyId)->getCertificate()->toPem();
```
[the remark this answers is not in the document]
thijskh: I don't think this can happen because buildFromIdentifier will build an X509KeyPair with both private and public key set so there's not practical way we can arrive at that situation. All the better throw an error if this occurs because it should never.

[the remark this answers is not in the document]
MKodde: Thanks for checking that out

MKodde (approved):

PR #1222, opened 2023-02-20
Display no metadata endpoints with key default

thijskh:
Previously, the index page of Engineblock would look like this (an installation with two keys in the config: the same key is set as 'default' and '20181213'):

Screenshot EB index showing three links for each type of metadata

There's three links of metadata for each type. However, the first two are duplicated. The URLs without a keyslug are already implying that the default key is used. So there's no actual difference between those two. If the default key changes, both will change. This is confusing and also clutters the page, even more for installs that have a single key defined.

Skip the redundant key:default when generating the list of metadata variants. The URLs with key:default in them will still work but just are not displayed. They will still work because key:default is not otherwise 'special' and just another key in the config. Changing that (e.g.: having one key and an option to mark it as default or not) requires a significant amount of refactoring which does not seem currently to bring a lot of value.

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Controller/IndexController.php
+++ b/src/OpenConext/EngineBlockBundle/Controller/IndexController.php
@@ -18,6 +18,7 @@
 
 namespace OpenConext\EngineBlockBundle\Controller;
 
+use OpenConext\EngineBlock\Metadata\X509\KeyPairFactory;
 use Symfony\Component\HttpFoundation\Response;
 use Twig\Environment;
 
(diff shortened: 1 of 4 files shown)


tvdijen:
We've discussed this many times already, but the `default` key was supposed to hold all the keys to accommodate key-rollover.. This is still problematic, especially on the SP-metadata (we sign AuthnRequests)

thijskh:
I think what is contained in the metadata without keyslug is a good discussion and could certainly be changed, but whatever we choose there does not change the fact that `/metadata` and `/metadata/key:default` are redundant now and also then.

thijskh:
Maybe file a separate issue to discuss changes to the default metadata content?

MKodde (approved):

PR #1226, opened 2023-03-28
If we have a request ID, it's by definition not unsolicited

thijskh:
SAML Profiles 2.0 4.1.5


MKodde (approved):
I see the reasoning to this change, and the error message became less vague which is a good thing! See my one suggestion to fix a typo.

PR #1227, opened 2023-03-28
No longer run security checker as part of integration tests

thijskh:
There's no relation between changes in a PR and the presence of a security issue or between the frequency of PRs and the frequency we want to check for security issues. So remove it here. We keep the script in bin/ for now, might come in handy at some point.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -122,18 +122,6 @@ jobs:
                   '
               env:
                   SYMFONY_ENV: ci
-            - name: Run security audit
-              if: always()
-              run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test-integration.yml
MKodde, line 1: This is precisely what we intended.

An additional thought: We could also get rid of the SensioLabs Security Check setup as Composer now ships with an audit feature. But feel free to leave the rest of the solution in place as-is.

MKodde (approved):

.github/workflows/test-integration.yml
thijskh: Yes, as said in the PR I think we can decide on whether to remove the script altogether later.

PR #1230, opened 2023-03-30
Support ForceAuthn to Stepup Gateway callout

MKodde:
Manage can push SP Metadata containing the coin:stepup:forceauthn coin.

That value is picked up in the Assembler, and saved on the SP eb5 roles, coin column.

During the StepUp round trip via StepUp Gateway, EB checks if the FoceAuthn flag should be added to that AuthNRequest. If this is specified for that SP, it is added. Otherwise no additional logic is performed.

When dealing with a trusted proxy setup, the stepup settings for the requesting sp are used, not that of the TP.

In order to get insghts in the outgoing AR to the StepUp Gateway, the mock gateway page that is used in the funcitonal tests now displays the received AR as an unsigned xml string. That string is used to perform xpath queries on, determiniing the presense or absence of the forceauthn flag.

See: https://www.pivotaltracker.com/story/show/184369636

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Coins.php
+++ b/src/OpenConext/EngineBlock/Metadata/Coins.php
@@ -43,7 +43,8 @@ class Coins
         $stepupRequireLoa,
         $disableScoping,
         $additionalLogging,
-        $signatureMethod
+        $signatureMethod,
+        $stepupForceAuthn
     ) {
         return new self([
             'isConsentRequired' => $isConsentRequired,
@@ -60,6 +61,7 @@ class Coins
             'signatureMethod' => $signatureMethod,
             'stepupAllowNoToken' => $stepupAllowNoToken,
             'stepupRequireLoa' => $stepupRequireLoa,
+            'stepupForceAuthn' => $stepupForceAuthn,
         ]);
     }
 
@@ -176,6 +178,15 @@ class Coins
     public function stepupRequireLoa()
     {
         return $this->getValue('stepupRequireLoa', '');
+    }
+
+    /**
+     * Should the Stepup authentication request (to the Stepup Gateway)
+     * have the ForceAuthn attribute in the AuthnRequest?
+     */
+    public function isStepupForceAuthn()
+    {
+        return $this->getValue('stepupForceAuthn', false);
     }
 
     // IDP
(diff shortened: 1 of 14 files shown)


pablothedude (approved):

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/StepupContext.php
```diff
@@ -135,6 +135,19 @@ public function spAllowsNoStepup($spName)
             ->save();
     }
 
+    /**
+     * @Given /^the SP "([^"]*)" forces authentication$/
```
thijskh, line 139: forces stepup authentication

src/OpenConext/EngineBlock/Metadata/Coins.php
```diff
@@ -178,6 +180,11 @@ public function stepupRequireLoa()
         return $this->getValue('stepupRequireLoa', '');
     }
 
+    public function isForceAuthn()
```
thijskh, line 183: If it's not too much trouble I'd love for the word Stepup to appear in the method name. This is used in various places in the code and I want it to be clear immediately that it concerns specifically ForceAuthn for the Stepup gateway, not the general concept which EB also supports.

tests/integration/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssemblerTest.php
```diff
@@ -432,4 +444,12 @@ private function validCoinValuesStringGuestQualifier() {
             ["string", "string"],
         ];
     }
+
+    private function validCoinValuesBoolFoceAuthn()
```
thijskh, line 448: s/Foce/Force

thijskh (commented):
It looks good from a first inspection. I would want to ask you to make it more explicit in some places that the code refers to ForceAuth _to the Stepup gateway_. As EB already knows about and handles the concept of ForceAuth, I think it's important to make it even more explict that this code is specifically for the stepup callout. See individual notes.

thijskh (approved):
Tested it, works as designed. If the naming could be changed as per my previous comment I'm very happy.

PR #1231, opened 2023-03-31
For stepup callout, force use of EB's default key

thijskh:
The selected keyID is normally determined by whatever the SP asked for in the AuthnRequest. However, this should not influence the callout to Stepup, as this is a direct connection between Engine and Stepup and the key that Stepup trusts will just be a single key, namely EB's default.


MKodde (approved):

tests/library/EngineBlock/Test/Corto/ProxyServerTest.php
```diff
@@ -98,6 +99,45 @@ public function testNameIDFormatIsSetFromRemoteMetaData()
...
+            new EngineBlock_X509_Certificate(openssl_x509_read(file_get_contents(__DIR__."/fixture/{$key}.pem.crt"))),
+            new EngineBlock_X509_PrivateKey(__DIR__."/fixture/{$cert}.pem.key")
+        );
+    }
+
+    public function provideKeyExpectations()
+    {
+        return [
+            'force default signing key' => [null, true, 'default'],
+            'require specific key is overridden by forcing the default' => ['rollover', true, 'rollover'],
```
thijskh, line 135: If you force the default the expection should always 'default'?

thijskh:
Yes

PR #1237, opened 2023-04-28
User's displaname is no longer used anywhere

thijskh:


MKodde (approved):
Looks good. I was  unable to grep any usages of the method, so should be safe to remove this :+1: 

Rebased this be able to merge it back to `master`

PR #1240, opened 2023-05-03
Show back to SP link on error page

MKodde:
When the preconditions are met, a Back to SP link is displayed on the error feedback page. The preconditions are described and discussed on the Pivotal story found in the link below.

The way this works is as follows: when EB sees that a response from the IdP is not successful, it throws an exception. This results in a user facing feedback page where the end user is instructed that authentication failed for a given reason (they can vary (cancelled, no auth context, ..). On that page the user is kind of stuck. But by pressing the back to sp button, the user can now travel back to the SP with the SamlFailedResponse. Allowing the SP to deal with the failed authentication instead of remain stranded at Engine.

Some what was most challenging was how to port all required data to the feedback page. You kind of need a lot of information to form a good Saml failed response. And it turned out, passing along the full response from the IdP was most convenient. This response is piggy backed to the feedback controller via the FeedbackInfo session.

Not the most elegant of solutions..

What I did to make up for that is:
1. I added a test for the SamlResponseHelper
2. Produced some additional readme content, describing how to test EngineBlock step-up authentications

See: https://www.pivotaltracker.com/story/show/176960606

### DIFF
--- a/src/OpenConext/EngineBlockBundle/Authentication/Service/SamlResponseHelper.php
+++ b/src/OpenConext/EngineBlockBundle/Authentication/Service/SamlResponseHelper.php
@@ -19,12 +19,12 @@
 namespace OpenConext\EngineBlockBundle\Authentication\Service;
 
 use DateTime;
+use EngineBlock_Saml2_ResponseAnnotationDecorator;
 use OpenConext\EngineBlock\Metadata\MetadataRepository\MetadataRepositoryInterface;
 use OpenConext\EngineBlockBundle\Exception\RuntimeException;
 use SAML2\Constants;
 use SAML2\Response as SAMLResponse;
 use SAML2\XML\saml\Issuer;
-use function sprintf;
 
 class SamlResponseHelper
 {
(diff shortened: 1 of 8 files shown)


src/OpenConext/EngineBlockBundle/Authentication/Service/SamlResponseHelper.php
```diff
@@ -38,20 +38,34 @@ public function __construct(MetadataRepositoryInterface $metaDataRepository)
...
-        $response->setIssuer($issuer);
-        $response->setIssueInstant((new DateTime('now'))->getTimestamp());
+        $response->setIssuer($originalResponse->getIssuer());
+        $response->setIssueInstant(time());
         $response->setInResponseTo($originalRequestId);
+        $status = $originalResponse->getStatus();
         $response->setStatus([
-            'Code' => Constants::STATUS_RESPONDER,
-            'SubCode' => Constants::STATUS_AUTHN_FAILED,
+            'Code' => array_key_exists('Code', $status) ? $status['Code'] : Constants::STATUS_RESPONDER,
```
pablothedude, line 55: Consider using a VO for the `$status` or consider to move this to dedicated methods on the `EngineBlock_Saml2_ResponseAnnotationDecorator`.

src/OpenConext/EngineBlockBundle/Authentication/Service/SamlResponseHelper.php
```diff
@@ -38,20 +38,34 @@ public function __construct(MetadataRepositoryInterface $metaDataRepository)
...
+            'SubCode' => array_key_exists('SubCode', $status) ? $status['SubCode'] : Constants::STATUS_AUTHN_FAILED,
             'Message' => $message
         ]);
+
+        // Copy of behavior found in: https://github.com/OpenConext/OpenConext-engineblock/blob/8aea9cdaa8162d92b391c35c7c66ce6802273f72/library/EngineBlock/Corto/ProxyServer.php#L582-L598
```
pablothedude, line 60: documentation :100:

pablothedude (approved):

src/OpenConext/EngineBlockBundle/Authentication/Service/SamlResponseHelper.php
MKodde: In this case, we are setting the status on a SAML2 library Response object. I could open a PR on that lib, and add a setter for specifically setting the authnfailed response codes. And introducing a VO in the library (where the  EngineBlock_Saml2_ResponseAnnotationDecorator lives is also out of scope as per the strangler strategy. 

In that light, we could argue that the SamlResponseHelper should not use the library `EngineBlock_Saml2_ResponseAnnotationDecorator`

pablothedude: What do you mean, because the `ResponseAnnotationDecorator` is living here?
`https://github.com/OpenConext/OpenConext-engineblock/blob/cd0657a4515843d07b8da040a90486a1fa905361/library/EngineBlock/Saml2/ResponseAnnotationDecorator.php`

PR #1244, opened 2023-05-22
Update regex for URNs

tvdijen:
RFC2141 was superseded by RFC8141 in 2017


thijskh:
Thanks. The testsuite fails. I think we need to update the testsuite if we change this method, because the urns contained in the unit test 'valid' scenario are not really corner cases. So we should besides fixing the unit test failures probably proactively add some 'valid' urns that we want to pass. I would rewrite the current test which is needlessly complex and just results in many more-or-less repeated tests.

tvdijen:
Agreed the valid cases make no sense.. It's just a huge list of fairly regular urns.
One of the 'invalid' cases was actually wrong, because `urn:test:%00` is perfectly valid (and already was under the old RFC)

> [2.3.1](https://datatracker.ietf.org/doc/html/rfc2141#section-2.3.1) The "%" character
> 
>    The "%" character is reserved in the URN syntax for introducing the
>    escape sequence for an octet.  Literal use of the "%" character in a
>    namespace must be encoded using "%25" in URNs for that namespace.
>    The presence of an "%" character in an URN MUST be followed by two
>    characters from the <hex> character set.

PR #1245, opened 2023-05-23
Remove templates that are identical to those in base/

thijskh:
The theme derives from base. Therefore a template present in the theme that is identical to base does not do anything.


thijskh:
It is unknown why apparently these files in openconext/ are required. If you use the skeune theme the metadata generation remains working after removing them.

PR #1249, opened 2023-05-25
Refuse to process incoming metadata push with 0 connections in it.

thijskh:
This would effectively empty out the database. That is never a desirable outcome, so reject it, whatever the cause.


quartje (approved):
We have tested it and everything is working as expected. Nice defence!

thijskh:
> I was searching for a doc-entry that describes the Metadata PUSH feature. But did not find it. If there is a written documentation in another repo somewhere, I'd suggest you add this constraint there.

That documentation did not exist yet, but I've taken the opportunity to create some in `docs/metadata_push.md`, including the new constraint.

PR #1252, opened 2023-06-13
Upgrade build tooling

MKodde:
Upgrade several outdated JS & CSS build tools. Different works have been undertaken;

1. Node Sass was replaced for Dart Sass
2. Node was upgraded from version 10 to 16 (moving to 18 proved
   difficult) as some of the build tools did not like it that much..
3. JS Yaml was upgraded, and deprecated function calls where updated
4. The new SASS version requires the definition of customSyntax
   definitions in the stylelint config. Also dictated by the stylelint
   config recomended package that was bumped to v4.0
5. Browserfy and Tinyfy where upgraded


theme/package.json
```diff
@@ -15,7 +15,7 @@
     "cypress:open": "npx cypress open",
     "audit": "npm audit",
     "lint": "npm run lint:js && npm run lint:css",
-    "lint:js": "jshint base/javascripts/",
+    "lint:js": "jshint **/javascripts/",
```
MKodde, line 18: This glob pattern change makes sure the `skeune` and `openconext` javascripts are also linted.

theme/scripts/build.js
```diff
@@ -16,7 +16,7 @@ const config = `${__dirname}/../../app/config/parameters.yml`;
 try {
     console.log('Reading contents of parameters.yml.\n');
     const fileContents = fs.readFileSync(config, 'utf8');
-    const parameters = yaml.safeLoadAll(fileContents);
+    const parameters = yaml.loadAll(fileContents);
```
MKodde, line 19: The removal of the `safe` prefixes is caused by the upgrade of `js-yaml`. As of version 4 the safe prefixed methods have been removed. All methods are now safe by design.

PR #1253, opened 2023-06-21
Fix WAYF placeholder.png having uninterpolated assetsVersion variable

thijskh:
The WAYF would perform requests for literally `/images/placeholder.png?v={{ assetsVersion }}`, because of this misuse of Twig variable interpolation.


MKodde (approved):
Good point! thanks for addressing (and fixing) this :+1:

PR #1259, opened 2023-08-07
Temporary fix for npm-force-resolutions

tvdijen:
See: https://github.com/rogeriochaves/npm-force-resolutions/issues/33

This may not be the proper fix, but I wasn't able to work around the issue and this actually fixed it.

NodeJS v14.21.3
NPM v6.14.18

### DIFF
--- a/theme/package.json
+++ b/theme/package.json
@@ -6,7 +6,7 @@
   "description": "Themes for OpenConext Engineblock",
   "main": "index.js",
   "scripts": {
-    "preinstall": "npx npm-force-resolutions",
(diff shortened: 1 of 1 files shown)


MKodde:
Thanks for your contribution @tvdijen 

I ran into this issue on my maintenance PR where I'm installing NPM updates. The team and us decided to swap NPM for yarn. Yarn is much more robust when it comes to installing resolutions (imho). Combine that with the fact that our other (Stepup) components all use Yarn, the choice was not that hard to make.

But this means that I'm closing the PR

PR #1263, opened 2023-08-24
JS dependency upgrades + Taking care of collateral damage

MKodde:


MKodde:
This work was merged in #1268

PR #1267, opened 2023-09-11
Move to modern BaseContainer base image

MKodde:
The 'openconext' container is now based on the BaseContainers Apache 2 PHP 7.2 Composer 2 Node 14 base image.

As the build tag script only utilizes npm & composer to build the tarbal, this image should suffice nicely.

To get a working tarbal a couple of minor tweaks were in order.

1. We for now ignore the dependency on ext_exif, could be added to the base container @quartje ?
2. To get npm in the path, the login script must be run when executing code in the container (`tag-release.yml` L22)

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Pull OpenConext build container
         run: cd docker && ${DOCKER_COMPOSE} up -d
       - name: Make the release files
-        run: cd docker && ${DOCKER_COMPOSE} exec -T openconext bash -c 'HOME=/home/runner/work/OpenConext-engineblock ./bin/makeRelease.sh ${{ steps.vars.outputs.tag }}'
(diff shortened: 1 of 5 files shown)


quartje:
Do we need ext_exif in EngineBlock development, or is it a build time dependency only?

quartje (commented):
Nice! Another project moved to OpenConext-Basecontainers. I will delete this container from OpenConext-containers

MKodde:
Thanks for your input @quartje 

Regarding the exif extension. It's a dependency of our PHPImageWorkshop dependency. I see no evidence of it being used anywhere. Maybe used in attribute manipulations? @thijskh can you shed your light on this?

The only evidence in history that I could find is that Martin added it to aid a logo generator

https://github.com/OpenConext/OpenConext-engineblock/commit/46227f5a842f1dbb114955e75601d950ea3d8899

I'd say. :wastebasket:

thijskh:
I can't think of any reason we should need this, and the codebase does not use it.

PR #1269, opened 2023-09-20
Option to add ribbon to engineblock pages indication which environmen…

thijskh:
…t you're working on

When left empty, nothing is shown.

The approach may look a bit unwieldy, but this is done so that envirnment specific config can be passed to CSS. This is making it a pure CSS solution that also would work in installs that have a CSP that disables inline styles, that's why we have to enumerate the available colors since you cannot use `attr()` with anything other than `content:`.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -77,6 +77,8 @@ twig:
         defaultTitle: "%view_default_title%"
         defaultHeader: "%view_default_header%"
         defaultLogo: "%view_default_logo%"
+        envName: "%env_name%"
+        envRibbonColor: "%env_ribbon_color%"
 
 doctrine:
     dbal:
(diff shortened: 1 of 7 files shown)


MKodde:
I like this solution. And TIL: you can safely use attr in :before and :after pseudoselectors. Caniuse first confused me that this feature would be unsafe to use, but after reading the widespread support for attr() in content selectors, my concerns where taken away.

What might be a good addition to this PR:
1. Add some user documentation on how to get the ribbon.
2. Make sure to state the colors that are currently supported.

See: https://caniuse.com/?search=css%20attr()

thijskh:
I have documented such in `parameters.yml.dist` and also created some docs in the EB wiki.

PR #1270, opened 2023-09-26
Substitue NPM for Yarn in makeRelease.sh

MKodde:
The tag-release script calls the `makeRelease.sh` script. That script was not yet made aware that we stopped using NPM. Now that it uses Yarn, we should now be able to create releases once again. 

Tested on my dev machine, and it did create a sensible Tarbal there.


bin/makeRelease.sh
```diff
@@ -78,8 +78,8 @@ fi
 # can be removed when we stop doing that
```
thijskh, line 78: This entire comment block can now also go, including the "todo" ✅

thijskh (approved):

bin/makeRelease.sh
MKodde: Fair point, I removed that comment

PR #1271, opened 2023-09-27
Request Access bugfixes

thijskh:
* The 'motivation' field seems to be actually required, since submitting the form without it filled fails (with no message). Simplest solution is to also make it required in the front end so it's clear to the user they need to fill this in. It also makes sense to require a motivation with the report.
* Mails fail to send because the From is set to the user's emailadress, which we may not be allowed to send mail as at all. Change this to the To address which we own and can send mail as.

### DIFF
--- a/src/OpenConext/EngineBlock/Service/RequestAccessMailer.php
+++ b/src/OpenConext/EngineBlock/Service/RequestAccessMailer.php
@@ -94,7 +94,7 @@ TPL;
         $message = new Swift_Message();
         $message
             ->setSubject($subject)
-            ->setFrom($email, $name)
+            ->setFrom($this->requestAccessEmailAddress)
             ->setTo($this->requestAccessEmailAddress)
             ->setBody($body, 'text/plain');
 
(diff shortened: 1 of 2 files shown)


MKodde (approved):
Looking good.

1. The validation message is now only a browser message right? Thats probably different from the messages that the other fields get? But I should start my VM to verify that. I'm totaly fine with that as long as it meets your expectations.
2. Good point on setting the FROM header to use the same address we send to. That increases chance of delivery. It however is slightly unexpected from a casual point of view. So maybe add a comment that explains why we use the samen address for both situations?

thijskh:
1. This is indeed the case. Previously it just failed silently so this is an improvement over the status quo. As indicated elsewhere I don't see the business case for this feature to spend more time for just a validation message. If someone thinks this message should be presented differently, they can do that.
2. I will add a comment. A more structural way could be to introduce a new "mail sender address" setting in EB and use that everywhere, but again, seems to much work for the amount of gain.

PR #1272, opened 2023-10-03
Move html into templates

thijskh:
Robustify the templating by moving HTML from translatable strings into the Twig templates. This also allows us to reduce the amount of `|raw` Twig filtering we use, the vast majority of templates do not use `|raw` anymore now.

This may not be fully backwards compatible: those with language string overrides for the affected strings, or with templates based on the changed strings, may need to make updates.


theme/base/templates/modules/Authentication/View/IdentityProvider/perform-request-access.html.twig
```diff
@@ -3,10 +3,10 @@
 {% block content %}
 <div class="mod-form">
   <h1>{{ 'send_confirm'|trans }}</h1>
-    {{ 'send_confirm_desc'|trans|raw }}
+    {{ 'send_confirm_desc'|trans }}
```
MKodde, line 6: Translation dropped the `<p>`aragraph wraparound. Maybe add it back in template?

theme/base/templates/modules/Authentication/View/Feedback/authn-context-class-ref-blacklisted.html.twig
```diff
@@ -7,8 +7,8 @@
 
 {% block errorMessage %}
     {% if withIdpName %}
-        {{ 'error_authn_context_class_ref_blacklisted_desc'|trans({'%idpName%': feedbackInfoMap['identityProviderName']})|raw }}
+        {{ 'error_authn_context_class_ref_blacklisted_desc'|trans({'%idpName%': feedbackInfoMap['identityProviderName']}) }}
```
MKodde, line 10: Was previously wrapped in `<p>`

theme/base/templates/modules/Authentication/View/Feedback/authn-context-class-ref-blacklisted.html.twig
```diff
@@ -7,8 +7,8 @@
...
     {% else %}
-        {{ 'error_authn_context_class_ref_blacklisted_desc_no_idp_name'|trans|raw }}
+        {{ 'error_authn_context_class_ref_blacklisted_desc_no_idp_name'|trans }}
```
MKodde, line 12: Was previously wrapped in `<p>`

theme/base/templates/modules/Authentication/View/Feedback/authorization-policy-violation.html.twig
```diff
@@ -32,13 +32,13 @@
 
     {% endif %}
     {% if withIdpName and withSpName %}
-        {{ 'error_authorization_policy_violation_desc'|trans({'%spName%': feedbackInfoMap['serviceProviderName'], '%idpName%': feedbackInfoMap['identityProviderName']})|raw }}
+        {{ 'error_authorization_policy_violation_desc'|trans({'%spName%': feedbackInfoMap['serviceProviderName'], '%idpName%': feedbackInfoMap['identityProviderName']}) }}
```
MKodde, line 35: We are losing formatting here; we could still use raw, and combine it with strip_tags as such:

`error_authorization_policy_violation_desc| trans | striptags ('<i><br><p') | raw`

If you don't see the relevance; We did add the additional markup for a (UX) reason. Maybe we should inform Arnout and the other more UI/UX focussed PO's about this change? If they don't mind, it's totaly fine with me

theme/base/templates/modules/Default/View/Error/error.html.twig
```diff
@@ -56,7 +56,7 @@
 
             <div class="error-details">
                 <p class="error-details__intro-text">
-                    {{ 'error_feedback_info_intro'|trans|raw }}
+                    {{ 'error_feedback_info_intro'|trans }}
```
MKodde, line 59: We are losing markup here. And probably more significantly than most other cases

theme/base/templates/modules/Authentication/View/Feedback/missing-required-fields.html.twig
```diff
@@ -7,12 +7,20 @@
...
-        {{ 'error_missing_required_fields_desc_no_name'|trans|raw }}
+        {{ 'error_missing_required_fields_desc_no_name'|trans }}
     {% endif %}
+
+    <p>
+        <ul>
+            <li>UID</li>
+            <li>schacHomeOrganization</li>
+        </ul>
+    </p>
```
MKodde, line 24: I think the three different conditions should include different attribute examples right?

MKodde (commented):
I see why this is an improvement. And removing the `raw` filters should prevent problems in the future with the cases that are not causing any problems right now. 

However we are losing quite some markup this way. If that is acceptable. I'm fine with this too. But we could also think of a way to make the raw filter more safe. For example by providing a safer blacklist (scrapping the script tags and other possible xss avenues).

After seeing that we are losing many `<p>` wraparounds, I stopped 1 by 1 checking each of them. So be warned there.

theme/base/templates/modules/Default/View/Error/error.html.twig
thijskh: I've inspected it and there's no discernable visible differences. So the markup was not doing anything and apparently that was acceptable before.

theme/base/templates/modules/Authentication/View/Feedback/authn-context-class-ref-blacklisted.html.twig
thijskh: This was not done consistently for the messages. I've gone through all of them and made it consistent now.

theme/base/templates/modules/Authentication/View/Feedback/authorization-policy-violation.html.twig
thijskh: It's about one emphasised term. I'm pretty sure it was added for no real reason. I've asked the PM and he agrees it's needless and can be removed.

thijskh:
I've checked the `<p>` and they were applied inconsistently before this PR already. I've now made that consistent.

I've checked and there's no objection to the small loss of markup. Where there was potentially bigger loss I've already solve that in moving the HTML into the templates.

PR #1279, opened 2023-11-20
Support overriding StepUp EntityId

MKodde:
Currently the SAML EntityID of the engineblock SP that us used to do Stepup (SFO) authenticatons to the Stepup-Gateway is always https://<engineblock.sever.domain.name>/authentication/stepup/metadata

The key used to sign the SAML AuthnRequests from this SP is the engineblock default key.

To facilitate a rolling configuration update I want the SP entityID that is used for Stepup to be configurable so that at the same time that the engineblock default key is updated, this entityID can be changed. This then allows two entities, with two different keys, to be configured in the Stepup-Gateway.

https://www.pivotaltracker.com/story/show/186200738

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,24 @@
 # UPGRADE NOTES
 
+## 6.13 -> 6.14
+Previously the SAML EntityID of the EngineBlock SP that was used to do Stepup (SFO) authentications to the Stepup-Gateway 
+always was https://<engineblock.sever.domain.name>/authentication/stepup/metadata. For these authentication the default
+EngineBlock key is always used for signing.
+
+If you'd like to key-rollover the StepUp entity (baked into EngineBlock).
+The key used to sign the SAML AuthnRequests from this SP is the engineblock default key.
+
+To facilitate a rolling configuration update I want the SP entityID that is used for Stepup to be configurable so that at the same time that the engineblock default key is updated, this entityID can be changed. This then allows two entities, with two different keys, to be configured in the Stepup-Gateway.
+
+There are two new parameters that configure this behavior.
+
+1. `feature_stepup_sfo_override_engine_entityid` [bool] enables/disables the feature. Default: disabled
+2. `stepup.sfo.override_engine_entityid` [string] should be set with the Entity ID you'd like to use for the stepup EntityId. Default: ''
+
+The feature flag was added mainly to aid our test suite to easily test this feature.
+
+By default this feature is disabled and the default Entity Id is used for the StepUp entity.
+
 ## 6.12 -> 6.13
 
 Some translatable strings have been changed and "raw" use of HTML in
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -223,6 +223,7 @@ parameters:
     feature_run_all_manipulations_prior_to_consent: false
     feature_block_user_on_violation: false
     feature_enable_consent: true
+    feature_stepup_sfo_override_engine_entityid: false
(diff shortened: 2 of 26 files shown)


UPGRADING.md
```diff
@@ -1,5 +1,24 @@
 # UPGRADE NOTES
 
+## 6.13 -> 6.14
+Previously the SAML EntityID of the EngineBlock SP that us used to do Stepup (SFO) authenticatons to the Stepup-Gateway
```
tvdijen, line 4: `was used` and `authentications`

MKodde: Thanks for the feedback! Updated the notes according to your typo-suggestions.

library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -487,6 +488,32 @@ public function sendStepupAuthenticationRequest(
...
+        $isEnabled = $features->isEnabled('eb.stepup.sfo.override_engine_entityid');
+
+        if ($isEnabled && $isConfigured) {
+            if (empty($entityIdOverrideValue)) {
+                throw new MissingParameterException(
+                    'When feature "feature_stepup_sfo_override_engine_entityid" is enabled,
+                     you must provide the "stepup.sfo.override_engine_entityid" parameter.'
+                );
+            }
+            $this->_logger->notice(
```
thijskh, line 505: I think I prefer `info` or even `debug`. I would expect this config to last for an undetermined amount of time, does not make sense to issue notices on every stepup callout about it.

thijskh (approved):

app/config/parameters.yml.dist
```diff
@@ -261,6 +262,7 @@ parameters:
     stepup.gateway.sfo.sso_location: 'https://gateway.stepup.vm.openconext.org/second-factor-only/single-sign-on'
     ## The public key from the Stepup Gateway IdP
     stepup.gateway.sfo.key_file: /etc/openconext/engineblock.crt
+    stepup.sfo.override_engine_entityid: 'https://engine.vm.openconext.com/new/stepup/metadata'
```
thijskh, line 265: The docs/stepup_callout.md file refers to this section for explanation of the Engineblock config. So I think this new setting should have at least a `#` comment like the ones above.

library/EngineBlock/Corto/Module/Service/StepupAssertionConsumer.php
```diff
@@ -138,6 +141,30 @@ public function serve($serviceName, Request $httpRequest)
...
+        $application = EngineBlock_ApplicationSingleton::getInstance();
+        $container = $application->getDiContainer();
+        $entityIdOverrideValue = $container->getStepupEntityIdOverrideValue();
+        $features = $container->getFeatureConfiguration();
+        $isConfigured = $features->hasFeature('eb.stepup.sfo.override_engine_entityid');
+        $isEnabled = $features->isEnabled('eb.stepup.sfo.override_engine_entityid');
+
+        $serviceEntityId = $this->_server->getUrl('stepupMetadataService');
+        if ($isEnabled && $isConfigured) {
+            return $entityIdOverrideValue;
```
MKodde, line 163: Setting the override causes this error down the line when EB is processing the assertion (stepupAssertionConsumerService)

`Response processing failed: "Invalid Assertion in SAML Response, errors: "The configured Service Provider [https://engine.vm.openconext.com/new/stepup/metadata] is not a valid audience for the assertion. Audiences: [https://engine.vm.openconext.org/authentication/stepup/metadata]"`

PR #1281, opened 2023-11-22
Stepup ACS: add check that returned NameID matches the one we requested.

thijskh:
This should always be the case, we request a LoA from Stepup for NameID A, then we should receive a response that contains that same NameID A. This is defense in depth, should at some point in the entire chain another check fail and an adversary be able to switch or replay assertions, this will guard against them being accepted by Engineblock as valid.


thijskh:
@phavekes

phavekes (approved):

PR #1282, opened 2023-11-29
Default feature block_user_on_violation to on

thijskh:
For new installs this makes sense - the scope check is only done when you add scopes in Manage anyway, but if you add them, it's a safer default that they are actually enforced.

The feature flag mostly exists to phase in scopes where you previously not had enabled them, so you want to start logging first.


thijskh:
Feature flag will be removed in #1298 but this seems a useful change until then nonetheless

MKodde (approved):

PR #1291, opened 2024-03-12
Proposed fix behat

thijskh:


MKodde:
@thijskh I included your fix in #1279, this concept was merged with a DI based solution.

PR #1293, opened 2024-04-15
Remove unused class

tvdijen:
I ran into this because I want to harden my PHP install by setting `allow_url_fopen` to `false`..

I've discussed this on Slack with @thijskh and the class is not being used anywhere at all..
Thijs thought this was an old feature where it was possible to enable schema-validation on messages and/or metadata, but it no longer exists.

An alternative to this removal would be to restore the feature, but use local certs from the saml2-lib, that has all of them available for use without downloading them from anywhere remote.


thijskh (approved):
Thanks, indeed this has not been doing anything for years and apparently was not missed. If it’s required in the future we can implement it in a more modern way.

PR #1298, opened 2024-05-08
Implement scope checking in the existing attribute validator

### LINKED ISSUE #1802: Error message for invalid Subjectid is confusing
If an IdP sends an invalid Subjectid (e.g., with a capital letter) EB shows the generic error for "missing attributes", which is confusing.

thijskh:
Advantages:
- makes use of existing Engineblock facility
- vastly reduces code duplication
- makes it configurable which attributes are subject to the check
- makes it configurable which attributes will block vs which will only log
- makes it possible to display scope violation in the debug page already

Todo:
- Improve exception message to indicate cause of blockage
- Update functional tests for changed error screen
- Integrate into debug page
- Remove now unused previous code


thijskh:
Replaces the current separate Scope check filters in the EB processing pipeline which are not very configurable.

Might revive this PR when I would have a working EB development environment.

baszoetekouw:
@thijskh do you think you will have time to continuing this PR?  Or would it be better to transfer the work to @johanib or @kayjoosten ?

thijskh:
I do not have a working EB environment. If this works again I can finish the PR

PR #1299, opened 2024-05-14
Make the /health and /info endpoints also available via engine-api

thijskh:
Before, the api protection came first so the endpoints provided by the monitor bundle cannot be accessed on engine-api vhost (without credentials). Change the order such that the monitor bundle endpoints are always available, both on engine and engine-api.


MKodde (approved):
Simple yet very effective!

PR #1300, opened 2024-06-04
When AM error occurs, accept any type of exception for logging

thijskh:
The exception can be an EngineBlock_Exception, but e.g. a ParseError is not unlikely to occur in practice. A ParseError would previously give a confusing Uncaugt TypeError, removing the type declaration makes it being processed and logged as a problem with the AM which is what it is.

### DIFF
--- a/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php
+++ b/library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php
@@ -102,7 +102,7 @@ class EngineBlock_Attributes_Manipulator_ServiceRegistry
                 eval($manipulationCode);
             },
             // Should an error occur, log the input, if nothing happens, then don't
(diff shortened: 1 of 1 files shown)


library/EngineBlock/Attributes/Manipulator/ServiceRegistry.php
```diff
@@ -102,7 +102,7 @@ function()
                 eval($manipulationCode);
             },
             // Should an error occur, log the input, if nothing happens, then don't
-            function(EngineBlock_Exception $exception)
+            function($exception)
```
MKodde, line 105: Nitpick: Have you considered leaving a type definition, but making it more generic.

For example, `Throwable` should fit both the Eb_Exception and the TypeError.

MKodde (approved):

Issue #1303, opened 2024-06-26
IdP metadata keywords in WAYF do not follow a language switch


cdbesten:
The WAYF provides a way to search and IdP based on the value of keywords.(language). When switching language, it seems the same keywords are kept in place.

In our scenario we have keywords.nl and keywords.en; The presented WAYF in NL or the WAYF in EU language both have the keywords.en as searchable keywords for an IdP.

PR #1304, opened 2024-06-26
Locale is considered when providing keywords to WAYF

thijskh:
When constructing the WAYF data, previously the current locale was not considered and the method would just in order use the first available of en, nl, or pt. With this change, Engine will first see if there are keywords for the current locale of the WAYF, if not, use en as a fallback, otherwise return Undefined as before.

Closes: #1303


cdbesten:
Have applied this patch in our environment and found it to be working just fine!

MKodde (approved):
Nice improvement!

Issue #1302, opened 2024-06-18
Release an attribute under another name


thijskh:
EB uses standardised attribute names. However, sometimes an SP requires the attribute to be delivered under a different name. We usually use an attribute manipulation for that but want to solve this more common case in configuration.

Manage has been extended to provide a string field in the ARP where one can define the name under which an attribute must be released instead of the normal name.

```json
   "arp_attributes" : {
      "urn:mace:dir:attribute-def:cn" : [
         {
            "release_as" : "some-example-name",
            "motivation" : "Hiervoor wordt het gebruikt",
            "value" : "*"
         }
      ]
   },
```

When this `release_as` key is present, EB will release the values of the attribute as-is, but no longer under e.g. `urn:mace:dir:attribute-def:cn` but rather under the name `some-example-name`.

Probably this is best done in either the ARP Corto filter or a new Corto output filter for this purpose.

MKodde:
Will Manage always release a `urn:mace:..` attribute substitution name? IIRC we also support the `urn:oid` format. Must we support that too? Or does Manage maybe indicate which format to use (aside from implicitly doing so by the value it releases)?

thijskh:
No, the idea is that the SP can require something nonstandard, e.g. just the bare string `FullName` or `mail`, instead of properly namespaced/standardised names.

thijskh:
To illustrate for one SP, we now have Attribute Manipulation code configured that does this:

```php
# Required attributes
 
$attr_gn       = 'urn:mace:dir:attribute-def:givenName';
$attr_sn       = 'urn:mace:dir:attribute-def:sn';
$attr_mail     = 'urn:mace:dir:attribute-def:mail';
 
# attributes to let through (ARP)
$requiredAttributes = array(
  'FirstName',
  'LastName',
  'Email'
);
 
if (isset($attributes) and ($attributes !== FALSE)) {
    if (!empty($attributes[$attr_mail][0])) {
        $subjectId = $attributes[$attr_mail][0];
        $attributes['Email'] = $attributes[$attr_mail];
    }
    if (!empty($attributes[$attr_gn])) {
        $attributes['FirstName'] = $attributes[$attr_gn];
    }
    if (!empty($attributes[$attr_sn])) {
        $attributes['LastName'] = $attributes[$attr_sn];
    }
}

# Remove all other attributes
foreach ($attributes as $k => $v) {
  if (!in_array($k, $requiredAttributes)) {
[...]

PR #1307, opened 2024-08-13
Create Corto input filter for the ARP directive `release_as` to be applied

MKodde:
This PR reads the ARP for the SP that is being logged in to. If the ARP contains `release_as` directions. The new input filter will apply that rule. For details, see the issue, describing this feature.

In this PR the ARP class representing the ARP settings for an SP was also modified to the `use_as_nameid`. That is considered boyscout work as the next PR will apply that ARP rule.

Fixes: #1302


thijskh:
Looks good. But I think it might need to be an output filter. Because if you e.g. release urn:mace:dir:attribute-def:mail as "EmailAddress", the consent screen will not know the latter attribute and can not provide e.g. a nice name for it. So I'm guessing it will work better if the translation of the attribute name is done after consent...

Issue #1301, opened 2024-06-18
Release a given regular attribute in the NameID


thijskh:
The SAML Subject NameID is set by Engineblock based on the SP's configuration to be a value generated by Engineblock, either persistent, transient or unspecified ID (collabPersonID).

Regularly services require the NameID to have the value of some specific other attribute that is released by the IdP, not generated by Engineblock. Most commonly email adress, eduPersonPrincipalName or uid. We now solve this with an attribute manipulation. However, we'd like EB to handle this in configuration for us.

A change has been made in Manage so that under the ARP for a given SP you can mark it as "release this attribute as name-id".

```json
   "arp_attributes" : {
      "urn:mace:dir:attribute-def:cn" : [
         {
            "use_as_name_id" : true,
            "motivation" : "Hiervoor wordt het gebruikt",
            "value" : "*"
         }
      ]
   },
```

When set to true the _first_ value of `urn:mace:dir:attribute-def:cn` will be used as the value for NameID. The NameID format can be set to the "unspecified" urn.

I propose this logic in the exising Corto output filter that handles NameID setting:
- If this is set for an attribute, release this attribute and set NameIDFormat to unspecified
- Otherwise use the NameIDFormat to determine value and format for any NameID, as it is now

Further considerations:
* When `use_as_name_id` is missing it needs to be assumed false
* Manage will guarantee that not more than one attribute in a given ARP will have set this to true.

MKodde:
Sounds like a great way to further eliminate the use case for attribute aggregation. With the implementation of this feature, can we start thinking about turning AA off? Or are there other important use cases left to keep it around?

PR #1308, opened 2024-08-15
Overwrite the NameID when specified in use_as_nameid ARP setting

MKodde:
This PR aims to close the feature described in issue: #1301 For functional details, please study the well described feature there.

For this PR some noteworthy things passed:

1. A small boyscout task was performed in commit b82468c
2. I tried to make things solid as possible. But the library code is challenging in that regard. But I tried to make up for that by adding as many tests as possible.

### DIFF
--- a/library/EngineBlock/Arp/NameIdSubstituteResolver.php
+++ b/library/EngineBlock/Arp/NameIdSubstituteResolver.php
@@ -1,0 +1,48 @@
+<?php
+
+use OpenConext\EngineBlock\Metadata\AttributeReleasePolicy;
+use Psr\Log\LoggerInterface;
+
+/**
+ * Copyright 2024 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class EngineBlock_Arp_NameIdSubstituteResolver
+{
+    /**
+     * @var LoggerInterface 
+     */
+    private $logger;
+
+    public function __construct(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    public function findNameIdSubstitute(AttributeReleasePolicy $arp, array $responseAttributes): ?string
+    {
+        $substituteAttribute = $arp->findNameIdSubstitute();
+        if ($substituteAttribute !== null && array_key_exists($substituteAttribute, $responseAttributes)) {
+            $this->logger->notice(
+                sprintf(
+                    'Found a NameId substitute ("use_as_nameid", %s will be used as NameID)',
+                    $substituteAttribute
+                )
+            );
+            return reset($responseAttributes[$substituteAttribute]);
+        }
+        return null;
+    }
+}
--- a/library/EngineBlock/Corto/Filter/Command/AddIdentityAttributes.php
+++ b/library/EngineBlock/Corto/Filter/Command/AddIdentityAttributes.php
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+use OpenConext\EngineBlock\Metadata\AttributeReleasePolicy;
 use Psr\Log\LoggerInterface;
 use SAML2\Constants;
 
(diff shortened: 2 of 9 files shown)


library/EngineBlock/Arp/NameIdSubstituteResolver.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public function __construct(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
+    }
+
+    public function findNameIdSubstitute(AttributeReleasePolicy $arp, array $responseAttributes): ?string
+    {
+        $substituteAttribute = $arp->findNameIdSubstitute();
+        if ($substituteAttribute !== null && array_key_exists($substituteAttribute, $responseAttributes)) {
```
parijke, line 37: I would suggest to encapsulate this complex if in a descriptive private method like isValidAttibute

src/OpenConext/EngineBlock/Metadata/AttributeReleasePolicy.php
```diff
@@ -139,6 +140,9 @@ public function findNameIdSubstitute(): ?string
         foreach ($this->attributeRules as $name => $rules) {
             foreach ($rules as $rule) {
                 if (isset($rule['use_as_nameid']) && $rule['use_as_nameid'] === true) {
+                    if (array_key_exists('release_as', $rule)) {
+                        return $rule['release_as'];
+                    }
```
parijke, line 145: Can this be combined in a local method like ruleHasReleaseAs or so?

```php
if ($this->ruleHasReleaseAs($rule)) {
    return $rule['release_as'];
}
```

parijke (approved):
Nice job again. I set some suggestions. Feel free to adapt or ignore them

library/EngineBlock/Corto/Filter/Command/AddIdentityAttributes.php
```diff
@@ -66,16 +73,25 @@ public function execute()
...
+            $arpSubstitute = $this->substituteResolver->findNameIdSubstitute($arp, $this->getResponseAttributes());
+            if ($arpSubstitute !== null) {
+                $nameId->setFormat(Constants::NAMEID_UNSPECIFIED);
+                $nameId->setValue($arpSubstitute);
+                $this->_response->getAssertion()->setNameId($nameId);
+            }
+
+            // Find out if the EduPersonTargetedId is in the ARP of the destination SP.
+            // If the ARP is NULL this means no ARP = let everything through including ePTI.
+            // Otherwise, only add ePTI if it's actually in the ARP.
```
thijskh, line 88: I think this comment needs to be reworded because the logic is now spread to other places.

Maybe here something like: if there's an ARP, but it does not contain the EPTI, we're done now.

And below: We arrive here if either: 1) the ARP is NULL, this means no ARP = let everything through including ePTI; or 2) the ARP is not null and does contain the EPTI attribute. In both cases, set the EPTI attribute.

An alternative is to change the logic and keep everything related to EPTI together. So remove this condition and return here, but check for ARP === null OR ARP contains EPTI below.

library/EngineBlock/Corto/Filter/Command/AddIdentityAttributes.php
```diff
@@ -66,16 +73,25 @@ public function execute()
...
         $this->_response->getAssertion()->setNameId($nameId);
 
-
-        // Find out if the EduPersonTargetedId is in the ARP of the destination SP.
-        // If the ARP is NULL this means no ARP = let everything through including ePTI.
-        // Otherwise only add ePTI if it's acutally in the ARP.
         $arp = $destinationMetadata->getAttributeReleasePolicy();
-        if (!is_null($arp) && !$arp->hasAttribute(Constants::EPTI_URN_MACE)) {
-            return;
+        if (!is_null($arp)) {
```
thijskh, line 77: Should we go through the trouble to resolve a nameid above if we're going to overwrite it anyway?

library/EngineBlock/Corto/Filter/Command/AddIdentityAttributes.php
MKodde: Thanks for the suggestion. I think improving the inline doc is a nice improvement. Moving the logic to the input filter will not result in the desired outcome. This is what I tried first. But then the AddIdentityAttributes output filter would overwrite the NameId again.

Or do you suggest another solution?

library/EngineBlock/Corto/Filter/Command/AddIdentityAttributes.php
MKodde: I looked at it like this: 

1. We need to set the nameId to the resolved value.
2. We might need to overwrite it with the substitute value. 

What could be done is to move the NameIdResolver logic down below, and only resolve the nameId if we did not use the substitute.

library/EngineBlock/Corto/Filter/Command/AddIdentityAttributes.php
thijskh: I think just a very clear comment like there's now suffices to alleviate the concern of understandability.

PR #1317, opened 2024-09-02
Improve error message on duplicate SP entityID in PUSH

thijskh:
Make it clear where this occurs (MFA list) and mention the ID that's duplicated


MKodde (approved):
That makes reviewing the error message much more intuitive!

Issue #1321, opened 2024-09-04
Warning triggered when release_as set for attribute without values


thijskh:
1. Set in the ARP the "release_as" setting for attribute A
2. Log in with an IdP that does not release any value for "A"
3. The consume assertion endpoint from EB logs a PHP warning:
```
 Warning: Invalid argument supplied for foreach() in /opt/openconext/OpenConext-engineblock-6.15.0-beta2/vendor/simplesamlphp/saml2/src/SAML2/Assertion.php on line 1701
```

The problem being that the attribute does not exist. But is 'created' as an empty/null attribute in the ReleaseAsEnforcer. So the ReleaseAsEnforcer probably also needs to check if the source attribute is actually present in this login before doing anything with it at all. If it's not there, it does not need to get another name.

PR #1322, opened 2024-09-05
Prevent adding empty renamed attribute

MKodde:
When Manage configures an override for an attribute name. But the targetted attribute is not in the assertion, then Engine would add an empty attribute with the overridden value to the assertion. The SAML2 lib would then warn us that an empty attribute value can not be processed.

To fix that. I added a test statement that verifies if the targetted attribute is present in the assertion. If not, the substitution is not made and a warning is logged.

See, Fixes: https://github.com/OpenConext/OpenConext-engineblock/issues/1321

### DIFF
--- a/src/OpenConext/EngineBlock/Service/ReleaseAsEnforcer.php
+++ b/src/OpenConext/EngineBlock/Service/ReleaseAsEnforcer.php
@@ -37,6 +37,29 @@ class ReleaseAsEnforcer implements ReleaseAsEnforcerInterface
     {
         foreach ($releaseAsOverrides as $oldAttributeName => $overrideValue) {
             $newAttributeName = $overrideValue[0]['release_as'];
+            if (!array_key_exists($oldAttributeName, $attributes)) {
+                $this->logger->notice(
+                    sprintf(
+                        'Releasing "%s" as "%s" is not possible, "%s" is not in assertion',
+                        $oldAttributeName,
+                        $newAttributeName,
+                        $oldAttributeName
+                    )
+                );
+                continue;
+            }
+            if (is_null($attributes[$oldAttributeName])) {
+                $this->logger->warning(
+                    sprintf(
+                        'Releasing "%s" as "%s" is not possible, value for "%s" is null',
+                        $oldAttributeName,
+                        $newAttributeName,
+                        $oldAttributeName
+                    )
+                );
+                unset($attributes[$oldAttributeName]);
+                continue;
+            }
             $attributeValue = $attributes[$oldAttributeName];
             unset($attributes[$oldAttributeName]);
             $this->logger->notice(
--- a/tests/unit/OpenConext/EngineBlock/Service/ReleaseAsEnforcerTest.php
+++ b/tests/unit/OpenConext/EngineBlock/Service/ReleaseAsEnforcerTest.php
@@ -53,8 +53,18 @@ class ReleaseAsEnforcerTest extends TestCase
 
         foreach ($releaseAsOverrides as $oldName => $override) {
             $this->assertArrayNotHasKey($oldName, $result);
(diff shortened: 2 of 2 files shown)


src/OpenConext/EngineBlock/Service/ReleaseAsEnforcer.php
```diff
@@ -37,6 +37,29 @@ public function enforce(array $attributes, array $releaseAsOverrides)
...
+            if (empty($attributes[$oldAttributeName])) {
+                $this->logger->warning(
+                    sprintf(
+                        'Releasing "%s" as "%s" is not possible, value for "%s" is empty',
+                        $oldAttributeName,
+                        $newAttributeName,
+                        $oldAttributeName
+                    )
+                );
+                unset($attributes[$oldAttributeName]);
```
MKodde, line 60: @thijskh in this testcase; edupersontargetedId is in the assertion but is set to be an empty array as value.

The enforcer now tests if the value is empty, and pops it out of the assertion. Is that expected behavior for Engine? I think, prior to this change. EB would allow this.

thijskh: Indeed. The problem only occurs when the attribute is completely unset. So I would limit the solution to that only.

src/OpenConext/EngineBlock/Service/ReleaseAsEnforcer.php
```diff
@@ -37,6 +37,29 @@ public function enforce(array $attributes, array $releaseAsOverrides)
     {
         foreach ($releaseAsOverrides as $oldAttributeName => $overrideValue) {
             $newAttributeName = $overrideValue[0]['release_as'];
+            if (!array_key_exists($oldAttributeName, $attributes)) {
+                $this->logger->warning(
```
thijskh, line 41: I think this is in principle also 'normal' behaviour. The ARP specification of release_as basically should be read as "If this attribute is released, then use this name". So if it's not present it should not be causing warning level messages. Info is probably fine.

src/OpenConext/EngineBlock/Service/ReleaseAsEnforcer.php
```diff
@@ -37,6 +38,29 @@ public function enforce(array $attributes, array $releaseAsOverrides)
...
+                    sprintf(
+                        'Releasing "%s" as "%s" is not possible, "%s" is not in assertion',
+                        $oldAttributeName,
+                        $newAttributeName,
+                        $oldAttributeName
+                    )
+                );
+                continue;
+            }
+            if (is_null($attributes[$oldAttributeName])) {
```
mharte-ib, line 52: What happens when the attribute is set, but the value is an empty string?

tests/unit/OpenConext/EngineBlock/Service/ReleaseAsEnforcerTest.php
```diff
@@ -53,10 +53,21 @@ public function testEnforce($attributes, $releaseAsOverrides, $expectedResult, $
...
         }
     }
 
+
+    /**
+     * @dataProvider enforceDataProviderWarnings
+     */
+    public function testEnforceImpossible($attributes, $releaseAsOverrides, $expectedResult, $expectedLogMessage)
+    {
+
```
mharte-ib, line 65:
```suggestion
```

src/OpenConext/EngineBlock/Service/ReleaseAsEnforcer.php
MKodde: Then the assertion will have that attribute with the empty string value. That is behavior that should not be tackled here.

mharte-ib (approved):

PR #1329, opened 2024-10-08
Do not claim something has been sent to any administator

thijskh:
There's no inherent guarantee in EB  whatsoever that this error type will be sent to any administrator. Do not let people believe that it will be and wait for it to magically go away if they do nothing.


baszoetekouw (approved):
Definitely an improvement in the text, although I do think we should have more automated checks for these kinds of unexpected exceptions.

PR #1337, opened 2024-11-08
Docker: Log to json, fixes multiline logging in Docker

quartje:
Since Docker logs to stdout, the ident to distinguish between the authentication and application logs is no longer applicable.  
Furthermore, logging using the regular formatters lead to all logs landing in a single line, making it hard to read and distinguish. It also hits the 16KB logline length of Docker
Using the syslog formatter fixes those issues.


thijskh:
Is the information added by the AdditionalInfoFormatter also preserved in the JSON?

thijskh (approved):
Test showed that after this change the additional info is indeed still present

PR #1339, opened 2024-11-19
Bugfix: stricter regex; disallow trailing newline

tvdijen:
To prevent false positives, we can make use of the [PCRE_DOLLAR_ENDONLY](https://www.php.net/manual/en/reference.pcre.pattern.modifiers.php) flag (`D`) that will constraint the `$` anchor to match the end of the string only, not a newline character.


baszoetekouw:
Looks good to me.  Maybe for completeness, you could add a test that starts with a newline (should work with the existing regex, if the docs are right about the behaviour of `\A`).

Issue #1343, opened 2024-11-26
Stepup related logging shows wrong key_id


quartje:
I am not quite sure what it is that is logged in the key_id in stepup related logs. eg:
```
{"channel":"app","level":"WARNING","message":"After Stepup authentication update received LoA","context":{"key_id":"_2B40C7186DC7624C4804DE96F9454826","result":{}},"extra":{"session_id":"teia448dodmj9ea57sqegcjf4k","request_id":"6745867f3dd09"}}
```

The key_id for other logs is the id of the SAML signing key.

tvdijen:
Could this be related to https://github.com/OpenConext/OpenConext-engineblock/issues/1241 ?

PR #1344, opened 2024-11-26
Remove confusing key_id from stepup callout logging

thijskh:
This value is not the keyID but the original authnrequest ID. This is superfluous as EB log messages already log tracing information in the session_id and request_id parameters that are logged with every message. Just leave out this confusingly named value.

Closes: #1343


quartje (approved):
Removing is indeed the best strategy!

PR #1781, opened 2025-01-09
Update Devconf installation and docs

pablothedude:
Use Devconf settings in the parameters.yml.dist, so devconf will work out-of-the-box for development. Also remove the old StepupDeploy VM parameters.yml. And update the readme to reflect the use of devconf.

https://github.com/OpenConext/OpenConext-devconf/pull/29


ci/qa/all.sh
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+cd $(dirname $0)/../../
+
+./ci/qa/phpmd.sh
+./ci/qa/phpcs.sh
+./ci/qa/docheader.sh
+./ci/qa/phpunit.sh
+./ci/qa/behat.sh
+./ci/qa/lint.sh
```
johanib, line 11: Do you know if the e2e cypress tests are functional? 
Can they be added here?

app/AppKernel.php
```diff
@@ -60,17 +60,19 @@ public function registerContainerConfiguration(LoaderInterface $loader)
...
     public function getLogDir()
     {
         return $this->rootDir . '/logs/' . $this->environment;
     }
+
+    public function getProjectDir()
+    {
+        // This is needed to define the project dir without composer.json
+        // @see: https://symfony.com/doc/current/reference/configuration/kernel.html#configuration-kernel-project-directory
+        return \dirname(__DIR__);
```
johanib, line 75: Weird this is now needed :thinking:

johanib (approved):
LGTM

app/AppKernel.php
pablothedude: Because in prod composer.json is not in the release and that file is used to determine this location.
But because the same `parameters.yaml.dist` file is used this is needed to get the same path to the `attributes.json` configuration.

ci/qa/all.sh
pablothedude: The default Behat test will work but not the Cypress Behat tests. I prefer to do that in a separate PR, but at least it's now possible to have the most important dev stuff available.

johanib:
Not sure if it should be in this repo or in in devconf, but maybe update the instructions on how to test?

For example, in docs/php_testing.md add:
* point to qa/ci/all.sh
* How to put the env in ci mode?

PR #1790, opened 2025-01-27
Set width of debug page to browser width

cdbesten:
The current width is too small; allow it to be as wide as the current browser size.

### DIFF
--- a/theme/base/stylesheets/components/old-not-converted/header.scss
+++ b/theme/base/stylesheets/components/old-not-converted/header.scss
@@ -16,7 +16,7 @@
     }
 
     .l-container {
-        max-width: 75rem;
+        max-width: 100vw;
         margin-bottom: 0;
     }
 


thijskh:
This sets the width of the container for the _header_ to 100vw, I don't think that has a relation to your goal?

cdbesten:
No, that doesn't. But damn .... what an enormous amount of css includes. Maybe someone with more UX experience can figure out which l-container needs to be wider ;)

cdbesten:
So do we have an UX engineer who understands all the CSSes of EB  which in the end make up material.css

sed -i 's/l-container.wide{max-width:75rem}/l-container.wide{max-width:100vw}/' web/stylesheets/material.css

This will do what I want :)

thijskh:
I've pushed that to `main` now.

cdbesten:
Thanks.

Issue #1328, opened 2024-10-01
Explicit check if signing certificate of SAML response is known


baszoetekouw:
For debuggability, it would be nice if EB could explicitly check if the signing certificate of an incoming SAML Response is trusted for this specific IdP.  Hopefully :-) this is already checked, s it should be a matter of adding an explicit error.

Currently, we only get a signature verification failure, which, while correct, describes multipe failures (i.e., wrong certificate used to sign, or bad signature in Response)

pablothedude:
The validation is handled in the SAML 2 library. In EB we only supply the SP and IdP config for the SAML2 library and catch exceptions and handle that further.
https://github.com/OpenConext/OpenConext-engineblock/blob/main/library/EngineBlock/Corto/Module/Bindings.php#L492
The SAML2 library uses XMLseclib under the hood to verify the signature and that throws the same exception in all cases.
 https://github.com/simplesamlphp/saml2/blob/v4.16.14/src/SAML2/Utils.php#L196
I presume that's deliberately done because the same is applied in the new v5 version of the SAML2 library.

pablothedude:
It seems a lot of work for a minor gain so after consultation with @baszoetekouw we decided to close this issue.

pablothedude:
After IRL discussion I started and implemented the functionality needed. Now some basic validation is added to Corto. 
The validation is done on multiple places, initial validation in the controller with some basic sanity checks, then the validation in Corto where this is added and later on real in the SAML2 library where the real validation is done with the signature digest. So there are a lot of different places the validation is done.

There is currently one outstanding question on how much keys a response can have and should be validated to support all EB flows. If that's clear I'll finish this.

baszoetekouw:
Works: 

<img width="1033" height="777" alt="Image" src="https://github.com/user-attachments/assets/738d4cc6-ba32-4ba7-98d8-f935f850655e" />

PR #1791, opened 2025-01-27
Add explicit IdP signing key feedback

pablothedude:
An explicit IdP signing key check is added for easier debugging.

The ease of debugging should outweigh the drawbacks of moving the responsibility of the SAML2 library also to EB and the performance penalties for doing this twice.
https://github.com/OpenConext/OpenConext-engineblock/issues/1328

### DIFF
--- a/languages/messages.pt.php
+++ b/languages/messages.pt.php
@@ -193,6 +193,8 @@ HTML
     'error_unknown_identity_provider_no_idp_name'   => 'Erro - %organisationNoun% desconhecido',
     'error_unknown_identity_provider_desc'     => 'O %organisationNoun% a que pretende autenticar-se é desconhecido para a %suiteName%.',
     'error_unknown_identity_provider_desc_no_idp_name'     => 'O %organisationNoun% a que pretende autenticar-se é desconhecido para a %suiteName%.',
+    'error_unknown_signing_key' => 'Erro - chave de assinatura desconhecida',
+    'error_unknown_signing_key_desc' => 'A chave de assinatura utilizada não é conhecida por %suiteName%. Isto pode ser um erro de configuração.',
     'error_generic'                     => 'Erro - Ocorreu um erro',
     'error_generic_desc'                => 'A sua autenticação falhou e não sabemos exactamente porquê. Tente de novo e no caso de voltar a não funcionar, entre em contacto com o suporte da %idpName% para pedir ajuda.',
     'error_missing_required_fields'     => 'Erro - Campo necessário em falta',
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -342,6 +342,11 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
         // check the IssueInstant against our own time to see if the SP's clock is getting out of sync
         $this->_checkIssueInstant( $sspResponse->getIssueInstant(), 'IdP', $idpEntityId );
 
+        // This is an initial check to validate when a signing key is in the
+        // response if it's the one that's expected by EB for that IdP,
+        //  so a decent error message can be shown if this is not the case.
+        $this->assertKnownIdPSigningKey($cortoIdpMetadata, $sspResponse);
+
         try {
             // 'Process' the response, verify the signature, verify the timings.
             if ($this->hasEncryptedAssertion($sspResponse)
@@ -885,5 +890,45 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
             }
         }
         return $hasEncryptedAssertion;
+    }
+
+    /**
+     * This is an initial check to validate when a signing key is in the
+     * response, if it's the one that's expected by EB for that IdP,
+     * so a decent error message can be shown if this is not the case.
+     */
+    private function assertKnownIdPSigningKey(IdentityProvider $metadata, Response $sspResponse): void {
+
+        // If no key is in the response we don't need to validate if it's known because the
+        // validation if the signature is present is the responsibility of the SamlResponseValidator.
+        if (count($sspResponse->getCertificates()) === 0) {
+            return;
+        }
+
+        if (count($sspResponse->getCertificates()) > 1) {
+            $this->_logger->notice(
+                sprintf(
+                    'Tried to verify a message from issuer "%s", but requests with multiple signing keys are not supported.',
+                    $metadata->entityId
+                )
+            );
+
+            throw new EngineBlock_Corto_Module_Bindings_Exception(sprintf('Multiple signing keys supplied for issuer: %s', $metadata->entityId));
+        }
+
+        foreach ($metadata->certificates as $metaDataCertificate) {
+            if ($metaDataCertificate->toCertData() === $sspResponse->getCertificates()[0]) {
+                return;
+            }
+        }
+
+        $this->_logger->notice(
+            sprintf(
+                'Tried to verify a message from issuer "%s", but the expected signing key does not match.',
+                $metadata->entityId
+            )
+        );
+
+        throw new EngineBlock_Corto_Exception_UnknownIdentityProviderSigningKey('Unknown signing key supplied for issuer', $metadata->entityId);
     }
 }
--- a/theme/base/templates/modules/Authentication/View/Feedback/unknown-signing-key.html.twig
+++ b/theme/base/templates/modules/Authentication/View/Feedback/unknown-signing-key.html.twig
@@ -1,0 +1,12 @@
+{% extends '@theme/Default/View/Error/error.html.twig' %}
+
+{% set pageTitle = 'error_unknown_signing_key'|trans %}
+{% block pageTitle %}
+    {{ 'error_unknown_signing_key'|trans }}
+{% endblock %}
+{% block title %}{{ parent() }} - {{ pageTitle }} {% endblock %}
+{% block pageHeading %}{{ pageTitle }}{% endblock %}
+
+{% block errorMessage %}
+    {{ 'error_unknown_signing_key_desc'|trans }}
+{% endblock %}
--- a/theme/openconext/templates/modules/Authentication/View/Feedback/unknown-signing-key.html.twig
+++ b/theme/openconext/templates/modules/Authentication/View/Feedback/unknown-signing-key.html.twig
@@ -1,0 +1,12 @@
+{% extends '@theme/Default/View/Error/error.html.twig' %}
+
+{% set pageTitle = 'error_unknown_signing_key'|trans %}
+{% block pageTitle %}
+    {{ 'error_unknown_signing_key'|trans }}
+{% endblock %}
+{% block title %}{{ parent() }} - {{ pageTitle }} {% endblock %}
+{% block pageHeading %}{{ pageTitle }}{% endblock %}
+
+{% block errorMessage %}
+    {{ 'error_unknown_signing_key_desc'|trans }}
+{% endblock %}
--- a/languages/messages.en.php
+++ b/languages/messages.en.php
@@ -197,6 +197,8 @@ HTML
     'error_unknown_identity_provider_no_idp_name'   => 'Error - Unknown %organisationNoun%',
     'error_unknown_identity_provider_desc'     => '%idpName%, which you are trying to log in with, is unknown to %suiteName%.',
     'error_unknown_identity_provider_desc_no_idp_name'     => 'The %organisationNoun% you are trying to log in with is unknown to %suiteName%.',
+    'error_unknown_signing_key' => 'Error - unknown signing key',
+    'error_unknown_signing_key_desc' => 'The signing key used is not known to %suiteName%. This is possibly a configuration error.',
     'error_generic'                     => 'Error - An error occurred',
     'error_generic_desc'                => 'Logging in has failed and we don\'t know exactly why. Please try again first by going back to %spName% and logging in again. If this doesn\'t work, please contact the service desk of %idpName%.',
     'error_generic_desc_no_sp_name' => 'Logging in has failed and we don\'t know exactly why. Please try again first by going back to the service and logging in again. If this doesn\'t work, please contact the service desk of %idpName%.',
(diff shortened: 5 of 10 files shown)


library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -886,4 +891,37 @@ private function hasEncryptedAssertion(Response $sspResponse)
...
+
+        // If no key is in the response we don't need to validate if it's known because the
+        // validation if the signature is present is the responsibility of the SamlResponseValidator.
+        if (count($sspResponse->getCertificates()) == 0) {
+            return;
+        }
+
+        $validKey = false;
+        foreach ($metadata->certificates as $metaDataCertificate) {
+            if ($metaDataCertificate->toCertData() === $sspResponse->getCertificates()[0]) {
```
thijskh, line 910: This always checks only the first certificate. Is it certain that there never will be > 1 certificate?

pablothedude: Hi Thijs, that was something I was wondering too. Because it seemed to me that a response would only be signed by one key. And therefore I wanted to ask @baszoetekouw if and when this would happen. But you beat me to it.

pmeulen: The XML signature standard does not limit the number of KeyInfo or X509Data elements in a Signature. So there could be any number, including none according to XML Signature.

The purpose of a KeyInfo is to specify what key was used to sign, an thus what key must be used to verify the signature: "KeyInfo indicates the key to be used to validate the signature." In my opinion it does not make sense to have keyInfo pointing to multiple keys.

This is how the Stepup SAML Bundle handles this: https://github.com/OpenConext/Stepup-saml-bundle/blob/main/src/SigningSignatureVerifier.php#L40

I think it is unlikely that an IdP or SP would be putting more than one KeyInfo with an X509Data element, or more than X509Data elements in a KeyInfo in a Signature. I've never seen this happen, but I cannot rule out that someone out there is doing this.

baszoetekouw: @pmeulen your link is broken, unfortunately?  Were you referring to https://github.com/OpenConext/Stepup-saml-bundle/blob/main/src/Signing/SignatureVerifier.php ?  The flow there is simpler than what we propose here.  Afaics, Stepup simply tries to verify the signature with all trusted keys; it doesn't seem to differentiate between the wrong key being specified in the assertion (i.e., unknown key in `<KeyInfo>`) or any other type of signature validation failing.

Anyway, I have been trying to track down in the SAML2 code where the signing certificates are added, but I'm having trouble finding it.  It's not clear for example which `<ds:Signature>`s exactly are considered.   
The SAML spec at least says that multiple elements in the Response can be signed, and that the signature then applies to all descendent element, so that suggests that it considers (for example) both the Response signature and the Assertion signature.

So it's probably best to iterate over all certificates returned by `$sspResponse->getCertificates()`. The question remains of course, of whether to raise an error if _all_ of them are unknown, or if _any_ of them is unknown.  The latter sound safer (and might defend against wrapping attacks even), but might cause failures for SAML Reponses from obscure IdPs.
@thijskh @pmeulen what do you think?

baszoetekouw: I think we should err of the safe side, and check all of the signatures.  If even one of them is unknown, reject the message.
Any normal message should pass this test; an IdP that does something strange like signing a message with two different keys should be investigated anyway.

pmeulen: Yes I meant https://github.com/OpenConext/Stepup-saml-bundle/blob/main/src/Signing/SignatureVerifier.php#L40

pmeulen: I'm a fan of being strict. 
- Only allow Signatures element in normal locations (on Assertion, on SAMLRequest)
- Only one SignatureValue per Signature
- Only one KeyInfo of the X509Data per SignatureValue

If we're think this breaks things, we make these checks log a warning at first, and run this in production for a wile. Then we investigate any warnings that we see, and decide what to do.

But I don't expect this to be an issue. See e.g. the checks that the (new) SSP library does:

https://github.com/simplesamlphp/xml-security/blob/feb95e14e9c9941681ac6d2fc5880691b61afff7/src/XML/ds/Signature.php#L114-L148

pablothedude:
I've added a check to prevent more than one certificate because that doesn't seem to occur.

languages/messages.pt.php
```diff
@@ -193,6 +193,8 @@
     'error_unknown_identity_provider_no_idp_name'   => 'Erro - %organisationNoun% desconhecido',
     'error_unknown_identity_provider_desc'     => 'O %organisationNoun% a que pretende autenticar-se é desconhecido para a %suiteName%.',
     'error_unknown_identity_provider_desc_no_idp_name'     => 'O %organisationNoun% a que pretende autenticar-se é desconhecido para a %suiteName%.',
+    'error_unknown_signing_key' => 'Error - unknown signing key',
```
johanib, line 196: How does this work? Will PT translations be added later?

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -342,6 +342,11 @@ public function receiveResponse($serviceEntityId, $expectedDestination)
         // check the IssueInstant against our own time to see if the SP's clock is getting out of sync
         $this->_checkIssueInstant( $sspResponse->getIssueInstant(), 'IdP', $idpEntityId );
 
+        // This is an initial check to validate when a signing key is in the
+        // response if it's the one that's expected by EB for that IdP,
+        //  so a descent error message can be shown if this is not the case.
```
johanib, line 347:
```suggestion
        //  so a decent error message can be shown if this is not the case.
```

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -886,4 +891,48 @@ private function hasEncryptedAssertion(Response $sspResponse)
         }
         return $hasEncryptedAssertion;
     }
+
+    /**
+     * This is an initial check to validate when a signing key is in the
+     * response if it's the one that's expected by EB for that IdP,
```
johanib, line 897:
```suggestion
     * response, if it's the one that's expected by EB for that IdP,
```

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -886,4 +891,48 @@ private function hasEncryptedAssertion(Response $sspResponse)
...
+     * so a descent error message can be shown if this is not the case.
```
johanib, line 898:
```suggestion
     * so a decent error message can be shown if this is not the case.
```

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -886,4 +891,48 @@ private function hasEncryptedAssertion(Response $sspResponse)
...
+        // If no key is in the response we don't need to validate if it's known because the
+        // validation if the signature is present is the responsibility of the SamlResponseValidator.
+        if (count($sspResponse->getCertificates()) == 0) {
+            return;
+        }
+
+        if (count($sspResponse->getCertificates()) > 1) {
+            $this->_logger->notice(
+                sprintf(
+                    'Tried to verify a message from issuer "%s", but we have unsupported multiple signing keys.',
```
johanib, line 911: Maybe make more specific?

`Tried to verify a message from issuer "%s", but requests with multiple signing keys are not supported.`

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -886,4 +891,48 @@ private function hasEncryptedAssertion(Response $sspResponse)
...
+            );
+
+            throw new EngineBlock_Corto_Module_Bindings_Exception(sprintf('Multiple signing keys supplied for issuer: %s', $metadata->entityId));
+        }
+
+        $validKey = false;
+        foreach ($metadata->certificates as $metaDataCertificate) {
+            if ($metaDataCertificate->toCertData() === $sspResponse->getCertificates()[0]) {
+                $validKey = true;
+                break;
```
johanib, line 923: Maybe `return` instead of break, so the $validKey variable can be removed?

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -886,4 +891,48 @@ private function hasEncryptedAssertion(Response $sspResponse)
...
+    /**
+     * This is an initial check to validate when a signing key is in the
+     * response if it's the one that's expected by EB for that IdP,
+     * so a descent error message can be shown if this is not the case.
+     */
+    private function assertKnownIdPSigningKey(IdentityProvider $metadata, Response $sspResponse): void {
+
+        // If no key is in the response we don't need to validate if it's known because the
+        // validation if the signature is present is the responsibility of the SamlResponseValidator.
+        if (count($sspResponse->getCertificates()) == 0) {
```
johanib, line 904:
```suggestion
        if (count($sspResponse->getCertificates()) === 0) {
```

theme/base/templates/modules/Authentication/View/Feedback/unknown-signing-key.html.twig
```diff
@@ -0,0 +1,12 @@
...
+
+{% set pageTitle = 'error_unknown_signing_key'|trans %}
+{% block pageTitle %}
+    {{ 'error_unknown_signing_key'|trans }}
+{% endblock %}
+{% block title %}{{ parent() }} - {{ pageTitle }} {% endblock %}
+{% block pageHeading %}{{ pageTitle }}{% endblock %}
+
+{% block errorMessage %}
+    {{ 'error_unknown_signing_key_desc'|trans }}
```
johanib, line 11: suiteName parameter needs to be added here?

theme/openconext/templates/modules/Authentication/View/Feedback/unknown-signing-key.html.twig
```diff
@@ -0,0 +1,12 @@
...
+
+{% set pageTitle = 'error_unknown_signing_key'|trans %}
+{% block pageTitle %}
+    {{ 'error_unknown_signing_key'|trans }}
+{% endblock %}
+{% block title %}{{ parent() }} - {{ pageTitle }} {% endblock %}
+{% block pageHeading %}{{ pageTitle }}{% endblock %}
+
+{% block errorMessage %}
+    {{ 'error_unknown_signing_key_desc'|trans }}
```
johanib, line 11: suiteName parameter needs to be added here?

languages/messages.pt.php
baszoetekouw: Let's put in the chatGPT version for now; I'll ask a Portugese developer to review:
```php
'error_unknown_signing_key' => 'Erro - chave de assinatura desconhecida',
'error_unknown_signing_key_desc' => 'A chave de assinatura utilizada não é conhecida por %suiteName%. Isto pode ser um erro de configuração.',
```

library/EngineBlock/Corto/Module/Bindings.php
baszoetekouw: Could we also dump the entire SAMLResponse here?  That would make it a lot easier to debug this, if it ever occurs.

library/EngineBlock/Corto/Module/Bindings.php
```diff
@@ -886,4 +891,48 @@ private function hasEncryptedAssertion(Response $sspResponse)
...
+            }
+        }
+
+        if(!$validKey) {
+            $this->_logger->notice(
+                sprintf(
+                    'Tried to verify a message from issuer "%s", but the expected signing key does not match.',
+                    $metadata->entityId
+                )
+            );
```
baszoetekouw, line 933: Could we be more verbose here?  For example:
```php
sprintf(
    'SAML Response from issuer "%s" signed with unknown key: "%s"',
    $metadata->entityId, $sspResponse->getCertificates()[0]
)
```

library/EngineBlock/Corto/Module/Bindings.php
baszoetekouw: Or does the exception handler do this already?

baszoetekouw (changes requested):
Could you please also add tests for this case?

languages/messages.pt.php
baszoetekouw, line 196:
```suggestion
    'error_unknown_signing_key' => 'Erro - chave de assinatura desconhecida',
```

theme/openconext/templates/modules/Authentication/View/Feedback/unknown-signing-key.html.twig
pablothedude: Those should be handled as default placeholders: 
`\OpenConext\EngineBlockBundle\Twig\Extensions\Extension\I18n::prepareDefaultPlaceholders`

pablothedude:
> Could you please also add tests for this case?

I've added a functional test and I shall also add a unit test.

library/EngineBlock/Corto/Module/Bindings.php
pablothedude: I think it is already logged here:

https://github.com/OpenConext/OpenConext-engineblock/blob/feature/add-expicit-signing-key-check/library/EngineBlock/Corto/Module/Bindings.php#L310

If that's not sufficient, could you elaborate on what else you would expect?

library/EngineBlock/Corto/Module/Bindings.php
pablothedude: They whole message is already logged with a `session_id` and `request_id` so I would expect there is already enough information.
https://github.com/OpenConext/OpenConext-engineblock/blob/feature/add-expicit-signing-key-check/library/EngineBlock/Corto/Module/Bindings.php#L310

languages/messages.pt.php
pablothedude: Changed as suggested and validated in Slack

johanib (approved):

library/EngineBlock/Corto/Module/Bindings.php
baszoetekouw: Looks good indeed!

baszoetekouw:
Apart from the missing test, this looks good to me!

baszoetekouw:
As discussed, testing this would mostly test mocks and wouldn't add much.  So the missing test is fine.

Issue #1777, opened 2024-12-11
Reduce backend calls (pdp, AA, ...) timeout 


baszoetekouw:
The current PDP timeout is rather long (30s), so the loadbalancer gives up sooner than EB.  Let's look in the logs what the median reponse time and determine a more reasionable timeout based on that.

Ideally, this timeout should be confgurable in the EB config.

thijskh:
Maybe make it more generic. Every backend API call from EB (to PDP, to AA) should not take more than $seconds where $seconds is most definitely < 10.

baszoetekouw:
Agreed.

PR #1797, opened 2025-01-29
Add configurable client timeout

pablothedude:
This is done for AA and PDP to be able to handle
timeouts before infrastructure limits are met.

https://github.com/OpenConext/OpenConext-engineblock/issues/1777


src/OpenConext/EngineBlockBundle/Resources/config/services.yml
```diff
@@ -202,6 +202,7 @@ services:
               # Verify CAs for certificates for prod, but not for other environments
               # as we are working with self signed signatures
               verify: "@=service('kernel').getEnvironment() === 'prod'"
+              timeout: "%http_client.timeout%"
```
johanib, line 205: I also expected the connect timeout. But the docs are not very clear what `timeout` entails. 

https://docs.guzzlephp.org/en/stable/request-options.html#connect-timeout
https://docs.guzzlephp.org/en/stable/request-options.html#read-timeout
https://docs.guzzlephp.org/en/stable/request-options.html#timeout

I'm not sure which are in scope for this ticket, so FYI.

pablothedude: According to the docs it should be the total timeout in the request. So that should be the time from the request till the response is returned.

PR #1819, opened 2025-03-03
Feature: Add SRAM integration

johanib:


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Discoveries.feature
pablothedude, line 1: Is this some boyscouting?

app/config/parameters.yml.dist
```diff
@@ -231,6 +231,7 @@ parameters:
     feature_enable_consent: true
     feature_stepup_sfo_override_engine_entityid: false
     feature_enable_idp_initiated_flow: true
+    feature_enable_sram_interrupt: true
```
pablothedude, line 234: Should this be the default?

library/EngineBlock/Application/DiContainer.php
```diff
@@ -306,6 +307,16 @@ protected function getSymfonyContainer()
         return $this->container;
     }
 
+    public function getSbsAttributeMerger(): SbsAttributeMerger
+    {
+        return $this->container->get('engineblack.sbs.attribute_merger');
+    }
+
+    public function getSbsClient(): \OpenConext\EngineBlockBundle\Sbs\SbsClientInterface
```
pablothedude, line 315: Can this FQDN be simplified?

library/EngineBlock/Application/TestDiContainer.php
```diff
@@ -49,11 +61,31 @@ public function getPdpClient()
...
+    {
+        $this->sbsClient = $sbsClient;
+    }
+
+    public function getSbsClient(): SbsClientInterface
+    {
+        return $this->sbsClient ?? parent::getSbsClient();
+    }
+
+    public function setFeatureConfiguration(?FeatureConfigurationInterface $featureConfiguration)
```
pablothedude, line 79: Why is this needed here, why can't the methods in the EngineBlockContext be used?

- featureIsEnabled
- featureIsDisabled

library/EngineBlock/Corto/Filter/Input.php
```diff
@@ -90,6 +90,8 @@ public function getCommands()
                 $diContainer->getAttributeAggregationClient()
             ),
 
+            new EngineBlock_Corto_Filter_Command_SRAMTestFilter(),
```
pablothedude, line 93: Can you add a docblock above to describe what this exactly does.

library/EngineBlock/Corto/Module/Service/SRAMInterrupt.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * route that receives the user when they get back from their SBS interrupt,
+     * fetches the attribues from SBS,
+     * and resumes the AuthN flow.
+     *
+     * @param $serviceName
+     * @param Request $httpRequest
+     */
+    public function serve($serviceName, Request $httpRequest)
```
pablothedude, line 78: It seems that this method is handling more than the SBS interrupt callout. Is that really needed in order to let this function properly?

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/SbsFlowIntegration.feature
```diff
@@ -0,0 +1,93 @@
+Feature:
```
pablothedude, line 1: :100: nice!

pablothedude (changes requested):
This is an initial review mainly focusing on the code. This feature however should be thoroughly reviewed functionally before it's merged back and after all open todo's are addressed.

mrvanes:
Ik ben aan het testen, maar loop tegen deze conditie aan in SRAMTestFilter:
```
if ($this->_serviceProvider->getCoins()->collabEnabled() === false) {
    return;
}
```
Hoe krijg ik die in manage, onder welke versie van manage (ik draai nu 9.0.0.1)?
Mist er misschien een row in de database voor collab_enabled?

mrvanes:
Manage collabEnabled() problem has been solved. However, EB now gives me a "Session Lost" on returning to https://engine.dev.openconext.local/authentication/idp/process-sraminterrupt?ID=...
I see you have tested with sram.base_url served by EB itself so you may have lifted on more session information than you can count on in real life?

I suggest you use the SBS stub server that was originally included in my branch to test your redirects:
https://github.com/OpenConext/OpenConext-engineblock/tree/feature/sram-interrupt/sbs-stub

johanib:
> Manage collabEnabled() problem has been solved. However, EB now gives me a "Session Lost" on returning to https://engine.dev.openconext.local/authentication/idp/process-sraminterrupt?ID=... I see you have tested with sram.base_url served by EB itself so you may have lifted on more session information than you can count on in real life?
> 
> I suggest you use the SBS stub server that was originally included in my branch to test your redirects: https://github.com/OpenConext/OpenConext-engineblock/tree/feature/sram-interrupt/sbs-stub

The mock SBS does/should not need or rely on any session state. It receives the `continue_url` on the authz call, like in the example python script, and redirects the browser to that url when the interrupt is performed.
As it's only there for facilitating the functional tests and development, it does not support concurrent requests etc. It just stores the last received authz payload on filesystem.

When testing, does the `ID=...`  contain an actual value?
If so, it seems like that value cannot be matched to a valid message request ID in the browser session.

mrvanes:
Yes, the ?ID= parameter contains everything in the original continue_url. EngineBlock fails to continue on the contents of this continue_url. That's the bug. I would advice against testing this against an internal EB redirect but externalising like this will be done in the real deploy.

library/EngineBlock/Application/TestDiContainer.php
johanib: This is for the corto tests.
I think the EngineBlockContext is only available for the behat tests?

mrvanes (approved):

PR #1824, opened 2025-03-25
Feature/sf 34 fix deprecations

Jong-Vincent:


tests/library/EngineBlock/Test/Corto/ProxyServerTest.php
```diff
@@ -44,15 +45,7 @@ public function testNameIDFormatIsNotSetByDefault()
...
-
-        $this->assertNotContains(
-            'Format',
-            array_keys($nameIdPolicy),
-            'The NameIDPolicy should not contain the key "Format"',
-            false,
-            true,
-            true
-        );
+        $this->assertSame(['AllowCreate' => true], $nameIdPolicy);
```
tvdijen, line 48: This is not testing what it used to test.

tvdijen (commented):
There is also one more deprecation notice in the logs that should be resolved as part of this PR:

> Deprecation Notice: Not quoting the scalar "%kernel.project_dir%/application/configs/attributes.json" starting with the "%" indicator character is deprecated since Symfony 3.1 and will throw a ParseException in 4.0 on line 39. in /var/www/html/vendor/symfony/symfony/src/Symfony/Component/Yaml/Inline.php:359

Issue #1849, opened 2025-08-11
Engine ARP must not apply to User Atrributes


pmeulen:
The user_attributes sent by engine currently takes the ARP into account. This means that now, when an attribute is not in the ARP, it is not sent in the Stepup callout even if it's present in the attributes that are received from the IdP.

It must be possible to send any attribute that was received from the IdP as a user attrubute in the Stepup callout. Since these attributes do not go to the SP, the ARP does not apply.

Ideally any attributes that were retrieved via attribute aggregation can be sent as a user attribute in the Stepup callout. This would broaden the use case (e.g. we could use an attribute aggregator to lookup the user's token ID). However in the current envisioned use-case this is not necessary, so I consider this a nice to have.

PR #1850, opened 2025-08-11
Prevent Stepup user attributes filtered by ARP

pablothedude:
https://github.com/OpenConext/OpenConext-engineblock/issues/1849


library/EngineBlock/Corto/Module/Service/AssertionConsumer.php
```diff
@@ -168,6 +168,7 @@ public function serve($serviceName, Request $httpRequest)
         // or another service.
         EngineBlock_Corto_Model_Response_Cache::rememberIdp($receivedRequest, $receivedResponse);
 
+        $originalAssertions = clone $receivedResponse->getAssertions()[0];
```
johanib, line 171: clone vs serialize/deserialize: Is it correct to make a shallow copy in this scenario?

PR #1860, opened 2025-09-15
Make metadata expiration time configurable

AramMutluWork:
The expiration time of the metadata defines the validUntil until attribute in the metadata of Engineblock. By default this a constant in the code set to 86.400 seconds (24 hours).

At the request of one of the connected Service Providers, we want to increase this value to 1,814,400 seconds (21 days). If something goes wrong with the automatic reading of our metadata, they will have more time to fix this error before users experience login problems.

To avoid having to set this up again with every release, we want to make this configurable, I left the default at 24 hours as before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Maintenance:
 * `symfony/monolog-bundle` upgraded to ^4.0; review your monolog configuration if you have customised it outside of the defaults.
 
 Changes:
+* The metadata expiration time (`validUntil` attribute) is now configurable via the `metadata_expiration_time` parameter in `parameters.yml`.
+  * Action required: Add `metadata_expiration_time` to `parameters.yaml`, suggested value: `86400`. This is the old behaviour in which metadata is cached for 24 hours.
 * The `consent.deleted_at` should be not nullable, and have a default value of `0000-00-00 00:00:00`.
   * Because `deleted_at` is part of the PK, no migration is provided. The database engine should not allow this to be null in the first place, so it is probably not nullable already on your db.
   * The `0000-00-00 00:00:00` is added for clarity/consistency, as this is probably the default behaviour of your database already.
--- a/tests/unit/OpenConext/EngineBlock/Xml/MetadataRendererTest.php
+++ b/tests/unit/OpenConext/EngineBlock/Xml/MetadataRendererTest.php
@@ -20,6 +20,7 @@ namespace OpenConext\EngineBlock\Xml;
 use DOMDocument;
 use EngineBlock_Saml2_IdGenerator;
 use Exception;
+use InvalidArgumentException;
 use Mockery as m;
 use Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;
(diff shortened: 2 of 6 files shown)


MKodde (commented):
A nice and useful new contribution! 

I see a lot of nice things, but looking at this critically, I have some suggestion/questions:

1. Now that the exp time is no longer hard coded; would we not need some validation/verification if a correct value is configured? What happens with negative values, floating points, string values,... This validation does not need to be very comprehensive, and by looking at your type definition, the application only allows for integer values. But that can still be a negative value. Maybe check for that in the MetadataRenderer constructor?
2. I'd love to see the MetaDataRendererTest be extended with a couple of additional tests to verify that the expected expiration time matches with the one configured

tvdijen:
I think allowing a little clock-skew would be appropriate here. SAML2INT allows a clock-skew between 3-5 minutes for _SAML messages_. For SAML metadata (SAML2MDIOP) there is no such requirement, but I think it makes sense to allow _some_ skew.  I'm keen to know how SURF looks at this.

Wasn't there a setting for clock-skew already? Or am I mixing up with the SimpleSAMLphp-project that has this setting?

tests/unit/OpenConext/EngineBlock/Xml/MetadataRendererTest.php
```diff
@@ -295,8 +296,48 @@ public function metadata_add_no_requested_attributes()
...
+    }
+
+    #[Group('Metadata')]
+    #[Test]
+    public function the_configured_metadata_expiration_time_is_reflected_in_the_valid_until_attribute()
+    {
+        $fixedTime = mktime(0, 0, 0, 1, 1, 2026);
+        $expirationTime = 3600;
+
+        $timeProvider = $this->createMock(TimeProvider::class);
```
kayjoosten: Nitpick: for consistency i would recommend also use mockery like the other test

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Maintenance:
 * `symfony/monolog-bundle` upgraded to ^4.0; review your monolog configuration if you have customised it outside of the defaults.
 
 Changes:
+* The metadata expiration time (`validUntil` attribute) is now configurable via the `metadata_expiration_time` parameter in `parameters.yml` (default: 86400 seconds / 24 hours).
```
kayjoosten: I would also suggest to add that the parameter is required to be set. Maybe UPGRADING.md is a good place to do this.

kayjoosten (changes requested):

johanib:
> I think allowing a little clock-skew would be appropriate here. SAML2INT allows a clock-skew between 3-5 minutes for _SAML messages_. For SAML metadata (SAML2MDIOP) there is no such requirement, but I think it makes sense to allow _some_ skew. I'm keen to know how SURF looks at this.
> 
> Wasn't there a setting for clock-skew already? Or am I mixing up with the SimpleSAMLphp-project that has this setting?

I can't find a clock skew setting / feature in EB. 
Since this PR is straight forward, I'll create a new ticket to track metadata clock skew seperatly.

kayjoosten (approved):

PR #1868, opened 2025-09-30
Upgrade sass, fontawesome, remove broken minifier and fix js build

johanib:
Two commits in this PR, advisable to review both commits separate.

---

Upgrade npm packages, `sass`, fontawesome

Javascript package maintenance.
Updated some packages.

Most notable, are sass and fontawesome.

Updated sass from 1.66.1 > 1.93.2.
This is the biggest change, specifically:
`@import and global built-in functions (i.e. un-namespaced builtins) are deprecated, and Sass is moving toward requiring @use / @forward and module-based builtins`.

---

Fix/rework building of javascript

Prior to this change, in the previous commit, the `tinyify` package was
upgraded. This package minified the js and css.
However, after the update, the javascript was no longer functioning.
This manifested itself by errors in the browser console.

This had two root causes:
* The javascript in EB used circular dependencies.
* Even after fixing that, the tinyify package simply broke the
   javascript by removing functions that could not be removed.

This change reworks the javascript to remove the circular dependencies.
It also replaces the unmaintained tinyify plugin by using the `terser` directly.


MKodde (approved):
A bit much to take in. I approve this as I do not see any red flags. I trust you that the changes result in a sane build of the EB themes.

Maybe It is useful for other theme builders to post some highlights of what to be aware of when they are patching their own theme?

Issue #1861, opened 2025-09-16
Ensure consistent database schema for migrations


johanib:
**Current situation**
Currently, in devconf (`core/scripts/init.sh`) and in the test scripts, the database is created using `doctrine:schema:create`.
<img width="647" height="111" alt="Image" src="https://github.com/user-attachments/assets/098b8b7e-c880-4c96-a148-dd5346a934d4" />

Also, there are migrations. 

The problem is that the schema created by `doctrine:schema:create`, changes over time, but the migrations assume a steady base.
So currently, they are not compatible with each other.

When you first run `doctrine:schema:create`, and then run the migrations, the following error occurs:
`SQLSTATE[42S02]: Base table or view not found: 1146 Table 'eb_test.sso_provider_roles' doesn't exist`

What this means is: 
 * In development/pipeline, the tests are not executed on the same database schema as used in production
 * Development is not done on the same schema as used in production.
 * When a developer creates a migration, it is probably based on a schema that is currently not in production!

**Proposed solution:**
Never use `doctrine:schema:create`, always execute the migrations.

**Changes needed**
Ensure the `doctrine:migrations:migrate` command can be executed on an empty database.

This can be done by:
* Adding a migration at the beginning of the migration that sets the base state.
* OR
* Squash current migrations: Take the current production database schema (ensure all migrations are executed, including `Version20230824090020`), delete all old migrations, and create a new migration that creates the schema based on that export.

In both scenario's, the migration should not be executed on production environments, and be set to executed manually.

**Related**
This is not applied on production? https://github.com/OpenConext/OpenConext-engineblock/pull/1262

PR #1872, opened 2025-10-07
Upgrade to doctrine/dbal:4

johanib:
Needed for SF 6.4


src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
```diff
@@ -48,19 +48,19 @@ class ServiceProvider extends AbstractRole
...
      */
-    #[ORM\Column(name: 'assertion_consumer_services', type: \Doctrine\DBAL\Types\Types::ARRAY, length: 65535)]
+    #[ORM\Column(name: 'assertion_consumer_services', type: 'engineblock_indexed_service_array', nullable: true)]
     public $assertionConsumerServices;
 
     /**
      * @var string[]
      */
-    #[ORM\Column(name: 'allowed_idp_entity_ids', type: \Doctrine\DBAL\Types\Types::ARRAY, length: 6777215)]
+    #[ORM\Column(name: 'allowed_idp_entity_ids', type: 'json', length: 16777215, nullable: true)]
```
MKodde: Was the (significant) length increase intentional?

I also see that the field became nullable, but that is now explicitly set for all of the updated columns. I'm assuming that was previously the default setting which now needs an explicit setting.

MKodde (commented):
Nice work! We had a small discussion IRL where I questioned the current solution:

1. Could you make a more 'generic' solution for the object and array types that you now added. This would need to include passing the type of the array items that are stored for additional type checking. And maybe passing a max length as some types have different requirements for that..
2. Keep the current setup but add some additional (unit or integration) tests. I've found a couple of unit tests [for two of the existing types](https://github.com/OpenConext/OpenConext-engineblock/blob/99fa3dd3a033250aab53c2a9c0818028de8d9d9f/tests/unit/OpenConext/EngineBlockBundle/Doctrine/Type/CollabPersonUuidTypeTest.php)

Question: are the types used in the existing functional tests? Or are they bypassed by the custom mock solution?

src/OpenConext/EngineBlock/Metadata/Entity/ServiceProvider.php
johanib: @MKodde I checked the length with: https://github.com/OpenConext/OpenConext-engineblock/issues/1861#issuecomment-3479676422

Its currently mediumtext on prod, so it should be 16777215. I htink 6777215 was a typo.

kayjoosten (changes requested):
Test are still failing

johanib:
> Nice work! We had a small discussion IRL where I questioned the current solution:
> 
> 1. Could you make a more 'generic' solution for the object and array types that you now added. This would need to include passing the type of the array items that are stored for additional type checking. And maybe passing a max length as some types have different requirements for that..
> 2. Keep the current setup but add some additional (unit or integration) tests. I've found a couple of unit tests [for two of the existing types](https://github.com/OpenConext/OpenConext-engineblock/blob/99fa3dd3a033250aab53c2a9c0818028de8d9d9f/tests/unit/OpenConext/EngineBlockBundle/Doctrine/Type/CollabPersonUuidTypeTest.php)
> 
> Question: are the types used in the existing functional tests? Or are they bypassed by the custom mock solution?

1. Yes, this was the correct way to go from the start. Implemented now. 
2. Also added tests for the two 'new' types. 
3. The behat tests fail if I omit the serialize, so yes, they are used in the behat tests.

kayjoosten (approved):

src/OpenConext/EngineBlock/Metadata/Entity/AbstractRole.php
```diff
@@ -134,25 +136,25 @@ abstract class AbstractRole
...
      * @deprecated Will be removed in favour of using the Mdui value object, use the getter for this field instead
      */
-    #[ORM\Column(name: 'logo', type: \Doctrine\DBAL\Types\Types::OBJECT)]
+    #[ORM\Column(name: 'logo', type: SerializedObjectType::NAME, nullable: false)]
     public $logo;
 
     /**
      * @var Organization
      */
-    #[ORM\Column(name: 'organization_nl_name', type: \Doctrine\DBAL\Types\Types::OBJECT, nullable: true, length: 65535)]
```
MKodde: Let me start by complimenting you on this change. This makes the PR so much more to the point!

Two things caught my eye: 
- Is there no need to set the column length anymore? Did we set the default value excessively before?
- And, I was also triggered by you setting the `nullable` value for each field now. I'm OK with that, before we would only set it if the non-default value was required. 
- Does this change cause 0 schema changes? If it does, you might need to think about a strategy to roll out that change safely. But I guess there are none judging the code.

MKodde (approved):
Nice work! Thanks for taking extra care in checking the schema changes, `required` and `length` constraints.

PR #1874, opened 2025-10-08
Feat: Allow Dependency Injection from Symfony into Corto

johanib:
Prior to this change, it was not possible to inject dependencies into Corto in a modern way. Corto pulled all dependencies from the Symfony service container itself.
This caused issues, because it is no longer possible to pull twig from the service container in Symfony 6.4.

This change introduces a DiContainerRuntime, which can be used to inject dependencies into Corto. This was not possible using the DiContainer, because the DiContainer is constructed during bundle bootstrapping, where DI is not available. It is not possible to move it from there, as it is required by various constructions in the bootstapping process.
The DiContainerRuntime is injected into Corto after bootstrapping, but before the request is handled.

https://github.com/OpenConext/OpenConext-engineblock/pull/1874

### DIFF
--- a/config/services/services.yml
+++ b/config/services/services.yml
@@ -2,6 +2,12 @@ services:
     _defaults:
         public: true
 
+    OpenConext\EngineBlockBundle\Bridge\EngineBlockBootstrapper:
+        autowire: true
+        autoconfigure: true
+        tags:
+            - { name: kernel.event_subscriber }
+
     OpenConext\EngineBlock\Xml\:
         resource: '../../src/OpenConext/EngineBlock/Xml'
         exclude: '../../src/OpenConext/EngineBlock/Xml/ValueObjects'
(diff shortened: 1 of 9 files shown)


MKodde (commented):
IRL @johanib convinced me to not big bang move all services to the new Runtime version. I then suggested to add some nuance to the instructions found in the PHPDoc

johanib:
> IRL @johanib convinced me to not big bang move all services to the new Runtime version. I then suggested to add some nuance to the instructions found in the PHPDoc

After discussion:
* The thing is, the new situation is also not ideal. As it is not possible to directly inject dependencies into the classes that need them. 
* The old way cannot be removed 100%, because that would probably be a very big refactor, as the bootstrapping would be changed. But it would be possible to move 90% of the services to the new container.

I agree it would be nice to move most services over.

johanib:
> IRL @johanib convinced me to not big bang move all services to the new Runtime version. I then suggested to add some nuance to the instructions found in the PHPDoc

Done.

PR #1888, opened 2025-11-12
Update composer packages

johanib:


MKodde (approved):
Looks good!

I took a look at the cypress test results. And I see a couple are marked 'pending' (synonym for skipped test IIRC). Is that intentional, or something that was done temporarily?

PR #1893, opened 2025-11-29
add use IDP LOA (authnContextClassReff) in Stepup Decision

### LINKED ISSUE #1883: received LOA is not respected when making stepup disicion
Scenario:

An SP asks for a higher LOA using requered.
IDP1 is configured for transparent LOA and handles the MFA request itself before responding.
IDP2 uses Stepup for it's MFA needs.

SP sends a LOA defined for stepup in the RequestedAuthnContext element.

A user from IDP2 can now select his/her IDP using the WAYF and login.
Afterwards stepup is correctly triggerd to provide MFA before a response to the SP is send.

A user from IDP1 however chooses his/her IPD using the WAYF.
LOA is send to the IDP using the transparant LOA config (so far so good).
At IDP1 the users provides credentials and MFA token.
The response from IDP1 sends a reponse with the higher LOA but this higher LOA is ignored.
Instead the user is then send to Stepup for an aditional token which they don't have and as a result can't log in.

It seems that the LOA from the response is not used in the decision if stepup is required in:
https://github.com/OpenConext/OpenConext-engineblock/blob/main/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php#L192

This seems against SAML2.0 specification.
Can the LOA from the IDP be added to the decision process to facilitate this usecase?

ArnoutvdKnaap:
adds a feature flag to enable engineblock to use the AuthnContextClassReff in the response from the IDP in the stepupDecision.
This allows for IDPs to use their own MFA in combination with stepup.

### DIFF
--- a/config/packages/engineblock_features.yaml
+++ b/config/packages/engineblock_features.yaml
@@ -15,3 +15,4 @@ parameters:
         eb.feature_enable_idp_initiated_flow: "%feature_enable_idp_initiated_flow%"
         eb.stepup.sfo.override_engine_entityid: "%feature_stepup_sfo_override_engine_entityid%"
         eb.stepup.send_user_attributes: "%feature_stepup_send_user_attributes%"
+        eb.stepup_use_idp_response_loa: "%feature_use_idp_loa_in_stepup_decision%"
--- a/config/packages/engineblock_features.yaml
+++ b/config/packages/engineblock_features.yaml
@@ -15,4 +15,4 @@ parameters:
         eb.feature_enable_idp_initiated_flow: "%feature_enable_idp_initiated_flow%"
         eb.stepup.sfo.override_engine_entityid: "%feature_stepup_sfo_override_engine_entityid%"
         eb.stepup.send_user_attributes: "%feature_stepup_send_user_attributes%"
-        eb.stepup_use_idp_response_loa: "%feature_use_idp_loa_in_stepup_decision%"
+        eb.stepup.use_idp_response_loa: "%feature_use_idp_loa_in_stepup_decision%"
--- a/src/OpenConext/EngineBlock/Stepup/StepupDecision.php
+++ b/src/OpenConext/EngineBlock/Stepup/StepupDecision.php
@@ -44,13 +44,12 @@ class StepupDecision
      */
     private $pdpLoas = [];
     /**
-     * @var bool
-     */
-    /**
      * @var string|null
      */
     private $idpResponseLoa = null;
-
+    /**
+     * @var bool
+     */
     private $spNoToken;
 
     private $logger;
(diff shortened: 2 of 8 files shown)


config/packages/engineblock_features.yaml
```diff
@@ -15,3 +15,4 @@ parameters:
         eb.feature_enable_idp_initiated_flow: "%feature_enable_idp_initiated_flow%"
         eb.stepup.sfo.override_engine_entityid: "%feature_stepup_sfo_override_engine_entityid%"
         eb.stepup.send_user_attributes: "%feature_stepup_send_user_attributes%"
+        eb.stepup_use_idp_response_loa: "%feature_use_idp_loa_in_stepup_decision%"
```
tvdijen: For consistency, the first underscore should be a dot.

ArnoutvdKnaap:
the commit history is not very nice any more but all tests are passing now

src/OpenConext/EngineBlock/Stepup/StepupDecision.php
```diff
@@ -46,6 +46,11 @@ class StepupDecision
     /**
      * @var bool
      */
+    /**
```
tvdijen: The `@var bool` above this line belongs to the property below this change.

tests/unit/OpenConext/EngineBlock/Stepup/StepupDecisionTest.php
```diff
@@ -160,4 +160,103 @@ private static function buildLoa(string $loaLevel): Loa
...
+        $logger = m::mock(LoggerInterface::class)->shouldIgnoreMissing();
+
+        $stepupDecision = new StepupDecision($idp, $sp, [], [], 'loa20', $repo, $logger);
+
+        $this->assertTrue($stepupDecision->shouldUseStepup());
+        $this->assertEquals('loa30', $stepupDecision->getStepupLoa()->getIdentifier());
+    }
+
+    #[\PHPUnit\Framework\Attributes\Test]
+    #[\PHPUnit\Framework\Attributes\Group('Stepup')]
```
tvdijen: I would personally add a few use-statements on the top of this file for these attributes and the exception below to keep the code better readable.

ArnoutvdKnaap: I only added 1 for the exception. That is more in line with the other tests.

MKodde (commented):
Sorry to be so late to the party, but I wanted to give some additional feedback and ask a fundamental question.

Other than that your change is looking nice and neat, I see we merged some other features that mix with your changes, so a rebase is required before merging. I would also love to see a Behat test covering this feature. Should be doable..

PR #1899, opened 2025-12-16
Add rollover metadata containing all available keys

### LINKED ISSUE #1223: /metadata endpoint should expose all the available keys
The IDP- and SP-metadata endpoints should add KeyDescriptor-elements for all (or all non-expired) available keys to accommodate for a key-rollover.
This has been discussed multiple times, but every time I have to do key rollover I run into this issue, esp. because we sign our AuthnRequests towards IDPs.
See this old thread; https://groups.google.com/g/openconext/c/1oexBnx5KHA/m/Z9V9vYWIBwAJ

tvdijen:
We sign our AuthnRequests to our IdPs, so we need to have metadata containing both the current and the rollover-certificate.
With this patch, the default metadata (keyslug=default or no keyslug) will contain all the configured keys.

### DIFF
--- a/config/packages/ci/parameters.yml
+++ b/config/packages/ci/parameters.yml
@@ -1,0 +1,8 @@
+parameters:
+    encryption_keys:
+        default:
+            publicFile: /config/engine/engineblock.crt
+            privateFile: /config/engine/engineblock.pem
+        rollover:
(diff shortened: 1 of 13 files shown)


johanib (changes requested):

johanib (approved):

baszoetekouw:
tested.

Issue #1902, opened 2026-01-20
Health endpoint doesn't work


baszoetekouw:
In Engineblock 7.0.0-beta, the `/health` and `/info` endpoints return a 404.  These should be provided by the OpenConext-monitor bundle, but apparently the routes are not found.

When starting EB in devconf, I get a 404 instead.  Same happens for `/info`, `/internal/health` and `/internal/info`.
```
╰─▶ docker exec -ti core-engine-1 curl -v http://localhost/health
*   Trying ::1:80...
* connect to ::1 port 80 failed: Connection refused
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> GET /health HTTP/1.1
> Host: localhost
> User-Agent: curl/7.74.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Date: Mon, 19 Jan 2026 17:02:30 GMT
< Server: Apache
< X-Content-Type-Options: nosniff
< Cache-Control: max-age=0, must-revalidate, private
< Expires: Mon, 19 Jan 2026 17:02:30 GMT
< Set-Cookie: lang=en; expires=Fri, 20-Mar-2026 17:02:30 GMT; Max-Age=5184000; path=/; domain=.dev.openconext.local; secure; samesite=lax
< Transfer-Encoding: chunked
< Content-Type: text/html; charset=UTF-8
[...]
```

Could you take a look how to reenable these endpoints?  This is a blocker for EB 7.0.

Please make the endppoint available at `/internal/health` and `/internal/info`.  The old (top-level) paths are deprecated.

Issue #1502, opened 2024-11-28
monitor-bundle: change SELECT 1 to something more specific


phavekes:
[This issue is imported from pivotal - Originaly created at Jul 27, 2018 by Thijs Kinkhorst](https://www.pivotaltracker.com/story/show/159361260)

The monitor-bundle uses a query "SELECT 1" to determine database upness. This catches a number of scenarios, but not for example: access rights problems, or locking issues.

It would be better to make the query select something actually in the database (but preferably not something expensive). Maybe `select count(*) from migration_versions`?

phavekes:
@bartgeesink Updating the actual query to something more specific. Like `SELECT COUNT(*) FROM migration_versions;`. A great suggestion, the only problem is that the bundle should test if doctrine migrations are in use and then query the migrations table. 

Another solution might be to use the `SELECT table_name FROM information_schema.tables where table_schema=\'[current-shema-name]\';
`; output and query one of its tables. This might be problematic in certain situations. (would be incompatible with other databases)

A third solution might be to configure a query in the monitor bundle config that would be used in the `DoctrineConnectionHealthCheck`

All solutions have their pros and cons. I\'d like your feedback and ideas in this @bartgeesink  and @bstrooband  (Michiel Kodde - Jul 31, 2018)

phavekes:
I don\'t like the second solution: it\'s not much better than the current one since it does not detect EB db access problems or locking issues.

I would be fine with the third solution. It allows to specify something that makes most sense to check for the specific component. (Thijs Kinkhorst - Aug 13, 2018)

phavekes:
https://github.com/OpenConext/Monitor-bundle/pull/2 In addition to the changes made in this PR, another change is coming up, addressing the faulty default SELECT 1; query that is in use now. (Michiel Kodde - Aug 14, 2018)

phavekes:
I\'m now integrating the bundle change into EB. I\'m opting for this feature always being enabled for the EB users as of the upcoming release. You could opt out of the feature by ~~leaving the query empty~~ removing the service config override, which results in falling back to default behaviour in the DoctrineConnectionHealthChecker (monitor-bundle). @thijskh is this what you had in mind?

https://github.com/OpenConext/OpenConext-engineblock/pull/589 (Michiel Kodde - Aug 15, 2018)

PR #1903, opened 2026-01-20
Reinstate monitor endpoints

johanib:
Prior to this change, the `/internal/health` and `/internal/info` routes did not exist.

This change re-introduces these routes and ensures they work.

In order to make this happen, the  `AuthenticationStateInitializer` listener needed to be made compatible with invokable controllers, which the monitor bundle uses.

Fixes: https://github.com/OpenConext/OpenConext-engineblock/issues/1902

### DIFF
--- a/config/routes.yaml
+++ b/config/routes.yaml
@@ -15,3 +15,7 @@ engineblock_api_controllers:
         domain: ".*"
     defaults:
         domain: "%domain%"
+
+open_conext_monitor:
+    resource: "@OpenConextMonitorBundle/Resources/config/routing.yml"
+    prefix: /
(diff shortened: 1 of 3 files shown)


MKodde (commented):
Looks good. While testing the endpoints some questions raised;

1. The `/internal/health` endpoint does not return an `UP` status, but yields `DOWN` with message: `Unable to execute a query on the database.` This indicates it tries to test if the database is up, but does not know what query to perform to figure that out. The Monitor Bundle docs should help you in that area. And [this health checker](https://github.com/OpenConext/OpenConext-engineblock/blob/99fa3dd3a033250aab53c2a9c0818028de8d9d9f/src/OpenConext/EngineBlockBundle/HealthCheck/DoctrineConnectionHealthCheck.php) is now causing the issue. Did you get the same issue? Maybe it's a dev-box thing that only I run into?
2. Given `/health` and `/info` are deprecated. Would it be OK to block those routes? Or maybe start thinking about a new MonitorBundle release without the deprecated routes? Food for thought.

Monitor Bundle readme: https://github.com/OpenConext/Monitor-bundle/blob/main/README.md
See: https://github.com/OpenConext/OpenConext-engineblock/commit/a3e134f2f954050f35cc52d0d4221c69391b6629
Also: https://github.com/OpenConext/OpenConext-engineblock/issues/1502
And: https://github.com/OpenConext/OpenConext-engineblock/pull/589

Issue #1859, opened 2025-09-10
Check PDO prepared statements


baszoetekouw:
It turns out that PDO prepared statements are (by default) not actually prepared statements, but are rather parsed by some internal PDO prepared statement parser.  And of course, this PDO parser does not implement a full SQL parser, which opens up PDO prepared statements too all kinds of injection attacks, like for example described here: https://slcyber.io/assetnote-security-research-center/a-novel-technique-for-sql-injection-in-pdos-prepared-statements/

Fortunately, there is a simple fix: set `PDO::ATTR_EMULATE_PREPARES` to `false`, which disables this ridiculous behaviour and reverts back to the behaviour you'd expect: let the database engine handle prepared statements.

Please check if we indeed disable this setting in EB, and if not, add the setting.  I expect us not to depend on weird PDO behaviour, so I would hope that changing this setting won't have any impact (famous last words...).

Note that the PDP docs say this:
> Whether enable or disable emulation of prepared statements. Some drivers do not support prepared statements natively or have limited support for them. If set to [true](https://www.php.net/manual/en/reserved.constants.php#constant.true) PDO will always emulate prepared statements, otherwise PDO will attempt to use native prepared statements. In case the driver cannot successfully prepare the current query, PDO will always fall back to emulating the prepared statement.

(https://www.php.net/manual/en/pdo.setattribute.php)
The latter part of this worries me.  Does PHP emit a warning when it falls back to emulating the prepared statement?  What exactly does it mean that the MySQL driver has "limited support" for prepared statements?  What kind of statements would fail preparing by the driver?

tvdijen:
Interesting read on the impact of setting flag to `false` i.r.t. Doctrine:
https://stackoverflow.com/questions/56853024/prevent-php-pdo-and-mysql-mariadb-from-returning-integers-as-strings-without-d

PR #1919, opened 2026-02-27
fix: disable PDO emulated prepared statements (issue #1859)

kayjoosten:
Set PDO::ATTR_EMULATE_PREPARES to false on all DBAL connections.
 By default PDO uses its own C-level parser to interpolate parameters,
 which is incomplete and vulnerable to SQL injection via multi-byte
 charset encoding tricks. Setting this to false forces native prepared
 statements on the MariaDB server side, where parameter binding is
 handled correctly.

issue #1859


johanib (commented):
Though PR...

Found this interesting: https://phpdelusions.net/pdo#emulation

I think these changes are 👍  and should be good. 
Also tested the disabling of this feature in a random project and nothing broke. So only if you have custom queries with named params, something can break. 

One thing to note, which might be relevant to @baszoetekouw : 
Before this change, PDO would send the query to the database in one go. 
After this change, it is potentially sent to the db in two parts: First the query, then the parameters, because we use db native prepared statements. 

This is not an issue when the latency between webserver and db server is very low (<1ms). But if there are scenario's where the db is remote, this can cause delays, as each query can potentially need two instead of one request to the database.
This cost applies to all queries using parameters, because doctrine basically does this:
```php
if (count($params) > 0) {
    $stmt = $connection->prepare($sql);   // ← COM_STMT_PREPARE
    $this->bindParameters($stmt, $params, $types);
    $result = $stmt->execute();           // ← COM_STMT_EXECUTE
} else {
    $result = $connection->query($sql);   // ← single round trip, no params
}
```
mysqlnd does cache the prepared statements on a connection/process level, so the actual performance impact might be minimal.

PR #1920, opened 2026-02-27
Fix: add time to error screen.

### LINKED ISSUE #1805: Show current date and time in user error screens
In error screen like this: 

Image

please add the datetime to the data fields on the bottom.

Currently, it's hard to find a specific issue if we don't know when the user encountered the error.

kayjoosten:
Its hard to find errors in the logs without a timestamp as we dont know when the error occured. This is something we want to help the support department. Therefore, the time the screen appeard should have a iso timestamp.

page-2026-03-06T10-07-33-508Z

### DIFF
--- a/library/EngineBlock/ApplicationSingleton.php
+++ b/library/EngineBlock/ApplicationSingleton.php
@@ -257,7 +257,11 @@ class EngineBlock_ApplicationSingleton
             $logRequestId = $logRequestId->get();
         }
 
+        $request = $this->getDiContainer()->getSymfonyRequest();
+
         $feedbackInfo = array();
+        $feedbackInfo['datetime'] = (new DateTime())->format(DateTimeInterface::ATOM);
+        $feedbackInfo['requestUrl'] = sprintf('%s%s', $request->getSchemeAndHttpHost(), $request->getPathInfo());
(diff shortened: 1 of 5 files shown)


library/EngineBlock/ApplicationSingleton.php
```diff
@@ -243,6 +243,12 @@ public function collectFeedbackInfo(Throwable $exception)
         }
 
         $feedbackInfo = array();
+        $feedbackInfo['datetime'] = (new DateTime())->format(DateTimeInterface::ATOM);
+        $requestUri = $_SERVER['REQUEST_URI'] ?? '';
+        $urlWithoutQuery = strtok($requestUri, '?');
+        $scheme = (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off') ? 'https' : 'http';
+        $host = $_SERVER['HTTP_HOST'] ?? '';
+        $feedbackInfo['requestUrl'] = $host ? $scheme . '://' . $host . $urlWithoutQuery : $urlWithoutQuery;
```
johanib: Maybe better use `$request->getSchemeAndHttpHost()` as it takes the trusted_hosts config into consideration

johanib (approved):
Image

Image

PR #1929, opened 2026-03-06
refactor: convert ConsentType to a PHP 8.1 backed enum

kayjoosten:
Replace the ConsentType value class (with its deprecated public constructor since ~2015) with a proper backed enum.

- ConsentType is now `enum ConsentType: string` with cases Explicit = 'explicit' and Implicit = 'implicit'
- The deprecated public constructor, the explicit()/implicit() named constructors and the equals() method are removed
- DbalConsentRepository uses ConsentType::from() to hydrate from DB
- Legacy Consent model methods now accept ConsentType and pass ->value into SQL parameter arrays
- All call sites updated from TYPE_EXPLICIT/TYPE_IMPLICIT constants and named constructors to enum cases
- ConsentTypeTest updated: invalid-string rejection tested via from(), equality via assertSame() on enum singletons


johanib (approved):

johanib:
Looks good, but need to fix test before merging

johanib (approved):

PR #1952, opened 2026-03-20
Remove reserved schacHomeOrganization validator

### LINKED ISSUE #1296: Remove coin:schacHomeOrganization feature
There's a feature in Engineblock that you can set the field `coin:schachomeorganization` for an IdP _A_, say to example.gov. This does two things:

1. It sets the value of the released schacHomeOrganization of this IdP statically to example.gov on outgoing assertions, regardless of the input attribute value.
1. It makes this value reserved, i.e. other IdPs cannot deliver this schacHomeOrganization anymore, it's checked that incoming SchacHomeOrganization of an IdP _B_ is not listed in `coin:schachomeorganization` for any other IdP. You can enable this check in `attributes.json`.

Point 2 is broken en when enabled generates an Exception.

Proposed to remove since it's unused, at least partly broken and introduces quite a bit of complexity in the login processing.

kayjoosten:
## Summary
- remove the legacy reserved schacHomeOrganization validator path
- keep the outgoing IdP static schacHomeOrganization override intact
- remove metadata repository support and tests that only served the deleted validator

Closes #1296


johanib (approved):
👍 Looks like a clean cleanup

PR #1954, opened 2026-03-24
refactor: remove direct xmlseclib calls (issue #1939)

### LINKED ISSUE #1939: Remove direct calls to xmlseclib
Most of our SAML/XML handling is done via the SimpleSAMLphp/SAML2 library. But in three places, we still call xmlseclibs directly:

- https://github.com/OpenConext/OpenConext-engineblock/blob/main/src/OpenConext/EngineBlock/Metadata/X509/X509PrivateKey.php#L64
- https://github.com/OpenConext/OpenConext-engineblock/blob/main/src/OpenConext/EngineBlock/Xml/DocumentSigner.php#L46
- https://github.com/OpenConext/OpenConext-engineblock/blob/main/library/EngineBlock/X509/PrivateKey.php#L45

Investigate whether we can replace these with functionality of SAML2.

The advantage of that is that we have a single point of entrey into xmlseclibs, and because SAML2 is implementing additional safeguards against wrapping attacks and such, it makes us a little more robust against xmlseclibs/libxml bugs.

kayjoosten:
## Summary

- Replaces direct `XMLSecurityDSig` usage in `DocumentSigner` with `SAML2\Utils::insertSignature()`, routing metadata XML signing through the SAML2 library which adds XML wrapping-attack protection
- Fixes a latent bug where the private key was loaded twice (`toXmlSecurityKey()` already loaded it; the redundant `loadKey()` call is gone)
- Fixes missing `loadXML()` error handling in `DocumentSigner`
- Adds clarifying docblocks to `X509PrivateKey::toXmlSecurityKey()` (modern) and `EngineBlock_X509_PrivateKey::toXmlSecurityKey()` (legacy) explaining they are the intentional single bridges to xmlseclib — necessary because SAML2 v4's public API still requires `XMLSecurityKey` objects as input

## Test Plan

- [ ] `DocumentSignerTest` — 2 tests pass (metadata XML signing produces identical output)
- [ ] `X509PrivateKeyTest` — 2 tests pass (key loading and signing unaffected)
- [ ] PHPCS clean on all changed files
- [ ] `grep -rn 'use RobRichards\\XMLSecLibs' src/ library/ --include="*.php" | grep -v 'FunctionalTestingBundle\|/Tests/'` — `DocumentSigner.php` no longer appears; only the two intentional bridge classes and constant-only usages remain

Closes #1939

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.worktrees/
 config/packages/parameters.yml
 
 var/cache/*
--- a/library/EngineBlock/X509/PrivateKey.php
+++ b/library/EngineBlock/X509/PrivateKey.php
@@ -22,7 +22,7 @@ class EngineBlock_X509_PrivateKey
 {
     private $_filePath;
 
-    function __construct($filePath)
+    public function __construct($filePath)
     {
         if (!file_exists($filePath)) {
             throw new EngineBlock_Exception(sprintf('Private key file "%s" does not exist.', $filePath));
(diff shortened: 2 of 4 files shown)


johanib (approved):
LGTM

This is what's possible within wat saml2 currently offers 👍

PR #1957, opened 2026-03-27
Reject metadata push when IdP discovery entry is missing English name

### LINKED ISSUE #1865: EB: Push metadata from Manage should ensure IdP has discovery information
Based on: https://github.com/OpenConext/OpenConext-devconf/pull/60

Manage migrated the IdP name to the discovery name fields.
However, if for some reason manage does not post the discoveries in the metadata push to EB, EB will still accept the metadata push. 

However, when rendering the wayf, EB will crash with a type error (NULL passed, string expected).

**Change request:**
EB should not accept the Metadata push if the discovery name is empty (or fallback to the IdP name during metadata push).

kayjoosten:
## Summary

- Fixes #1865: when Manage pushes IdP metadata with a `discoveries` entry that has no English name (`name_en`), EngineBlock now rejects the push with HTTP 400 instead of silently dropping the entry
- Changed `DiscoveryAssembler::assembleDiscoveries()` to throw `InvalidDiscoveryException` instead of silently skipping; the existing `ConnectionsController` catch-all propagates this as HTTP 400
- Added tests at three levels: unit (`DiscoveryAssemblerTest`), assembler integration (`PushMetadataAssemblerTest`), and HTTP functional (`ConnectionsControllerTest`)

## Test plan

- [ ] Run unit and integration tests: `./vendor/bin/phpunit --configuration=./tests/phpunit.xml --testsuite=unit,integration`
- [ ] Verify `POST /api/connections` with an IdP discovery missing `name_en` returns HTTP 400 with an error message containing `name_en`
- [ ] Verify a normal metadata push with a valid discovery (including `name_en`) still succeeds

### DIFF
--- a/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/DiscoveryAssembler.php
+++ b/src/OpenConext/EngineBlock/Metadata/Entity/Assembler/DiscoveryAssembler.php
@@ -42,16 +42,21 @@ class DiscoveryAssembler
             return [];
         }
 
+        $entityId = $connection->name ?? 'unknown';
         $discoveries = [];
-        foreach ($connection->metadata->discoveries as $discovery) {
+        foreach ($connection->metadata->discoveries as $index => $discovery) {
             $names = $this->extractLocalizedFields($discovery, 'name');
             $keywords = $this->extractLocalizedFields($discovery, 'keywords');
             $logo = $this->assembleLogo($discovery);
 
             if (!isset($names['en'])) {
                 throw new InvalidDiscoveryException(
-                    'A discovery entry is missing a required English name (name_en). ' .
-                    'Refusing to process metadata push.'
+                    sprintf(
+                        'Discovery entry #%d for entity "%s" is missing a required English name (name_en). ' .
+                        'Refusing to process metadata push.',
+                        $index,
+                        $entityId
+                    )
                 );
             }
             $discoveries[] = Discovery::create($names, $keywords, $logo);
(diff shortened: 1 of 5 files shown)


src/OpenConext/EngineBlock/Metadata/Entity/Assembler/DiscoveryAssembler.php
```diff
@@ -47,9 +48,13 @@ public function assembleDiscoveries(stdClass $connection): array
             $keywords = $this->extractLocalizedFields($discovery, 'keywords');
             $logo = $this->assembleLogo($discovery);
 
-            if (isset($names['en'])) {
-                $discoveries[] = Discovery::create($names, $keywords, $logo);
+            if (!isset($names['en'])) {
```
johanib, line 52: Maybe good to add  the problem entity to the response, otherwise you don't know which of the 1000 services might not be correct.
```php
public function assembleDiscoveries(stdClass $connection): array
{
    if (!isset($connection->metadata->discoveries)) {
        return [];
    }
    $entityId = $connection->name ?? 'unknown';
    $discoveries = [];
    
    foreach ($connection->metadata->discoveries as $index => $discovery) {
        $names = $this->extractLocalizedFields($discovery, 'name');
        $keywords = $this->extractLocalizedFields($discovery, 'keywords');
        $logo = $this->assembleLogo($discovery);
        if (!isset($names['en'])) {
            throw new InvalidDiscoveryException(
                sprintf(
                    'Discovery entry #%d for entity "%s" is missing a required English name (name_en). ' .
                    'Refusing to process metadata push.',
                    $index,
                    $entityId
                )
            );
        }
        $discoveries[] = Discovery::create($names, $keywords, $logo);
    }
    return empty($discoveries) ? [] : ['discoveries' => $discoveries];
}
```

johanib (changes requested):
lgtm, 1 improvement that we should still add imo

johanib (approved):

PR #1959, opened 2026-03-31
issue-1931 nameid lookup api

### LINKED ISSUE #1931: Add internal API to look up NameID for specific user/SP
# New Feature Request

## Summary

For security reasons (malware, hacked accounts) it is sometimes necessary to find a nameid of a specific user, given their `shcacHomeOrganisation` and `uid`.  We currently have separate scripts to do this, which read directly from the database.

Instead, it would be much cleaner to let Engineblock handle this itself.  This would also allow us to automate this and make it available in a self-service interface to authorized users.

## Proposed Solution

Therefore I propose to add two internal, EB-API endpoints that can handle these type of lookups:
 - `/info/users/nameid`: translate sho/uid/sp to nameid
 - `/info/users/id`: translate nameid to sho/uid/sp

### id --> nameid
Note: this call could work even if the user has never logged into this service yet.  I.e., the resulting anmeid doesn't have to exist yet, because it can eb calculated based on the user uuid and service.  Possibly: add a flag in the output if the nameid was calculated instead of lookup up in the database.

Example:
```
POST https://engine-api.openconxt.dev/info/users/nameid
[
  {"schacHomeOrganization": "example.edu", "uid": "student001", "sp_entityid": "https://service.example.com/"},
  {"schacHomeOrganization": "example.org", "uid": "employee001", "sp_entityid": "https://service.example.com/"}
]

["62546075f3ad801acdad821aacb006ff61298603","5e76eca7df5c64ec30b55071bd9a47716457a480"]
```

### nameid --> id
Examples:
```
POST https://engine-api.openconxt.dev/info/users/id
["62546075f3ad801acdad821aacb006ff61298603","5e76eca7df5c64ec30b55071bd9a47716457a480"]

[
  {"schacHomeOrganization": "example.edu", "uid": "student001", "sp_entityid": "https://service.example.com/"},
  {"schacHomeOrganization": "example.org", "uid": "employee001", "sp_entityid": "https://service.example.com/"}
]
```

## logging
For each of the requests, log which ids were requested and by which user.

kayjoosten:


config/packages/ci/parameters.yml
```diff
@@ -1,4 +1,8 @@
 parameters:
+    api.users.nameIdLookup.username: nameid
+    api.users.nameIdLookup.password: secret
+    feature_api_users_nameid: true
+    feature_api_users_id: true
```
johanib: The original ticket does not mention feature flags. I see the deprovision controller does also use a feature flag. Did you check with Bas if we want feature flags?

I think it fits, but would be good to doublecheck.
Also, one or two feature flags?

config/services/services.yml
```diff
@@ -87,6 +87,13 @@ services:
             - '@OpenConext\EngineBlockBundle\Authentication\Repository\SamlPersistentIdRepository'
             - '@OpenConext\EngineBlockBundle\Authentication\Repository\ServiceProviderUuidRepository'
 
+    OpenConext\EngineBlock\Service\NameIdLookupService:
```
johanib, line 90: Open question: Should this be here or in th api namespace / service.yaml?

migrations/DoctrineMigrations/Version20260331000000.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace OpenConext\EngineBlock\Doctrine\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Corrects the column comment on saml_persistent_id.persistent_id.
+ *
+ * The original comment read "SHA1 of service_provider_uuid + user_uuid", which was
+ * inaccurate in two ways: the operand order was wrong, and the COIN: salt was omitted.
```
johanib, line 29: `salt was omitted` What does this mean? There is no salt in the new comment?

edit: Oh, 'COIN:' is the salt :)

johanib: Alternatively, I wasn't aware we were keeping documentation in the database comments.. So personally, I'd just remove the comment. 

But might have been a good reason to do so. So, lets just keep your current migration, as that's what was discussed with Bas.

src/OpenConext/EngineBlock/Service/NameIdLookupService.php
```diff
@@ -0,0 +1,109 @@
...
+        private readonly UserRepository $userRepository,
+        private readonly ServiceProviderUuidRepository $spUuidRepository,
+        private readonly SamlPersistentIdRepository $persistentIdRepository,
+        private readonly LoggerInterface $logger
+    ) {
+    }
+
+    public function resolveNameId(string $schacHomeOrganization, string $uid, string $spEntityId): ?array
+    {
+        $collabPersonId = new CollabPersonId(sprintf(
```
johanib: Use `\OpenConext\EngineBlock\Authentication\Value\CollabPersonId::generateWithReplacedAtSignFrom` instead?

src/OpenConext/EngineBlock/Service/NameIdLookupService.php
```diff
@@ -0,0 +1,109 @@
...
+            '%s:%s:%s',
+            CollabPersonId::URN_NAMESPACE,
+            $schacHomeOrganization,
+            str_replace('@', '_', $uid)
+        ));
+
+        $user = $this->userRepository->findByCollabPersonId($collabPersonId);
+        if ($user === null) {
+            $this->logger->debug('NameIdLookupService: user not found', [
```
johanib, line 50: debating whether a service should log.

Also, because you return an array, I think it would be nicer to return a VO, that can also return a not-found reason or something. 

Because the service in this context does not know what is happening. In this context (the api lookup), you might want this level of logging. 
But in other theoretical context, it might be totally expected to not find the actual user, and the logs are just noise.

Otoh, this is how other services log.. So for consistency, OK to leave it in.

src/OpenConext/EngineBlock/Service/NameIdLookupService.php
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        $parts = explode(':', $user->collabPersonId->getCollabPersonId(), 5);
+        $schacHomeOrganization = $parts[3] ?? '';
+        $uid = $parts[4] ?? '';
+
+        return [
+            'schacHomeOrganization' => $schacHomeOrganization,
+            'uid' => $uid,
+            'sp_entityid' => $spEntityId,
```
johanib: I would opt for a VO, as the return type is a random array.

src/OpenConext/EngineBlock/Service/NameIdLookupService.php
```diff
@@ -0,0 +1,109 @@
...
+        $spEntityId = $this->spUuidRepository->findEntityIdByUuid($entry->serviceProviderUuid);
+        if ($spEntityId === null) {
+            $this->logger->warning(
+                'NameIdLookupService: saml_persistent_id entry exists but SP UUID record is missing',
+                ['serviceProviderUuid' => $entry->serviceProviderUuid]
+            );
+            return null;
+        }
+
+        $parts = explode(':', $user->collabPersonId->getCollabPersonId(), 5);
```
johanib: Here, the CollabPersonId is being decomposed. 

It would be better to extract this piece of logic out of this service, maybe into the `CollabPersonId->getHomeOrganisation`, etc.

Also, the `@` is not being put back in? I'm not sure what should happen with that, the original script does not have the `@` replacement part. Not sure if relevant.

src/OpenConext/EngineBlock/Service/NameIdLookupService.php
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        $userUuid = $user->collabPersonUuid->getUuid();
+        $stored = $this->persistentIdRepository->findByUserAndSpUuid($userUuid, $spUuid);
+
+        if ($stored !== null) {
+            return ['nameid' => $stored->persistentId, 'stored' => true];
+        }
+
+        $calculatedNameId = sha1(self::PERSISTENT_NAMEID_SALT . $userUuid . $spUuid);
```
johanib: Extract to a new factory method: `SamlPersistentId::generate(string $userUuid, string $spUuid)`?

src/OpenConext/EngineBlock/Service/NameIdLookupService.php
```diff
@@ -0,0 +1,109 @@
...
+            $this->logger->warning(
+                'NameIdLookupService: saml_persistent_id entry exists but SP UUID record is missing',
+                ['serviceProviderUuid' => $entry->serviceProviderUuid]
+            );
+            return null;
+        }
+
+        $parts = explode(':', $user->collabPersonId->getCollabPersonId(), 5);
+        $schacHomeOrganization = $parts[3] ?? '';
+        $uid = $parts[4] ?? '';
```
johanib: Should it fallback to ''? It should fail if something fails that should not happen.

src/OpenConext/EngineBlockBundle/Authentication/Repository/SamlPersistentIdRepository.php
```diff
@@ -57,4 +57,17 @@ public function deleteByUuid(CollabPersonUuid $uuid)
...
+    /**
+     * @param string $userUuid
+     * @param string $serviceProviderUuid
+     * @return SamlPersistentId|null
+     */
+    public function findByUserAndSpUuid(string $userUuid, string $serviceProviderUuid): ?SamlPersistentId
+    {
+        return $this->findOneBy([
+            'userUuid' => $userUuid,
+            'serviceProviderUuid' => $serviceProviderUuid,
```
johanib, line 70: Nitpick: Not sure if theoretical, but what if two matching records exist? It would return a random of the existing records. 

You could argue that db constraints should prevent that from occuring if that is the constraint, so ok to leave it as is.

src/OpenConext/EngineBlockBundle/Authentication/Repository/UserRepository.php
```diff
@@ -74,4 +75,18 @@ public function deleteUserWithCollabPersonId(CollabPersonId $collabPersonId)
...
+
+    /**
+     * @param CollabPersonUuid $uuid
+     * @return null|User
+     */
+    public function findByCollabPersonUuid(CollabPersonUuid $uuid): ?User
+    {
+        return $this
+            ->createQueryBuilder('u')
+            ->where('u.collabPersonUuid = :uuid')
```
johanib, line 87: This is indexed, but not a unique key in the db 🤔

src/OpenConext/EngineBlockBundle/Controller/Api/UserController.php
```diff
@@ -0,0 +1,167 @@
...
+    #[Route(path: '/info/users/nameid', name: 'api_users_nameid', defaults: ['_format' => 'json'], methods: ['POST'])]
+    public function nameIdAction(Request $request): JsonResponse
+    {
+        if (!$this->featureConfiguration->isEnabled('api.users_nameid')) {
+            throw new ApiNotFoundHttpException('NameID lookup API is disabled');
+        }
+
+        $this->assertAuthorized();
+
+        $entries = $this->decodeJsonArray($request);
```
johanib, line 63: I see the supplied python example supplies multiple ids, and returns multiple ids. And is a POST. 

Since this is a lookup, it would be much nicer to:
* Make it GET
* Accept three named route parameters.

It's not just the the code that would become better, but in general, you search for one thing at a time.
But I'm not sure what the context is. If this is a 'batch' endpoint, we can rename it as such. 

@baszoetekouw Is that an option, from a consumer perspective?
In summary: Should this route accept multiple lookups per request?

tests/unit/OpenConext/EngineBlock/Service/NameIdLookupServiceTest.php
```diff
@@ -0,0 +1,294 @@
...
+use OpenConext\EngineBlockBundle\Authentication\Repository\SamlPersistentIdRepository;
+use OpenConext\EngineBlockBundle\Authentication\Repository\ServiceProviderUuidRepository;
+use OpenConext\EngineBlockBundle\Authentication\Repository\UserRepository;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
+use Ramsey\Uuid\Uuid;
+
+final class NameIdLookupServiceTest extends TestCase
```
johanib, line 39: The service doesn't really do anything, except tie together other parts.  Especially after the relevant business logic is extracted to the correct places. 
So, this test just tests the implementation details. 

I'd be ok with dropping it.

johanib (changes requested):

config/services/services.yml
kayjoosten: NameIdLookupService is a domain service, not an API-specific concern it could in theory be used outside the API context too. The controller (UserController) is correctly registered in config/services/controllers/api.yml. Keeping the service in services.yml follows the same pattern as other domain services in this project.

src/OpenConext/EngineBlockBundle/Authentication/Repository/SamlPersistentIdRepository.php
kayjoosten: Duplicates for the same user+SP combination are impossible by design: persistentId is the primary key and is computed as sha1('COIN:' + userUuid + serviceProviderUuid). The same pair always produces the same SHA1, so the DB would reject a duplicate insert. findOneBy is safe here.

src/OpenConext/EngineBlockBundle/Authentication/Repository/UserRepository.php
kayjoosten: You're right, idx_user_uuid on the user table is a regular index, not a UNIQUE constraint. 

However, UUIDs are generated via CollabPersonUuid::generate() which uses Uuid::uuid4(). Collisions are astronomically unlikely, and the getOneOrNullResult() call would throw a NonUniqueResultException if it ever happened, making the problem visible rather than silent. 

This is an application-level guarantee, same as the existing findByCollabPersonId method in the same repository which follows the same pattern. 

Adding a UNIQUE constraint is a separate concern outside this PR's scope i think. What do you think?

config/packages/ci/parameters.yml
kayjoosten: I have discussed with bas and we have decided that 1 feature flag will be enough so ill change that

src/OpenConext/EngineBlockBundle/Controller/Api/UserController.php
kayjoosten: I have discussed this with bas as well the batch endpoints are intended

config/packages/ci/parameters.yml
johanib: @kayjoosten 
Still 2 feature flags?

Is this PR ready for re-review?

src/OpenConext/EngineBlock/Service/NameIdLookupService.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        $userUuid = $user->collabPersonUuid->getUuid();
+        $stored = $this->persistentIdRepository->findByUserAndSpUuid($userUuid, $spUuid);
+
+        if ($stored !== null) {
+            return new NameIdResult($stored->persistentId, true);
+        }
+
+        return new NameIdResult(SamlPersistentId::generate($userUuid, $spUuid)->persistentId, false);
```
johanib, line 69: 💯

src/OpenConext/EngineBlockBundle/Authentication/Repository/SamlPersistentIdRepository.php
johanib: Split into https://github.com/OpenConext/OpenConext-engineblock/issues/1974

src/OpenConext/EngineBlockBundle/Authentication/Repository/UserRepository.php
johanib: https://github.com/OpenConext/OpenConext-engineblock/issues/1974

tests/functional/OpenConext/EngineBlockBundle/Controller/Api/UserControllerTest.php
```diff
@@ -0,0 +1,347 @@
...
+        $conn = $this->connection();
+        $conn->executeStatement('DELETE FROM saml_persistent_id');
+        $conn->executeStatement('DELETE FROM service_provider_uuid');
+        $conn->executeStatement('DELETE FROM user');
+    }
+
+    private function insertUserFixture(string $userUuid): void
+    {
+        $collabId = 'urn:collab:person:' . self::SHO . ':' . self::UID;
+        $this->connection()->executeStatement(
```
johanib, line 51: I'm triggered by the use of fixture methods inside the tests. 

this is not reusable, and can get messy in the long run. 

But, afaiks, there is no pre existing pattern that does it differently. And as long as the delete queries do not cause issues with other tests (which would indicate issues/tech dept at other places), this should not be a problem. 

So, good enough to do it like this for now.

config/packages/parameters.yml.dist
```diff
@@ -77,6 +77,8 @@ parameters:
     api.users.profile.password: secret
     api.users.deprovision.username: lifecycle
     api.users.deprovision.password: secret
+    api.users.nameIdLookup.username: nameid
+    api.users.nameIdLookup.password: secret
```
johanib: Every other parameter name is in lowercase. Probably better to remain consistent.

johanib (approved):

PR #1966, opened 2026-04-08
Revert feature flag to enable/disable content hash cleanup

johanib:
The `feature_stable_consent_hash_migration` feature flag was confusing, and setting the old hash to null provided no real value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,9 +29,7 @@ Changes:
 
 * Stabilized consent checks
   * In order to make the consent hashes more robust, a more consistent way of hashing the user attributes has been introduced
-  * This feature automatically migrates from the old hashes to the new hashes, cleaning up the old hash.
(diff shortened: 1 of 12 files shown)


kayjoosten (changes requested):
I miss the migration where we make the consent hash nullable that is not needed anymore after this change right?

johanib:
> I miss the migration where we make the consent hash nullable that is not needed anymore after this change right

With this change, we no longer generate the old hashes? New hashes are always the stable hashes. So new users insert null into the old hash column?

johanib:
Discussed with Kay: We can remove the dropping of the not null from the current migration. If this were to be a new migration, we would not, as it does not really matter. But removing it prevents a unneeded operation on the prod db.

johanib:
> I miss the migration where we make the consent hash nullable that is not needed anymore after this change right?

Good catch! :+1:  Updated the migration (:see_no_evil: )

kayjoosten (approved):

PR #1985, opened 2026-04-22
Add support for preferred IdPs in WAYF display

### LINKED ISSUE #1970: Show preferred IdPs on a special position in the WAYF
Some IdPs have  a different status than regular IdPs. For example, EduId/MyConext already get an explicit notice in the WAYF: 

Image

We're also going to introduce a (proxy) IdP for all international users, which should be similarly made more visible than a regular IdP.

I therefore propose to replace the explicit "If your institution is not listed, eduID is available as an alternative" notice by the concept of "Preferred IdPs".  (In a backward-compatible way, see https://github.com/OpenConext/OpenConext-engineblock/issues/1970#issuecomment-4242120044)

Such IdPs are configured in the `parameters.yml` (list of entityids) , and will not be shown in the regular search list, but will have separate buttons at the top of the WAYF screen:

Image

If no priority-IdPs are connected to this SP, the WAYF should stay as it is:

Image

A user who has previously selected an IdP, still only sees that single one:

Image

Pressing the "Use another account" button leads to the full screen as shown above.

johanib:
Prior to this change, there was no way to configure priority IdPs.

This change adds a `wayf.preferred_idp_entity_ids` parameter to configure IdPs that should show prominent in the wayf.
IdPs in this list are shown on top of the wayf, outside of the regular list.

Also make the twig templates more explicit by not enabling the global variable scope.

Resolves #1970


src/OpenConext/EngineBlockBundle/Service/WayfRenderer.php
```diff
@@ -0,0 +1,86 @@
...
+        bool $backLink,
+        int $cutoffPoint,
+        bool $rememberChoice,
+        bool $showRequestAccess,
+        string $requestId,
+        ServiceProvider $serviceProvider,
+    ): string {
+        $split = $this->splitter->split($idpList, $preferredIdpEntityIds);
+        $showPreferredIdps = !empty($split['preferred']);
+        $isDefaultIdpPreferred = in_array($defaultIdpEntityId, $preferredIdpEntityIds, true);
```
kayjoosten: The banner is suppressed when the default IdP is in `$preferredIdpEntityIds`, but this is the raw config list it doesn't know whether the IdP is actually connected to this SP. `IdpSplitter` silently drops disconnected preferred IdPs, so the default IdP can be "on the guest list" but never show up to the party.

When this happens the default IdP is invisible everywhere: not in the preferred section (dropped by splitter), not in the regular list (preferred IdPs are excluded from regular), and the banner is suppressed too.

```php
// checks the config — doesn't know if the IdP is connected
$isDefaultIdpPreferred = in_array($defaultIdpEntityId, $preferredIdpEntityIds, true);

// fix: check what the splitter actually produced
$preferredEntityIdsShown = array_column($split['preferred'], 'EntityID');
$isDefaultIdpPreferred = in_array($defaultIdpEntityId, $preferredEntityIdsShown, true);
```

theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/preferredIdps.html.twig
```diff
@@ -0,0 +1,7 @@
+<section class="wayf__preferredIdps {% if connectedIdps.formattedPreviousSelectionList is not empty %}hidden{% endif %}">
```
kayjoosten, line 1: `remainingIdps.html.twig` has a visually-hidden `<h2>` so screenreader users know where they are. `preferredIdps.html.twig` has none a screenreader user landing in this section gets no context.

```twig
<section class="wayf__preferredIdps {% if connectedIdps.formattedPreviousSelectionList is not empty %}hidden{% endif %}">
    <h2 class="visually-hidden">{{ 'wayf_preferred_idps_title_screenreader'|trans }}</h2>
    <ul class="wayf__idpList wayf__idpList--preferred">
```

Or is it not needed here? What do you think?

src/OpenConext/EngineBlockBundle/Service/WayfRenderer.php
```diff
@@ -0,0 +1,86 @@
...
+            greenHeader: $serviceProvider->getDisplayName($currentLocale),
+            helpLink: '/authentication/idp/help-discover?lang=' . $currentLocale,
+            backLink: $backLink,
+            cutoffPointForShowingUnfilteredIdps: $cutoffPoint,
+            showIdPBanner: $showIdPBanner,
+            rememberChoiceFeature: $rememberChoice,
+            showRequestAccess: $showRequestAccess,
+            requestId: $requestId,
+            serviceProvider: $serviceProvider,
+            showRequestAccessContainer: true,
```
kayjoosten: **`showRequestAccessContainer` is hardcoded `true` the parameter does nothing**

`WayfRenderer::render()` always passes `showRequestAccessContainer: true` to the factory. It is carried through `WayfViewModelFactory` and stored in `WayfViewModel`, but no caller can ever change it. 

If it should always be `true`, can you not hardcode it in the factory?

kayjoosten (changes requested):

src/OpenConext/EngineBlockBundle/Service/WayfRenderer.php
johanib: :exploding_head: 

This took some time to figure out. 

If the default IdP (EduID) was not connected, the EduID banner would not show because the splitter correctly dropped EduID from the perferred section, but the banner logic checked the raw config list instead of what the splitter actually produced and incorrectly concluded the banner was not needed.

I verified and applied your fix because it makes sense.

But do note, this scenario could not actually occur on production, because the
```php
            // Do not show the default IdP in the disconnected IdPs section.
            if (!$isAccessible && $isDefaultIdP) {
                continue;
            }
```
prevents that scenario.

What would happen in this scenario
[Scenario](https://engine.dev.openconext.local/functional-testing/wayf?connectedIdps=3&unconnectedIdps=1&defaultIdpEntityId=https://unconnected.example.com/entityId/1&preferredIdpEntityIds%5B%5D=https://unconnected.example.com/entityId/1&preferredIdpEntityIds%5B%5D=https://example.com/entityId/2&addDiscoveries=false)
image

After the fix:

image  


But after by commit, that fixed scenario is still not visible, because I also prevented this scenario from forming on the functional-testing route, to accurately mirror the production scenarios.

johanib: One reason this situation/bug could occur was that the datastructure was just arrays, providing no encapsulation of internal data. 

So in line with the rest of the PR, I also improved the IdpSplitter to return a value object instead of a array.

theme/base/templates/modules/Authentication/View/Proxy/Partials/WAYF/preferredIdps.html.twig
johanib: The main header kinda says all there is to say. "Select an account to login". 
So I dont think an extra title is needed for screenreaders. This would create asymmetry. If a header is needed for screen readers, it should be added visually as well.
So I don't think it adds anything in this case.

src/OpenConext/EngineBlockBundle/Service/WayfRenderer.php
johanib: Ah, this was a leftover from the openconext theme. The theme actually uses an `is defined` check. Which is always I deleted the parameter.

kayjoosten (approved):

PR #1986, opened 2026-04-22
Add logging when old SAML requests are used

### LINKED ISSUE #1972: Check for and log old IssueInstants
As described in #1935, we want to know if SPs send SAMLRequests with old (>24h) `IssueInstant`s.  Add code to check for old `IssueInstant`s in SAMLRequests.  If they are detected, write a warning to the logs, but don't block the requests (no functional change).

johanib:
Prior to this change, EB would not make note of the fact that old SAML requests are used. Engine does make a log notice about possible clock drift, but does so if if the time is off by 30 seconds.

This change adds a warning if a request is received that is X seconds old. With the default being 1 day.

Functionally, EB does proces these requests as usual, but SPs might reject the requests.

Resolves #1792

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Changes:
   * The `0000-00-00 00:00:00` is added for clarity/consistency, as this is probably the default behaviour of your database already.
 * Removed unused index `consent.deleted_at`. Delete this from your production database if it's there.
 * Added a specific error page for unsolicited SAML responses (IdP-initiated SSO without a prior AuthnRequest).
+* Added `max_issue_instant_age` to parameters.yaml to configure the logging mechanism. EB will write log entries if it receives requests that are older than this value.
 
 * Stabilized consent checks
   * In order to make the consent hashes more robust, a more consistent way of hashing the user attributes has been introduced
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Changes:
   * The `0000-00-00 00:00:00` is added for clarity/consistency, as this is probably the default behaviour of your database already.
 * Removed unused index `consent.deleted_at`. Delete this from your production database if it's there.
 * Added a specific error page for unsolicited SAML responses (IdP-initiated SSO without a prior AuthnRequest).
+* Added `max_issue_instant_age` to parameters.yml to configure the logging mechanism. EB will write log entries if it receives requests that are older than this value.
 
 * Stabilized consent checks
   * In order to make the consent hashes more robust, a more consistent way of hashing the user attributes has been introduced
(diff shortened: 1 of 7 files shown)


tests/library/EngineBlock/Test/Corto/Module/BindingsTest.php
```diff
@@ -109,6 +110,72 @@ public function test_saml2_library_error_messages_we_specifically_test_have_not_
...
+        $this->injectLogger($bindings, $logger);
+
+        $method = new ReflectionMethod(EngineBlock_Corto_Module_Bindings::class, '_checkIssueInstant');
+        $method->invoke($bindings, $issueInstant, 'SP', 'https://sp.example.edu');
+    }
+
+    public function test_checkIssueInstant_does_not_warn_when_request_is_within_max_age(): void
+    {
+        $maxAge = 86400;
+        $issueInstant = time() - $maxAge + 1;
```
kayjoosten: 1 second seems flaky to me. What if computing for some reason takes longer the test will fail without i clear message. Maybe you can increase this to 60. What do you think?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Changes:
   * The `0000-00-00 00:00:00` is added for clarity/consistency, as this is probably the default behaviour of your database already.
 * Removed unused index `consent.deleted_at`. Delete this from your production database if it's there.
 * Added a specific error page for unsolicited SAML responses (IdP-initiated SSO without a prior AuthnRequest).
+* Added `max_issue_instant_age` to parameters.yaml to configure the logging mechanism. EB will write log entries if it receives requests that are older than this value.
```
kayjoosten: Very nitpick: the parameter reference isn't correct. It should be yml instead of yaml. As i said very nitpicky you dont have to change this.

Image

config/reference.php
```diff
@@ -1494,6 +1492,17 @@
  *         debug?: DebugConfig,
  *         web_profiler?: WebProfilerConfig,
  *     },
+ *     "when@prod"?: array{
```
kayjoosten, line 1495: just wondering why is this needed?

johanib: its not needed. Just reference.php being annoying: https://github.com/symfony/symfony/discussions/62588#discussioncomment-16100438

tests/library/EngineBlock/Test/Corto/Module/BindingsTest.php
johanib: I agree that 1 second can trigger false failures. 60 seconds is  a bit much, the whole corto test suite takes 500ms seconds to complete. Adjusted to 5 seconds.

CHANGELOG.md
johanib: adjusted :+1:

kayjoosten (approved):

PR #1992, opened 2026-05-04
Consolidate simple feedback routes into generic feedbackAction

### LINKED ISSUE #1758: De-clutter the Symfony error reporting setup
[This issue is imported from pivotal - Originaly created at Aug 20, 2019 by Michiel Kodde](https://www.pivotaltracker.com/story/show/167994147)

At this point, we receive an error situation from the kernel error listeners. This is the famous file with the longest if/elseif construction in history. Where certain exception types are matched with a custom error report.

They are then redirected to the FeedbackController who in turn renders a custom twig error template. Before doing so, twig functions are used to gather important error details. The `Feedback` Twig extension is a good example of this. It uses the Corto application object to load the feedback information from the session.

Needless to say. This setup can be improved upon. And #107565968 is an important part of doing so.

Instead of waiting for Corto to become obsolete. We can start with cleaning up technical debt in the way we now render the error templates.

A first effort was undertaken to fix this issue. The remnants can be found in: the [optimize-error-routing](https://github.com/OpenConext/OpenConext-engineblock/tree/feature/optimize-error-routing) feature branch.

In short, the following change was proposed and partly executed. But ended due to scope creep:

1. Routing configuration in feedback.yml is amended with additional route parameters. Being a page identifier and the expected HTTP error code.
     1. The page identifier should be used to load the custom error message and page title from the translations file.
     2. The error code will be used in the response to set the appropriate error code.
2. Any FeedbackController routes that do custom logic can be kept, all other routes should be updated to the generic error feedback route (already implemented in the branch)
3. Those errors using the generic route can be rid of their custom twig template. And a generic twig template should be introduced.
4. Some additional complexion lies in the retrieval of the feedback information. A new helper should be introduced. Tasked with retrieving this information from the Corto application session. This was partly implemented in the feature branch.
5. The custom translations for the error pages need to be translated. In doing so, the organization noun and other translation overrides must be applied. This was previously done in a twig extension of the translation function. This might need to move?

Greatest reason for not doing these tasks under the boyscout rule is the amount of changes required to apply this change. Additional tests, updating of tests, and rigorous user testing are required when applying this change.

johanib:
Prior to this change, there were many 'empty' identical twig templates in the error process. This change replaces them with a 'generic-error' twig template.

Resolves #1758


CHANGELOG.md
```diff
@@ -15,6 +15,27 @@ the EngineBlock wiki.
 Features:
 * Added `coin:azure_domain_hint` configuration option for IdPs. When set, EngineBlock appends a `whr=<domain>` query parameter to the HTTP-Redirect AuthnRequest sent to the IdP, allowing Microsoft Azure / EntraID to skip the account picker (#1864).
 
+### Translation key changes
+#### `error_invalid_acs_location`
+
+The `error_invalid_acs_location` translation key has changed meaning. Previously it held the **error description** text. It now holds the **page title**.
+
+If you have overridden this key in your theme translations (`theme/{name}/translations/messages.*.php`), rename it to `error_invalid_acs_location_desc` and add a new `error_invalid_acs_location` entry for the page title.
```
kayjoosten, line 33: Nice, but we are missing some key. You have also changed error_no_message, error_no_message_desc, error_stepup_callout_unknown_title and error_stepup_callout_user_cancelled_title.

CHANGELOG.md
```diff
@@ -15,6 +15,27 @@ the EngineBlock wiki.
...
+```
+
+**After:**
+```php
+'error_invalid_acs_location'      => 'Error - Invalid ACS location',
+'error_invalid_acs_location_desc' => 'Your custom description text.',
+```
+
+See https://github.com/OpenConext/OpenConext-engineblock/issues/1758
+
```
kayjoosten, line 47: also i would recommend writing something about the changed status codes. So monitoring doesn't break if you think that would be a possibility.

kayjoosten (approved):

PR #1994, opened 2026-05-08
Issue-1980 remove openconext theme

### LINKED ISSUE #1980: Repair or delete OpenConext theme
SURF currently uses the skeune theme, which is extends the base theme with minimal changes.

The openconext theme is currently broken. It has missing variables and is ignored. 

We need to either fix it and ensure it remains maintained, or remove it.

The theme is incomplete and would render broken for anyone actually using it (missing error pages, outdated consent/WAYF structure).

The main question: Is anyone actually deploying engineblock with the openconext theme in production?

kayjoosten:


theme/scripts/copy-assets.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-THEME=${EB_THEME:-openconext}
```
johanib, line 4: One could argue that the fallback / default should be 'base' instead of 'skeune'. (not saying this is the case per se).

johanib:
theme/README.md

still references the deleted theme.

theme/scripts/copy-assets.sh
johanib: Discussed: Not for now.

johanib (approved):

johanib (approved):

PR #1999, opened 2026-05-13
fix: remove HTTP Redirect binding support from AcsRequestValidator

### LINKED ISSUE #1750: Remove HTTP redirect support in AcsRequestValidator
[This issue is imported from pivotal - Originaly created at Oct 29, 2019 by Michiel Kodde](https://www.pivotaltracker.com/story/show/169422179)

**Problem**
Validator: `\OpenConext\EngineBlock\Validator\AcsRequestValidator` supports both HTTP POST and  Redirect binding. Where EB only supports POST binding.

**Result**
The validator only supports POST binding.

**Impact**
As we should have no IdPs that actually use Redirect to submit SAMLResponses, changing this should have no impact.  Incoming GET-requests on `/authentication/sp/consume-assertion` all result in a 400 or 405 error anyway, and seem to be only used by bots.

kayjoosten:
EB only supports HTTP POST binding on the ACS endpoint. The route already restricts to POST, so the GET branch was dead code. GET now throws InvalidRequestMethodException like any other unsupported method.

Closes #1750

### DIFF
--- a/src/OpenConext/EngineBlock/Validator/AcsRequestValidator.php
+++ b/src/OpenConext/EngineBlock/Validator/AcsRequestValidator.php
@@ -25,30 +25,27 @@ use Symfony\Component\HttpFoundation\Request;
 /**
  * The AcsRequestValidator verifies valid saml binding
  *
- * Valid requests method / binding combinations for SSO are:
+ * Valid requests method / binding combinations for ACS are:
  *
  *  | Request method | Parameter name  |
  *  | -------------- | --------------- |
- *  | GET            | SAMLResponse    |
  *  | POST           | SAMLResponse    |
  */
 class AcsRequestValidator implements RequestValidator
 {
-    private $supportedRequestMethods = [Request::METHOD_GET, Request::METHOD_POST];
+    private $supportedRequestMethods = [Request::METHOD_POST];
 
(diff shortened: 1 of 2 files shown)


src/OpenConext/EngineBlock/Validator/AcsRequestValidator.php
```diff
@@ -25,30 +25,27 @@
...
         if (!in_array($requestMethod, $this->supportedRequestMethods)) {
-            // Only Redirect binding is supported for Single Sign On
             throw new InvalidRequestMethodException(
                 sprintf('The HTTP request method "%s" is not supported on this SAML ACS endpoint', $requestMethod)
             );
         }
 
-        if (($requestMethod ===  Request::METHOD_POST && !$request->request->has('SAMLResponse')) ||
-            ($requestMethod ===  Request::METHOD_GET && !$request->query->has('SAMLResponse'))) {
+        if ($requestMethod === Request::METHOD_POST && !$request->request->has('SAMLResponse')) {
```
johanib, line 48: I think this could be simplified to  just `!$request->request->has('SAMLResponse')` as the request method has already been checked. 

But not a blocker.

johanib (approved):

PR #2001, opened 2026-05-15
Issue-1989 apply rector rules for 8.5

### LINKED ISSUE #1989: Apply or ignore rector rules after PHP 8.5 update
As part of the PHP 8.5 ticket (or Symfony updates ticket), we should have also processed the rector rules.

We use rector to ensure the codebase stays up to date with the latest standards and consistent. For example, it enforces no mixed annotations / attributes are used, but only attributes.

## Todo
From within the container, run:
```sh
./ci/qa/rector.sh --dry-run
```

Check the output, and apply the rules that can/should be applied. Add the rules that will potentially break stuff to the ignore list.

kayjoosten:
Applies rector rules after the PHP 8.5 upgrade. 81 files updated with modern PHP idioms: typed constants, #[Override] attributes, first-class callables, array_any/array_first, and __sleep/__wakeup replaced by __serialize/__unserialize.

Two rules were added to the skip list:

DeprecatedAnnotationToDeprecatedAttributeRector was skipped because #[Deprecated] is a runtime attribute in PHP 8.5 that emits E_DEPRECATED when the annotated method is called. PHPUnit is configured with failOnRisky=true and convertDeprecationsToExceptions, so tests calling deprecated methods start throwing. The @deprecated docblock has no runtime effect and is safe to keep.

NewMethodCallWithoutParenthesesRector was skipped because PHPMD uses the pdepend parser which does not support the PHP 8.4 new ClassName()->method() syntax and crashes when it encounters it.

Four files had custom XML reconstruction logic in __wakeup that Rector could not safely migrate: AuthnRequestAnnotationDecorator, ResponseAnnotationDecorator, MockIdentityProvider, and MockServiceProvider. Rector's generated __unserialize blindly assigns array values to properties but never runs the reconstruction step, leaving $sspMessage as null. These were migrated manually with the correct reconstruction logic preserved.


johanib (approved):
I did validation on the serialize functions, and they work correctly.

PR #2003, opened 2026-05-19
Fix Successful login being logged before login was successful

### LINKED ISSUE #1742: Log successful login only on success when AM before consent feature is enabled
[This issue is imported from pivotal - Originaly created at May 8, 2020 by bstrooband](https://www.pivotaltracker.com/story/show/172735430)

When `eb.run_all_manipulations_prior_to_consent` is enabled succesfull logins are logged when entering the ACs location of EB. This is happening before consent is given. When disabled this is happening after consent. This is happening because the output filtering is happening inn the input  filtering were the logging is triggered.

This seem to be happening after `https://github.com/OpenConext/OpenConext-engineblock/pull/336` was implemented.

It\'s called here:
https://github.com/OpenConext/OpenConext-engineblock/blob/55861359ac192a13fe7043af66ed693728fc5c36/library/EngineBlock/Corto/Module/Service/AssertionConsumer.php#L141

And is logged here:
https://github.com/OpenConext/OpenConext-engineblock/blob/b23d71d8df5e8c6c5974273a3f4e5864cd0b4e38/library/EngineBlock/Corto/Filter/Output.php#L65

johanib:
Prior to this change, the logs logged `login granted` during the assertionConsumerService phase, before consent was given. This happened because the Success log was implemented as a output filter. This is no problem on its own, but, when the `run_all_manipulations_prior_to_consent` feature is enabled, the output filters are performed before consent is handled. This caused the Log line to be logged before the consent was handled.

This change removes the LogLogin output filter and moves the code to a separate log fuction, to be called from the correct step in the flow.

Fixes #1742


src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/LoggingContext.php
```diff
@@ -89,6 +89,34 @@ public function theFollowingLogMessagesShouldHaveACorrelationId(TableNode $table
         }
     }
 
+    /**
+     * @Then the log message :message should appear after log message :anchor
+     */
+    public function theLogMessageShouldAppearAfterLogMessage(string $message, string $anchor): void
```
kayjoosten: I dont see this function used. Or was this already somewhere but never defined?

src/OpenConext/EngineBlockBridge/Logger/LoginLogger.php
```diff
@@ -0,0 +1,87 @@
...
+use EngineBlock_Saml2_AuthnRequestAnnotationDecorator;
+use EngineBlock_Saml2_ResponseAnnotationDecorator;
+use EngineBlock_SamlHelper;
+use OpenConext\EngineBlock\Metadata\Entity\IdentityProvider;
+use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
+use OpenConext\EngineBlock\Metadata\MetadataRepository\MetadataRepositoryInterface;
+
+class LoginLogger
+{
+    private AuthenticationLoggerAdapter $authenticationLogger;
```
kayjoosten: Now that we use php 8+ i think property promotion would be a cleaner approach

johanib: done

src/OpenConext/EngineBlockFunctionalTestingBundle/Features/Context/LoggingContext.php
johanib: you are right?, removed

kayjoosten (approved):

PR #2004, opened 2026-05-19
remove dead eb.encrypted_assertions_require_outer_signature feature flag

### LINKED ISSUE #2002: Fix or remove `encrypted_assertions_require_outer_signature`
When a user logs in via EngineBlock, an IdP sends back a SAML response containing the user's identity. That response can be encrypted and signed.

Signing can happen in two places:
 - On the outer envelope, the whole response message
 - On the inner assertion, just the user identity claim inside

Currently, an old feature flag/parameter and a skipped test exist in the code: `eb.encrypted_assertions_require_outer_signature`

What it does: "If a response arrives with encryption AND no outer signature, should EngineBlock reject it?"
 * true, default: reject it, require an outer signature
 * false: accept it, trust the inner assertion signature instead

**History**
1. 2016 The flag was implemented: https://github.com/OpenConext/OpenConext-engineblock/commit/7f6ab9908
2. 2018 How EngineBlock processes SAML messages was refactored. In that refactor, the flag was removed from the code, because "the underlying library does not support it". But it was not removed from the config. https://github.com/OpenConext/OpenConext-engineblock/commit/6f858ec76
3. 2026: The flag exists in config, defaults to true, but the code never reads it, the feature is broken. Setting it to false has no effect.

We should either implement support for this feature flag again or remove the feature flag.

kayjoosten:
The flag has been inoperative since 2018: Bindings.php unconditionally rejects encrypted assertions without an outer signature regardless of the flag value. Wiring it up is not viable because the SAML2 library checks wasSignedAtConstruction() on the original EncryptedAssertion objects (before decryption), which always returns false, so the library itself would reject such responses too.

Removes the flag from engineblock_features.yaml, parameters.yml.dist, TestFeatureConfiguration, and cleans up the Behat scenarios that referenced it (removing the @SKIP scenario that tested the broken path).

Closes #2002

### DIFF
--- a/config/packages/engineblock_features.yaml
+++ b/config/packages/engineblock_features.yaml
@@ -7,7 +7,6 @@ parameters:
         api.deprovision:     "%feature_api_deprovision%"
         api.users_nameid_lookup: "%feature_api_users_nameid_lookup%"
         eb.encrypted_assertions: "%feature_eb_encrypted_assertions%"
-        eb.encrypted_assertions_require_outer_signature: "%feature_eb_encrypted_assertions_require_outer_signature%"
         eb.run_all_manipulations_prior_to_consent: "%feature_run_all_manipulations_prior_to_consent%"
         eb.block_user_on_violation: "%feature_block_user_on_violation%"
         eb.enable_sso_notification: "%feature_enable_sso_notification%"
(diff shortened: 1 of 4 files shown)


johanib:
Don't merge before Bas has approved its removal.

baszoetekouw:
Yes, please merge!

kayjoosten:
@baszoetekouw approved

PR #2016, opened 2026-06-02
feat(db): add unique constraint on user.uuid

### LINKED ISSUE #1974: Missing unique index on user.uuid
The user table has a uuid column (char(36) NOT NULL, Doctrine type engineblock_collab_person_uuid) that stores a unique identifier per user (collabPersonUuid). Currently only a non-unique index idx_user_uuid exists on this column there is no unique constraint enforced at the database level.

I noticed this because queries like "SELECT WHERE uuid = X LIMIT 1", which raises flags on a non-unique column.

I do not thing this is causing any issues or is needed, just preventing theoretical issues. 
Let's briefly discuss.

kayjoosten:
The uuid column stores collabPersonUuid, which must be unique per user. The existing non-unique idx_user_uuid provided no enforcement.

Migration uses ALGORITHM=INPLACE, LOCK=NONE for zero-downtime on large tables. Skips if idx_user_uuid is absent (already migrated manually).

DONT release before the dbal 4 update

Closes #1974

### DIFF
--- a/migrations/DoctrineMigrations/Version20260602000000.php
+++ b/migrations/DoctrineMigrations/Version20260602000000.php
@@ -31,11 +31,11 @@ final class Version20260602000000 extends AbstractEngineBlockMigration
 
     public function up(Schema $schema): void
     {
-        $this->addSql('ALTER TABLE `user` CHANGE `uuid` `uuid` CHAR(36) NOT NULL COMMENT \'(DC2Type:engineblock_collab_person_uuid)\', ADD UNIQUE INDEX `uq_user_uuid` (`uuid`), DROP INDEX `idx_user_uuid`');
+        $this->addSql('ALTER TABLE `user` CHANGE `uuid` `uuid` CHAR(36) NOT NULL, ADD UNIQUE INDEX `uq_user_uuid` (`uuid`), DROP INDEX `idx_user_uuid`');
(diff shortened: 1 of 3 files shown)


johanib (approved):

migrations/DoctrineMigrations/Version20260602000000.php
```diff
@@ -0,0 +1,41 @@
...
+final class Version20260602000000 extends AbstractEngineBlockMigration
+{
+    public function getDescription(): string
+    {
+        return 'Replace non-unique idx_user_uuid with unique index uq_user_uuid on user.uuid';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE `user` ADD UNIQUE INDEX `uq_user_uuid` (`uuid`), DROP INDEX `idx_user_uuid`');
```
johanib: Tested: When you run this, and then run `./bin/console doc:mig:diff`, doctrine wants to apply: `CHANGE uuid uuid CHAR(36) NOT NULL`.

I think, adding a uniqueconstraint causes doctrine to make the column not null.

We can explicitly set the column to be nullable or add the NOT NULL to the query and the changelog.

johanib (approved):

PR #2029, opened 2026-06-19
Fix logos in WAYF not scaling properly (closes: #2028)

### LINKED ISSUE #2028: Logos in WAYF should be square en not be squashed
The CSS in the WAYF is forcing square logo's to be squashed vertically.  

The logos are rendered like this:
```html
<div aria-describedby="idp__titleremaining1" class="wayf__idp" data-connectable="true" data-index="1" data-count="0" tabindex="0" data-entityid="http://idp.example.com/" data-title="Example IdP" data-idpkey="http://idp.example.com/|2ad3124767dbf57b845ea2897febab8b342e2a8a067b179a33c76962cc35def3a4">
  <div class="idp__logo">
    <img src="https://static.surfconext.nl/media/idp/some_logo.png" alt="" loading="lazy">
  </div>
  <div class="idp__content">
    <!-- [removed for clarity -->
  </div>
</div>
```
The size is configured on the outer `<div class="idp__logo">` (rather than on the  itself) like this:

Image

There are two issues here:
 - the width and height are not equal
 - if an image does not fit, is is squashed in on direction rather than scaled in two directions (keeping the aspect ratio).

This causes square logos (which we are moving towards in SURFconext) to be always squashed.

The effect is subtile but noticable.  Compare current logo rendering (left) to correct logo (right), which was created by setting the `<div>` height to `5.67rem`.
Image

If possible, I'd prefer changing the logo height without changing the height of the entire IdP-button (so the screenshot above with the higher button is **not** the desired solution).

baszoetekouw:
Small fix make logos in the WAYF list scale properly

### DIFF
--- a/theme/base/stylesheets/pages/wayf/idpList.scss
+++ b/theme/base/stylesheets/pages/wayf/idpList.scss
@@ -44,10 +44,10 @@
             > .idp__logo {
                 align-self: center;
                 display: flex;
-                height: 5.05rem;
-                width: 5.67rem;
+                height: 5.15rem;
+                width: 5.15rem;
                 justify-self: center;
-                padding: .5rem;
+                padding: .3rem;
 
                 @include breakpoints.screen('mobile') {
                     margin: 0;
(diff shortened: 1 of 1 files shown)


kayjoosten (approved):
LGTM

PR #2034, opened 2026-06-26
feat(stepup): add service display name to Stepup callout AuthnRequest

### LINKED ISSUE #2011: Add display name to the Extenstions of SAML request
**Add DisplayName to the `<samlp:Extensions>` of the Stepup callout AuthnRequest**

**User story**
As a user, I want to see the name of the service I am logging into during second-factor authentication, so that I can verify I am authenticating for the right application and detect unexpected or malicious authentication attempts.

**Background**
This is the EB side of a broader initiative to show the service name during Stepup authentication. See the RFC for the full design: OpenConext/Stepup-Gateway#587.

EB acts as an SFO application connecting to Stepup-Gateway on behalf of thousands of services. Rather than showing a generic proxy name (e.g. "OpenConext"), we want to pass the actual service name to Stepup so it can be shown in the 2FA UI, push notifications, and SMS messages.

EB already knows the service name via the `name:en` and `name:nl` fields of the SP entity (`saml20_sp`) or RP entity (`oidc10_rp`) from OpenConext-manage, pushed to EB via the metadata push API.

**What needs to be done**

Introduce a feature flag `feature_stepup_send_service_name` in EB (following the existing feature flag pattern in `config/packages/parameters.yml.dist`).

When enabled, for every Stepup callout AuthnRequest, EB MUST add an `mdui:UIInfo` element to the `<samlp:Extensions>` containing a `mdui:DisplayName` for each available language. Example:

```xml
<samlp:Extensions>
  <mdui:UIInfo>
    <mdui:DisplayName xml:lang="nl">Electronische leeromgeving</mdui:DisplayName>
  </mdui:UIInfo>
</samlp:Extensions>
```

**Acceptance criteria**
- [ ] A feature flag `feature_stepup_send_service_name` is introduced, disabled by default
- [ ] When enabled, the `mdui:UIInfo` element with `mdui:DisplayName` entries for `en` and `nl` is added to the `<samlp:Extensions>` of the Stepup callout AuthnRequest
- [ ] The display names are sourced from `name:en` and `name:nl` or any other locale of the SP/RP entity from OpenConext-manage
- [ ] If a language's display name is not available, that `mdui:DisplayName` element is omitted (no empty values)
- [ ] When the feature flag is disabled, the AuthnRequest is unchanged from current behavior

**Related**
- RFC: OpenConext/Stepup-Gateway#587

kayjoosten:
## Summary

- Adds `mdui:UIInfo`/`mdui:DisplayName` to `<samlp:Extensions>` of the Stepup Gateway SFO callout AuthnRequest, so Stepup can surface the correct service name in its 2FA UI instead of the generic proxy name
- Controlled by feature flag `eb.stepup.send_service_name` (disabled by default)
- Uses the user's current engine locale, falling back to English; prefers mdui DisplayName over flat `name*` fields
- Closes #2011

## Technical notes

- New `StepupServiceNameExtension` class in `src/OpenConext/EngineBlock/Stepup/` follows the same pattern as `StepupGsspUserAttributeExtension`
- `xml:lang` attribute set via `setAttributeNS` (XML namespace) so XPath and SAML consumers resolve it correctly
- Locale resolved via existing `LocaleProvider::getLocale()` available through `DiContainer`

## Test plan

- [x] 7/7 unit tests pass (`StepupServiceNameExtensionTest`)
- [x] 27/27 Behat scenarios pass in `Stepup.feature` (includes 2 new scenarios: feature enabled / disabled)
- [x] `phpcs` and `docheader` clean on new files
- [x] Feature flag disabled by default — existing Stepup flow unchanged when flag is off

### DIFF
--- a/config/packages/parameters.yml.dist
+++ b/config/packages/parameters.yml.dist
@@ -235,6 +235,7 @@ parameters:
     feature_enable_idp_initiated_flow: true
     feature_stepup_sfo_override_engine_entityid: false
     feature_stepup_send_user_attributes: false
+    feature_stepup_send_service_name: false
     feature_enable_sram_interrupt: false
     feature_hide_bookmarkable_url: false
 
--- a/library/EngineBlock/Corto/ProxyServer.php
+++ b/library/EngineBlock/Corto/ProxyServer.php
@@ -27,6 +27,7 @@ use OpenConext\EngineBlock\Metadata\MfaEntity;
 use OpenConext\EngineBlock\Metadata\Service;
 use OpenConext\EngineBlock\Metadata\TransparentMfaEntity;
 use OpenConext\EngineBlock\Stepup\StepupGsspUserAttributeExtension;
+use OpenConext\EngineBlock\Stepup\StepupServiceNameExtension;
 use OpenConext\EngineBlock\Metadata\X509\KeyPairFactory;
 use OpenConext\EngineBlockBundle\Authentication\AuthenticationState;
 use OpenConext\EngineBlockBundle\Exception\UnknownKeyIdException;
@@ -486,7 +487,8 @@ class EngineBlock_Corto_ProxyServer
         Loa $authnContextClassRef,
         NameID $nameId,
         bool $isForceAuthn,
-        Assertion $originalAssertion
+        Assertion $originalAssertion,
+        ?ServiceProvider $sp = null,
     ): void {
         $ebRequest = EngineBlock_Saml2_AuthnRequestFactory::createFromRequest(
             $spRequest,
@@ -555,6 +557,13 @@ class EngineBlock_Corto_ProxyServer
             }
         }
 
+        $isSendServiceNameConfigured = $features->hasFeature('eb.stepup.send_service_name');
+        $isSendServiceNameEnabled = $features->isEnabled('eb.stepup.send_service_name');
+
+        if ($isSendServiceNameConfigured && $isSendServiceNameEnabled && $sp !== null) {
+            $locale = $container->getLocaleProvider()->getLocale();
+            StepupServiceNameExtension::add($sspMessage, $sp, $locale);
+        }
 
         // Link with the original Request
         $diContainerRuntime = EngineBlock_ApplicationSingleton::getInstance()->getDiContainerRuntime();
@@ -635,6 +644,7 @@ class EngineBlock_Corto_ProxyServer
             $nameId,
             $sp->getCoins()->isStepupForceAuthn(),
             $originalAssertion,
+            $sp,
         );
     }
 
--- a/library/EngineBlock/Corto/ProxyServer.php
+++ b/library/EngineBlock/Corto/ProxyServer.php
@@ -557,10 +557,11 @@ class EngineBlock_Corto_ProxyServer
             }
         }
 
-        $isSendServiceNameConfigured = $features->hasFeature('eb.stepup.send_service_name');
-        $isSendServiceNameEnabled = $features->isEnabled('eb.stepup.send_service_name');
-
-        if ($isSendServiceNameConfigured && $isSendServiceNameEnabled && $sp !== null) {
+        if (
+            $features->hasFeature('eb.stepup.send_service_name')
+            && $features->isEnabled('eb.stepup.send_service_name')
+            && $sp !== null
+        ) {
             $localeProvider = $container->getLocaleProvider();
             StepupServiceNameExtension::add(
                 $sspMessage,
--- a/library/EngineBlock/Corto/ProxyServer.php
+++ b/library/EngineBlock/Corto/ProxyServer.php
@@ -561,8 +561,14 @@ class EngineBlock_Corto_ProxyServer
         $isSendServiceNameEnabled = $features->isEnabled('eb.stepup.send_service_name');
 
         if ($isSendServiceNameConfigured && $isSendServiceNameEnabled && $sp !== null) {
-            $locale = $container->getLocaleProvider()->getLocale();
-            StepupServiceNameExtension::add($sspMessage, $sp, $locale);
+            $localeProvider = $container->getLocaleProvider();
+            StepupServiceNameExtension::add(
+                $sspMessage,
+                $sp,
+                $localeProvider->getLocale(),
+                $localeProvider->getDefaultLocale(),
+                $localeProvider->getAvailableLocales(),
+            );
         }
 
         // Link with the original Request
--- a/library/EngineBlock/Corto/ProxyServer.php
+++ b/library/EngineBlock/Corto/ProxyServer.php
@@ -488,7 +488,7 @@ class EngineBlock_Corto_ProxyServer
         NameID $nameId,
         bool $isForceAuthn,
         Assertion $originalAssertion,
-        ?ServiceProvider $sp = null,
+        ServiceProvider $sp,
     ): void {
         $ebRequest = EngineBlock_Saml2_AuthnRequestFactory::createFromRequest(
             $spRequest,
@@ -557,10 +557,8 @@ class EngineBlock_Corto_ProxyServer
             }
         }
 
-        if (
-            $features->hasFeature('eb.stepup.send_service_name')
+        if ($features->hasFeature('eb.stepup.send_service_name')
             && $features->isEnabled('eb.stepup.send_service_name')
-            && $sp !== null
         ) {
             $localeProvider = $container->getLocaleProvider();
             StepupServiceNameExtension::add(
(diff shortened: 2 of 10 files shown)


library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -555,6 +557,13 @@ public function sendStepupAuthenticationRequest(
             }
         }
 
+        $isSendServiceNameConfigured = $features->hasFeature('eb.stepup.send_service_name');
```
johanib: huh, why not only rely on `isSendServiceNameEnabled`?

library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -486,7 +487,8 @@ public function sendStepupAuthenticationRequest(
         Loa $authnContextClassRef,
         NameID $nameId,
         bool $isForceAuthn,
-        Assertion $originalAssertion
+        Assertion $originalAssertion,
+        ?ServiceProvider $sp = null,
```
johanib: Instead of passing the ServiceProvider down the line, consider passing the `string $serviceName`.

This makes the code more flexible and testable, as you do not need a instantiated $sp here. This also removes the `$container` access for the locale.

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
```diff
@@ -0,0 +1,73 @@
...
+
+use OpenConext\EngineBlock\Metadata\Entity\ServiceProvider;
+use SAML2\DOMDocumentFactory;
+use SAML2\Message;
+use SAML2\XML\Chunk;
+
+final class StepupServiceNameExtension
+{
+    private const MDUI_NS = 'urn:oasis:names:tc:SAML:metadata:ui';
+    private const SUPPORTED_LOCALES = ['en', 'nl', 'pt'];
```
johanib: Instead of hardcoding here, read symfony values or values?

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
```diff
@@ -0,0 +1,73 @@
...
+
+    public static function add(Message $message, ServiceProvider $sp, string $locale): void
+    {
+        if (!in_array($locale, self::SUPPORTED_LOCALES, true)) {
+            $locale = 'en';
```
johanib: fallback to global/symfony default?

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
```diff
@@ -0,0 +1,73 @@
...
+        }
+        $result = self::resolveName($sp, $locale);
+        if ($result === null && $locale !== 'en') {
```
johanib: This assumes the default `locale === 'en'`. Would be better to do `$locale !== $defaultLocale`

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $dom = DOMDocumentFactory::create();
+        $uiInfo = $dom->createElementNS(self::MDUI_NS, 'mdui:UIInfo');
+        $displayName = $dom->createElementNS(self::MDUI_NS, 'mdui:DisplayName');
+        $displayName->setAttributeNS('http://www.w3.org/XML/1998/namespace', 'xml:lang', $resolvedLocale);
+        $displayName->textContent = $name;
+        $uiInfo->appendChild($displayName);
+
+        $ext = $message->getExtensions();
+        $ext['mdui:UIInfo'] = new Chunk($uiInfo);
```
johanib, line 62: What if `$ext` already contains `mdui:UIInfo`. 
I would guess this is never the case, and this is correct, so no change warranted. 

But good to double check.

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
```diff
@@ -0,0 +1,73 @@
...
+        $message->setExtensions($ext);
+    }
+
+    /**
+     * @return array{string, string}|null
+     */
+    private static function resolveName(ServiceProvider $sp, string $locale): ?array
```
johanib, line 69: This function is duplicated business logic?

config/packages/parameters.yml.dist
```diff
@@ -235,6 +235,7 @@ parameters:
     feature_enable_idp_initiated_flow: true
     feature_stepup_sfo_override_engine_entityid: false
     feature_stepup_send_user_attributes: false
+    feature_stepup_send_service_name: false
```
johanib, line 238: add this to changelog.md

johanib (changes requested):

library/EngineBlock/Corto/ProxyServer.php
kayjoosten: isEnabled() throws LogicException if the feature isn't registered. Same pattern as the eb.stepup.sfo.override_engine_entityid check right above it, so the hasFeature guard is required.

library/EngineBlock/Corto/ProxyServer.php
kayjoosten: Considered it, keeping $sp avoids duplicating the mdui/name-fallback resolution at the call site; StepupServiceNameExtension::resolveName() now reuses the same nameEn/nameNl/namePt fields via an explicit match instead of a dynamic property read, and locale defaults now come from LocaleProvider instead of being hardcoded. Open to the string-passing refactor separately if you still want it.

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
kayjoosten: i checked it, $message->getExtensions() at this call site is always empty for a freshly-built Stepup AuthnRequest, so this is safe as-is.

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
johanib: Can't happen in real scenarios

src/OpenConext/EngineBlock/Stepup/StepupServiceNameExtension.php
johanib: can you check wayf?

johanib: discussed: OK to leave it as is. We also checked the `localizedName` function below. This is in line with other usages.

library/EngineBlock/Corto/ProxyServer.php
```diff
@@ -555,6 +557,19 @@ public function sendStepupAuthenticationRequest(
             }
         }
 
+        $isSendServiceNameConfigured = $features->hasFeature('eb.stepup.send_service_name');
+        $isSendServiceNameEnabled = $features->isEnabled('eb.stepup.send_service_name');
```
johanib: move to inline if, so it doesnt get called if feature flag is not set

johanib (approved):
Maybe check dots on i, looks good, no regression 👍

johanib:
and check commit message

johanib (changes requested):
On second thought, we discussed applying the constraints on the servicename from the RFC at the source. So, here in EB. 

Can we add the stepup bundle that supplies the sanitization logic so EB can sanitize it beforehand?

kayjoosten:
> On second thought, we discussed applying the constraints on the servicename from the RFC at the source. So, here in EB.
> 
> Can we add the stepup bundle that supplies the sanitization logic so EB can sanitize it beforehand?

Tbh i dont see the benefit, it gets sanitized or overwritten in gateway either way. I know this is something we said in the standup but i think to leave it up to gateway as there the final value will be send out. What do you think about that?

johanib:
> > On second thought, we discussed applying the constraints on the servicename from the RFC at the source. So, here in EB.
> > Can we add the stepup bundle that supplies the sanitization logic so EB can sanitize it beforehand?
> 
> Tbh i dont see the benefit, it gets sanitized or overwritten in gateway either way. I know this is something we said in the standup but i think to leave it up to gateway as there the final value will be send out. What do you think about that?

If these were official saml specs, we should apply it here. But, as long as we keep within the saml extension fields limit, it should be ok. 

Applying this only in gateway seems sensible. 
Maybe a good check here in EB would be, what happens if a service name is 2000 chars long? It should probably cut it at a sensible length, like 1000 chars?
Otoh, might as well apply the same sanitization as in stepup, if needed, just add the function to eb.

PR #2037, opened 2026-06-30
7.x: Fix console on bare-metal installs

tvdijen:
Open for discussion, because I don't know if this is the right solution, but we ran into the issue that the .env file was not loaded automatically when running `bin/console` commands. We run multiple instances of EB on the same machine, so it's crucial that we can set the environment variables (especially APP_SECRET) per instance instead of setting it globally.  This change seems to fix that.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -8,7 +8,7 @@ use Symfony\Component\ErrorHandler\Debug;
 
 set_time_limit(0);
 
-require dirname(__DIR__) . '/vendor/autoload.php';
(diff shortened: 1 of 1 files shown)


johanib (approved):
Thanks. I took a look, this seems a fine fix 👍  It's narrowly scoped, so should not break anything.

As a sidenote, I noticed a bug in the dotenv not loading `.env` override files, for example `.env.prod` is ignored. Made a separate pr for that: https://github.com/OpenConext/OpenConext-engineblock/pull/2042

PR #2046, opened 2026-07-07
fix(logging): filter Doctrine ORM deprecation noise

### LINKED ISSUE #2043: Deprecation warning
Engine (7.2-beta) gives a lot of these warnings:

```json
{"channel":"php","level":"INFO","message":"User Deprecated: Class \"Doctrine\\ORM\\Proxy\\Autoloader\" is deprecated. Use native lazy objects instead. (Autoloader.php:74 called by DoctrineBundle.php:136, https://github.com/doctrine/orm/pull/12005, package doctrine/orm)","context":{"exception":{}},"extra":{"request_id":"6a4b91420cf36","correlation_id":null,"session_id":null}}
```

kayjoosten:
## Summary
- Engine floods `application.log` with `"User Deprecated: Class \"Doctrine\\ORM\\Proxy\\Autoloader\" is deprecated ..."` on every request — triggered by `DoctrineBundle::boot()` registering the legacy proxy autoloader. Not actionable from our code; the deprecation lives inside DoctrineBundle's own boot code.
- Fix: call `Deprecation::ignoreDeprecations($link)` (from `doctrine/deprecations`, the library's own purpose-built suppression API) once in `Kernel::registerBundles()` — the earliest point common to both the HTTP and console entrypoints, before `DoctrineBundle::boot()` triggers the deprecation.
- 1-line change in `src/Kernel.php`. No monolog config, no custom handler.

## Test plan
- [x] Reproduced on unpatched `main`: 170 `"User Deprecated"` lines over one `behat --suite default` run
- [x] With this fix: 0 deprecation lines, other log entries (2048 lines) unaffected — checked in both `ci` and `dev` envs
- [x] `phpmd`, `phpcs` (src + legacy), `docheader` — all clean
- [x] `phpunit` unit (957), eb4 (221), integration (103), functional (119) — all green
- [x] `behat` default suite — 297/297 scenarios, 5531/5531 steps green

Fixes #2043

### DIFF
--- a/src/Kernel.php
+++ b/src/Kernel.php
@@ -18,6 +18,7 @@
 
 namespace App;
 
+use Doctrine\Deprecations\Deprecation;
 use Override;
 use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
 use Symfony\Component\Config\Loader\LoaderInterface;
@@ -35,6 +36,8 @@ class Kernel extends BaseKernel
 
     public function registerBundles(): iterable
     {
+        Deprecation::ignoreDeprecations('https://github.com/doctrine/orm/pull/12005');
+
         $contents = require $this->getProjectDir().'/config/bundles.php';
         foreach ($contents as $class => $envs) {
             if ($envs[$this->environment] ?? $envs['all'] ?? false) {


johanib (commented):
Was this issue reproducible, did you validate it works?
EB 'errors' often travel via \EngineBlock_Application_ErrorHandler::error, which takes all messages, like deprecations, and re-throws / converts them to errors. 

So I thought it might be possible to do the filtering there, instead of afterwards via monolog.

baszoetekouw:
> So I thought it might be possible to do the filtering there, instead of afterwards via monolog.

The upside of doing it like this is that's it's easy to disable/enable from configuration.  If we would filter inside engineblock, we would at least need a config parameter to toggle the filtering (because we do not want the filtering in dev/test, I think).

kayjoosten:
This should do the trick. Doctrine has something for it.

johanib:
Change looks good :+1: Please merge after cleaning up the pr description, as it no longer relates to the actual fix
- [x] PR cleaned up
- [x] Commit message cleaned up

johanib (changes requested):

johanib (approved):

Issue #1935, opened 2026-03-11
Check incoming SAML Requests for old IssueDates


baszoetekouw:
# New Feature Request

## Summary

users sometimes bookmark/save a url somewhere in the middle of an authentication flow.  This is an easy mistake to make for regular user, in particular for Services that redirect the users right away, without requiring a press on a login button.

The reason that this sometimes does work is that the WAYF screen includes the full SAMLRequest in the URL, e.g., https://engine.dev.openconext.org/authentication/idp/single-sign-on/key:20230403?SAMLRequest=jVLJTsMwFPyVyPfEiVMCtZpKhR6oxFLRwoELcsyjtXCeg5%2FD8ve4LUuREOJqz6KZeSNSre3kpA9rvIKnHigkr61FktuPmvUepVNkSKJqgWTQcjE5P5Miy2XnXXDaWZZMiMAH4%2FDEIfUt%2BAX4Z6Ph%2BuqsZusQOpKctwrVCrIQPURGvX%2FQDuE1ZGj5Ym2axlkI64zI8Y2D4PPLxZIl0wg3qDbi31KAK4O%2FS6kYBTAYvaVwc99xMriykJJZYRqfHuFNilyU%2BSAvWTKb1uyuGqih0g0c6hyGZSGgGWg4qIqqVMPhUSEijKiHGVJQGGoW6VWal2lRLPMjeSCkqG5ZMv%2Fo49jgfbT8u7xmByJ5ulzO013WG%2FC0zRkBbDzaTCC3xn5vlL9l1ecSbPyP3umr9xHfc9tZd%2FIiys%2Bmc2eNfksm1rqXEw8qQM0Kxsc7ys%2FjGb8D&RelayState=ss%3Amem%3Abc586a8a0abe67b673ed2899a2de3653f170bad31acc186e10091ef258be7609
For Engineblock, this is totally fine, and it will let the user log in correctly, but some SPs (in particular Shib) refuse to process these unsolicited messages and show confusing error messages.

It would be nice to either prevent this, or catch this error in another way and show a helpful error message, because not all Services support this, and this often leads to hard-to-debug issues.  

Easy way to reproduce this is by using Manage:
 - go to manage, save the URL in the WAYF
 - remove the manage cookies in your browser
 - try to use your saved url to login to manage.
 - Manage will show an `opensaml::SecurityPolicyException`

## Proposed Solution

Implement the following mitigations:
[...]

PR #2061, opened 2026-07-20
Log warning for SAMLRequests with IssueInstant older than 24 hours

### LINKED ISSUE #1972: Check for and log old IssueInstants

kayjoosten:
## Summary
- Adds a distinct warning log when an incoming SAMLRequest's IssueInstant is more than 24h in the past, per #1972 (follow-up to #1935)
- Existing generic clock-skew notice (>30s, either direction) is preserved for all other cases; the two never fire together for the same message — the 24h check takes priority
- SAMLRequests only (SP), Response/IdP handling unchanged; no requests are blocked

## Test plan
- [x] New unit tests in BindingsTest.php: old-SP warning, recent-SP notice, old-IdP still generic notice, fresh-SP no log
- [x] Full eb4 PHPUnit suite (225/225)
- [x] phpcs-legacy, docheader clean

Closes #1972

### DIFF
--- a/library/EngineBlock/Corto/Module/Bindings.php
+++ b/library/EngineBlock/Corto/Module/Bindings.php
@@ -49,6 +49,8 @@ class EngineBlock_Corto_Module_Bindings extends EngineBlock_Corto_Module_Abstrac
 
     const SAML_STATUS_MESSAGE_EMPTY = '(No message provided)';
 
+    const OLD_ISSUE_INSTANT_THRESHOLD_SECONDS = 86400; // 24 hours
+
     protected static $ASSERTION_SEQUENCE = array(
         'saml:Issuer',
         'ds:Signature',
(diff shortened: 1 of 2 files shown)


johanib (approved):
LGTM! 

Nitpick: 
The commit message contains implementation details.


```
How does it address the issue?                                        
This change adds a dedicated warning for severely stale AuthnRequests,
taking priority over the existing minor-drift notice so only the      
single most-applicable message fires per request. The check is scoped 
to AuthnRequests only, since Responses are handled differently and    
weren't part of the reported issue. This is logging only, no         
requests are blocked.                                                 
```

kayjoosten:
Only changed the commit message, doesnt need rereview

PR #2069, opened 2026-07-21
Add some attributes to authz_eb call

mrvanes:
Adds external Subject-id and email to the EB SRAM interrupt call

### DIFF
--- a/library/EngineBlock/Corto/Filter/Command/SramInterruptFilter.php
+++ b/library/EngineBlock/Corto/Filter/Command/SramInterruptFilter.php
@@ -107,6 +107,8 @@ class EngineBlock_Corto_Filter_Command_SramInterruptFilter extends EngineBlock_C
 
         $userId = $this->_collabPersonId ?? "";
         $eppn = $attributes['urn:mace:dir:attribute-def:eduPersonPrincipalName'][0] ?? "";
+        $externalSubject = $attributes['urn:oasis:names:tc:SAML:attribute:subject-id'][0] ?? "";
+        $email = $attributes['urn:mace:dir:attribute-def:mail'][0] ?? "";
         $continueUrl = $this->_server->getUrl('SramInterruptService', '') . "?ID=$id";
         $serviceId = $serviceProvider->entityId;
         $issuerId = $this->_identityProvider->entityId;
@@ -117,7 +119,9 @@ class EngineBlock_Corto_Filter_Command_SramInterruptFilter extends EngineBlock_C
             $eppn,
             $continueUrl,
             $serviceId,
-            $issuerId
+            $issuerId,
+            $externalSubject,
+            $email
         );
     }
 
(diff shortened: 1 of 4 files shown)


baszoetekouw:
Wouldn't it be more future proof to simply send all attributes over to SBS?  That's what we do for the AA and PDP, too.

library/EngineBlock/Corto/Filter/Command/SramInterruptFilter.php
```diff
@@ -107,6 +107,8 @@ private function buildRequest(ServiceProvider $serviceProvider): AuthzRequest
 
         $userId = $this->_collabPersonId ?? "";
         $eppn = $attributes['urn:mace:dir:attribute-def:eduPersonPrincipalName'][0] ?? "";
+        $externalSubject = $attributes['urn:oasis:names:tc:SAML:attribute:subject-id'][0] ?? "";
```
baszoetekouw: Please don't use subject-id here.  It is reserved for future use in Engine en you cannot use it to transport other data.

library/EngineBlock/Corto/Filter/Command/SramInterruptFilter.php
```diff
@@ -107,6 +107,8 @@ private function buildRequest(ServiceProvider $serviceProvider): AuthzRequest
...
+        $email = $attributes['urn:mace:dir:attribute-def:mail'][0] ?? "";
```
baszoetekouw: email is multi-valued

library/EngineBlock/Corto/Filter/Command/SramInterruptFilter.php
baszoetekouw: as discussed, this is the regular subjectid that are phasing in in the platform.

baszoetekouw: I wonder about the attribute name though.  Annoyingly `$subjectId` is used for something different internally so I see why you don't want to use that.  Maybe call it `$externalSubjectId` then to keep closer to the real name?

baszoetekouw:
Kun je nog een Changelog entry toevoegen?

baszoetekouw (changes requested):

library/EngineBlock/Corto/Filter/Command/SramInterruptFilter.php
```diff
@@ -107,17 +107,21 @@ private function buildRequest(ServiceProvider $serviceProvider): AuthzRequest
...
-            $serviceId,
-            $issuerId
+            userId: $userId,
+            eduPersonPrincipalName: $eppn,
+            externalSubjectId: $externalSubjectId,
+            email: $email,
+            continueUrl: $continueUrl,
+            serviceId: $serviceId,
+            issuerId: $issuerId,
+            attributes: $attributes
```
johanib: Passing the whole $attributes object as is seems weird. I think we should at least normalize it. Serializing it as is can lead to inconsistent responses in the case of attributes with Nested XML child elements or with eduPersonTargetedID which uses a NameID object.

So probably good to filter through normalize function:
```php
// snip, removed, see comment below                                                                           
```

This will potentially drop data from the attribute response, but if that is a problem we should come up with a proper api contract, not just throw in $attributes and hope for the best.

johanib: Checked how PDP and AA handle this:
PDP does not filter. AA filters any non-string value. So for consistency, we should mirror AA's approach: 
`$attributes = $this->filterNonStringValuesFromAttributes($this->getResponseAttributes());`
```php
// Filter the non string valued attributes                               
$attributes = self::filterNonStringValuesFromAttributes($attributes);    
                                                                         
private static function filterNonStringValuesFromAttributes($attributes) 
{                                                                        
    return array_filter($attributes, function ($attributeValues) {       
        foreach ($attributeValues as $attributeValue) {                  
            if (!is_string($attributeValue)) {                           
                return false;   // drop the whole attribute              
            }                                                            
        }                                                                
        return true;                                                     
    });                                                                  
}                                                                        

```

baszoetekouw (approved):

johanib (approved):
👍 Please squash when merging

PR #2083, opened 2026-08-18
Feature/postgres support

Stephan-Kok:
## PostgreSQL support

This PR is a follow-up to [[PR #2082](https://github.com/OpenConext/OpenConext-engineblock/pull/2082)](https://github.com/OpenConext/OpenConext-engineblock/pull/2082).

In this PR, we remove the `sso_provider_roles_eb5` table and introduce PostgreSQL support.

### PostgreSQL documentation

I have added a document with additional information about how to use the PostgreSQL database and how the integration works:

https://github.com/OpenConext/OpenConext-engineblock/blob/feature/postgres_support/postgres_support.md

The document also describes some additional changes that are required in the EngineBlock container and OpenConext-devconf to support PostgreSQL natively.

Additionally, I think we should discuss whether we can add a CI/CD test using a PostgreSQL database to ensure that PostgreSQL support continues to work going forward.

### Consent

One aspect I would like you to look into more deeply is the Consent functionality.

We currently do not use the Consent feature, but PostgreSQL does not support the SQL value `0000-00-00 00:00:00`, which is used in a hardcoded SQL query.

I have attempted to address this by replacing it with the epoch date `1970-01-01 00:00:00`. Please see `DbalConsentRepository.php` in this PR.

During my initial investigation, it appeared that `deleted_at` may have been incorrectly configured as a primary key, which is the reason this workaround seems to be necessary.

If this is not intentional, it would be preferable to remove this requirement and solve the underlying issue rather than keeping my current workaround. Please take a closer look at this part of the PR and let me know if the `deleted_at` primary key is intentional.


Stephan-Kok:
I have adjusted the CI scripts to contains tests using the native behat tests, but with a postgresql database. I have used the environment 'postgres' for this use case.

Issue #2085, opened 2026-08-20
Add endpoint to remove "remember my choice" cookie


baszoetekouw:
In https://github.com/OpenConext/OpenConext-engineblock/issues/1956 we describe a feature where a user can specify per SP which IdP they want to use. On a next login to that SP, the preferred IdP will be used automatically without user interaction.

This ticket describes an endpoint that allows users to reset these choices.

- Add a global configuration option `wayf.reset_choice_per_idp_redirect` to `parameters.yml` that contains a url to redirect the user to after resetting the choice (see below)

- Implement a front-end page/endpoint as follows:
  - It is located at `https://engineblock.dev.openconext.org/reset-remember-wayf`
  - When the user visits this page, check if they send  a cookie `rememberedidps` and if so, remove it.
  - Log a message with level INFO and message "WAYF-remember-my-choice cookie removed (had `<number>` entries)"
  - After removing the cookie, redirect the user (302) to the url set in `wayf.reset_choice_per_idp_redirect`

pmeulen:
- engine does a redirect. No UI in engine. return status as get parameter in the redirect.
- UI in Profile. @baszoetekouw UI design in Profile.
- Redirect URL in engine and profile config

baszoetekouw:
Return state to profile either one of:
 - cookies have been remove
 - there were no cookies

Issue #2066, opened 2026-07-20
Redesign "Remember my choice" checkbox in WAYF


baszoetekouw:
To support the new "Remember my choice per SP" feature of #1956, we want ot make adjustment to the UI of the WAYF screen.

For reference the current layout when `wayf.remember_choice` is enabled:

<img width="640" alt="Image" src="https://github.com/user-attachments/assets/0011a308-d320-42a0-a967-614172c4f01d" />

We would like to make small adjustments.

in the case that we use a per-sp "remember my choice" (both `wayf.remember_choice` and `wayf.remember_choice_per_idp` enabled):

<img width="640" alt="Image" src="https://github.com/user-attachments/assets/33b8a6c8-6503-42d6-b852-ec488412f8a7" />

The following changes have been made:
- Adjust the margins, switch the order of the checkbox and the text, adjust the text colour and weight, and add a tooltip ⓘ.  
- The mockup above uses this html (but tooltip hover doesn't work):
     ```html
     <form method="post" action="https://engine.test2.surfconext.nl/authentication/idp/process-wayf" class="wayf__rememberChoice">
         <input type="checkbox" name="rememberChoice" id="rememberChoice">
         <label for="rememberChoice">Remember my choice for this service for 90 days</label>
         <label class="tooltip" for="rememberChoice" style="margin-left: .5em;alignment-baseline: ideographic;">
             <span class="visually-hidden">Tooltip text here</span>
         </label>
     </form>
     ```
- Tooltip should show this text:
  - NL: "Als je deze optie aanzet, hoef je de komende 90 dagen voor deze dienst niet meer te kiezen hoe je wilt inloggen. Het inlogscherm wordt dan overgeslagen. **Let op**: gebruik je soms een ander account om in te loggen? Zet de optie dan uit. Je kunt de keuze later weer terugdraaien op [profile.surfconext.nl](http://profile.surfconext.nl/)."
  - EN: "If you enable this option, you won’t need to choose how you want to log into this service for the next 90 days. The login screen will then be skipped. **Please note**: do you sometimes use a different account to log in? If so, please disable this option. You can reverse this setting later at [profile.surfconext.nl](http://profile.surfconext.nl/)."
[...]

Issue #2065, opened 2026-07-20
Remember choice for WAYF per SP


baszoetekouw:
It is already possible to show a checkbox in the WAYF that allows users to save this IdP-choice for future logins.  If they select this, then future WAYFs in which their preferred IdP is present are skipped and the user is redirected to the preferred IdP without interaction.

We now want to able to specifiy this behaviour per SP.  So when a users enabled the checkbox, this only applies when logging in to this specific SP.  When the user logs in to a different SP, they still get the WAYF.

The following changes are required:

- Add the following extra controls to `parameters.yml`:
   - introduce feature-toggle `wayf.remember_choice_per_idp`.  All of the below should only be active if both `wayf.remember_choice` and `wayf.remember_choice_per_idp` are active.  
     - If `wayf.remember_choice` is true and `wayf.remember_choice_per_idp` is false, we keep the current global memory feature.  
     - If `wayf.remember_choice` is false, we disable all remember-my-choice features (both global and per-sp).  
     - If `wayf.remember_choice_per_idp` is not defined, act as if it were set to `false`
   - a parameter to control the life-time of the remembered IdP: `wayf.remember_choice_per_idp_lifetime`.  Typically this would be something like 3 months, but it would be nice to be able to set it to (e.g.) 60 seconds for easy debugging.
   - a parameter to control how many SP-IdP pairs we can store in the cookie `wayf.remember_choice_per_idp_lifetime`.  This defaults to `16`.  For the logic what to do if we exceed this limit, see below.
- support new `coin:wayf_remember_choice` metadata parameter for SPs.  If this is not present in the metadata, it defaults to `false`.
- if feature is enabled, read and set a cookie `rememberedidps` with content like this:
    ```json
    {
        "https://sp_entityid_1": { "idp": "https://idp_entityid_1", "expires": 1784534287 },
        "https://sp_entityid_2": { "idp": "https://idp_entityid_2", "expires": 1784512345 }
    }
    ```

  -  Ideally, we would like this cookie to be user-readable (for power users and privacy nerds), but cookies are limited to 4kB so that probably won't work.  So the safe way to handle this is to deflate and base64-encode the json, en put that in the cookie.
  - To make sure the structure fits in the cookie, limit this to max 32 entries.   
      - If there are more, or a new one needs to be added, drop the oldest ones until there are 32 left.  
[...]

PR #345, opened 2016-09-28
Include development wiki page

DRvanR:
Also standardized the layout and used markdown styles a bit.

PR #2091, opened 2026-09-14
Implement per-SP WAYF remember-choice core (#2065)

kayjoosten:
## Summary

Implements the EngineBlock backend core for per-SP WAYF remembered IdP choices, the first vertical slice of #2065 (part of epic #1956 — "Remember my Choice for this Service" feature in WAYF).

This PR is intentionally scoped to a safe, independently testable backend slice. It preserves all existing behavior (one-IdP shortcut, SSO-notification shortcut, and the existing global `rememberchoice` cookie feature) and adds a new, disabled-by-default, opt-in per-SP mode.

## What this PR adds

- **Configuration**: `wayf.remember_choice_per_idp` (default `false`), `wayf.remember_choice_per_idp_lifetime` (default `7776000`s / 90 days), `wayf.remember_choice_per_idp_max` (default `16`).
- **Metadata gate**: new SP metadata coin `coin:wayf_remember_choice` (Manage `metadata:coin:wayf_remember_choice`), defaulting to `false`, with strict boolean parsing so only `true`/`"1"` opt an SP in.
- **`RememberedIdpCookie` service** (`src/OpenConext/EngineBlock/Service/Wayf/`): a deterministic codec/policy service for the new `rememberedidps` cookie — JSON entries keyed by SP entity ID, deflated + base64 encoded, with strict validation (invalid data is always treated as absent and cleared, never used for IdP selection), expiry cleanup on every read, and over-limit pruning only when a new choice is written (never during passive reads).
- **Runtime bridge**: the cookie service and per-SP feature flag are exposed via `DiContainerRuntime` (not `DiContainer`), matching the legacy/modern bridge convention for new dependencies.
- **Read-side bypass** (`SingleSignOn`): when the global `wayf.remember_choice`, the new per-SP flag, and the SP's metadata coin are all enabled (and the request is not `ForceAuthn`/debug), a valid remembered IdP that is still a candidate IdP skips the WAYF automatically. The existing one-IdP shortcut and SSO-notification shortcut remain unconditionally first. The legacy global `rememberchoice` cookie path is now only used when per-SP mode is disabled.
- **Write-side persistence** (`ContinueToIdp`): when a user has explicitly checked "remember my choice" and all three per-SP gates are enabled, the `(SP, IdP)` pair is persisted in the cookie with a 90-day expiry. Cookie hygiene (invalid/expired cleanup) always runs regardless of whether the checkbox was checked.
- **Minimal frontend signal**: a hidden `rememberChoice` field is added to the WAYF's per-IdP form. In per-SP mode, checking the existing checkbox sets this field instead of creating the legacy client-side cookie; the checkbox itself is visually unchanged.
- **Cookie attributes**: the new cookie is `Secure`, `HttpOnly`, `SameSite=None`, scoped to the EngineBlock host only (not the wildcard locale-cookie domain), with the configured lifetime.

## What this PR does not include (tracked as follow-up work)

- #2066 — checkbox/tooltip visual redesign (existing checkbox markup/wording is unchanged here)
- #2085 — EngineBlock endpoint to reset/remove the cookie
- OpenConext/OpenConext-profile#345 and OpenConext/OpenConext-profile#350 — Profile integration (reset action + feedback)
- #2084 — full Cypress/Behat acceptance coverage for the epic (this PR adds focused unit/integration coverage only)
- OpenConext/OpenConext-profile#346 and OpenConext/OpenConext-profile#348 remain on hold per the epic discussion

Recommended merge order after this PR: #2085 → #2066 → Profile #345/#350 → #2084.

## Testing

All run inside the project's PHP 8.5 Docker container as `www-data`:

- `phpmd`, `phpcs` (both `src` and legacy `library` rulesets), `docheader`: clean
- PHPUnit `eb4`: 244/244 passed
- PHPUnit `unit`: 998/998 passed
- PHPUnit `integration`: 105/105 passed
- PHPUnit `functional` (`APP_ENV=test`): 119/119 passed
- Behat `default` suite: 300/300 scenarios, 5557/5557 steps passed
- `lint:twig theme/`: 114/114 valid
- `theme && yarn lint`: passed

New focused test coverage includes:
- `RememberedIdpCookieTest` — codec round-trip, all invalid-input cases, expiry cleanup, pruning-only-on-add, candidate-safe lookup, cookie write/clear, and the shared cookie-hygiene/gating helpers
- `WayfScriptConfigRenderTest` — a real-Twig-render regression test ensuring the per-SP flag actually reaches the rendered WAYF page (this caught and fixed a template-context bug during development)
- `ContinueToIdpTest` — write-gate coverage for the persistence decision
- Updated `DiContainerRuntimeTest`, `WayfRendererTest`, `WayfTest`, `PushMetadataAssemblerTest`, `ServiceProviderTest`, `BindingsTest`

## Behavior preserved

- One-candidate-IdP shortcut and SSO-notification shortcut remain first, unconditionally
- Existing global `rememberchoice` behavior is byte-for-byte unchanged when `wayf.remember_choice_per_idp` is `false`
- An SP without the new metadata coin never uses the new per-SP cookie and never silently falls back to the old global cookie once per-SP mode is globally enabled


config/packages/parameters.yml.dist
```diff
@@ -182,6 +182,12 @@ parameters:
 
     ## Allow users to save their selected IdP and then auto-select it on returning visits.
     wayf.remember_choice: false
+    ## Allow SPs to opt in to remembered IdP choices scoped per SP through Manage metadata.
+    wayf.remember_choice_per_idp: false
+    ## Lifetime for remembered IdP choices scoped per SP, in seconds.
+    wayf.remember_choice_per_idp_lifetime: 7776000
+    ## Maximum number of remembered IdP choice entries to retain per SP.
+    wayf.remember_choice_per_idp_max: 16
```
kayjoosten: Check what happens if we hit the 16 mark

src/OpenConext/EngineBlock/Metadata/Entity/Assembler/PushMetadataAssembler.php
```diff
@@ -46,6 +46,8 @@
  */
 class PushMetadataAssembler implements MetadataAssemblerInterface
 {
+    public const COIN_WAYF_REMEMBER_CHOICE = 'wayf_remember_choice';
```
kayjoosten: inline this

PR #2094, opened 2026-09-15
Add endpoint to reset the per-SP WAYF remember-choice cookie

kayjoosten:
## What

Implements #2085, adding `GET /reset-remember-wayf`: a public endpoint that
clears the per-SP `rememberedidps` cookie (introduced in #2065 / #2091) and
redirects the user to an operator-configured URL.

## Stacked on #2091

This branches from `wayf-remember-choice-core` (#2091, not yet merged),
since it depends directly on the `RememberedIdpCookie` service introduced
there. The diff shown here is the incremental change for #2085 only; once
#2091 merges to `main`, this PR's base should be retargeted to `main`.

## Details

- New `ResetRememberedWayfController` reuses the existing
  `RememberedIdpCookie` service: `normalize()` computes the valid entry
  count for logging, `clear()` removes the cookie using the same
  SameSite/secure flags it was written with. Both only run when the
  cookie is actually present — a visitor with no cookie is redirected
  without any log noise.
- Logs `'WAYF-remember-my-choice cookie removed (had %d entries)'` at
  INFO level, matching the sibling log message conventions from #2065.
- The controller validates its configured redirect URL at construction
  time and throws if it's blank, so a missing
  `wayf.reset_choice_per_idp_redirect` fails fast instead of silently
  redirecting nowhere.
- New config `wayf.reset_choice_per_idp_redirect` defaults to a working
  dev URL in `parameters.yml.dist` (matching the convention of other
  `wayf.*` dev defaults), so the endpoint works out of the box in
  dev/CI while still requiring operators to set a real destination in
  production.
- The functional test asserts the redirect target against the
  configured parameter rather than merely checking for a non-empty
  `Location` header — this app has a global exception listener that
  also turns uncaught exceptions into a 302, which would otherwise let
  a misconfiguration pass unnoticed. (Verified locally: temporarily
  blanking the config reproduces exactly this failure mode, and the
  strengthened assertion catches it.)

## Testing

All run inside the Docker dev container (PHP 8.5):

- `phpmd` / `phpcs` / `phpcs-legacy` / `docheader`: clean.
- `eb4`: 244/244. `unit`: 1003/1003 (999 baseline + 4 new). `functional`
  (`APP_ENV=test`): 121/121 (119 baseline + 2 new). `integration`:
  105/105.
- Behat `default` suite: 300/300 scenarios, 5557/5557 steps.
- Twig lint: 114/114. `yarn lint`: clean.

Refs: #2085


tests/functional/OpenConext/EngineBlockBundle/Controller/ResetRememberedWayfControllerTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+use OpenConext\EngineBlock\Service\Wayf\RememberedIdpCookie;
+use PHPUnit\Framework\Attributes\Test;
+use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\HttpFoundation\Response;
+
+final class ResetRememberedWayfControllerTest extends FunctionalWebTestCase
+{
+    #[Test]
+    public function visiting_the_endpoint_with_a_remembered_idp_cookie_clears_it_and_redirects(): void
```
kayjoosten: where do we check if the cookie is gone?

